PR #4, opened 2019-07-02
[WIP] localvolumeprovisionerdefault: add new storage class

gpaul:
This PR clones the `./stable/localvolumeprovisioner` storage class to `./stable/localvolumeprovisionerdefault`. It follows https://github.com/mesosphere/charts/pull/2 

A superior strategy would be to make refactor the two, to have `default` be a subchart of the main one:

- `./stable/localvolumeprovisioner/charts/default`: 
    Declares the LVP chart with `is-default-class: true`.
- `./stable/localvolumeprovisioner`: 
    imports the subchart but overrides `is-default-class: false`.

However, that goes against current convention where we distinguish storage classes as either "additional" or "default" as follows:

```
awsebscsidriverstorageclass / awsebscsidriverstorageclassdefault
awsstorageclass / awsstorageclassdefault
localstorageclass / localstorageclassdefault
```
~ https://github.com/mesosphere/kubeaddons-configs/tree/4b1681663f60b3a537492e07f5ebcfb4b3d5e5e3/templates

For this PR, we stick to convention, though the duplication is painful - and leave a refactor of this pattern (and concurrent cleanup of the quoted example classes) as future work.

### DIFF
--- a/stable/localvolumeprovisionerdefault/Chart.yaml
+++ b/stable/localvolumeprovisionerdefault/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Local persistent volume provisioner as the default storageclass for konvoy
+name: localvolumeprovisionerdefault
+version: 0.1.0
--- a/stable/localvolumeprovisionerdefault/templates/NOTES.txt
+++ b/stable/localvolumeprovisionerdefault/templates/NOTES.txt
@@ -1,0 +1,9 @@
+The following directories have been created on worker nodes:
+
+{{- range .Values.storageclasses }}
+/mnt/{{ .name }}
+{{- end }}
+
+The directory names match their corresponding storage classes.
+By mounting volumes into these directories, they will be made available
+as persistent volume in their respective storage class.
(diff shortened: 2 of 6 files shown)


stable/localvolumeprovisionerdefault/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Local persistent volume provisioner as the default storageclass for konvoy
+name: localvolumeprovisionerdefault
+version: 0.1.0
```
hectorj2f: Following the helm workflows, we'd have to bump this version :)

joejulian:
> ... that goes against current convention ...

Current convention was a crappy hack because we weren't using charts. I'd rather we use a flag in values to enable the annotation in the template.

hectorj2f:
> Current convention was a crappy hack because we weren't using charts. I'd rather we use a flag in values to enable the annotation in the template.

@gpaul  Yeah, let's add a flag that by default is enabled

gpaul:
Let's make `default` configurable on `localvolumeprovisioner` instead.

PR #7, opened 2019-07-02
Makefile: set the url of chart repository

joejulian:
We need to set the url of the chart repository to get the index with the appropriate entry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ publish:
 	@make all
 	@git add .
 	@git commit -m $(LAST_COMMIT_MESSAGE)
-	@echo git push -f publish master
+	@git push -f publish master
 	@git checkout -
 
 $(HELM):
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ STABLE_TARGETS = $(shell hack/chart_destination.sh $(STABLE_CHARTS))
 STAGING_CHARTS = $(wildcard staging/*/Chart.yaml)
 STAGING_TARGETS = $(shell hack/chart_destination.sh $(STAGING_CHARTS))
 
+LAST_COMMIT_MESSAGE := $(shell git reflog -1 | sed 's/^.*: //')
+
 TMPDIR := $(shell mktemp -d)
 HELM := $(shell bash -c "command -v helm")
 ifeq ($(HELM),)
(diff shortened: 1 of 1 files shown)


docs/stable/index.yaml
```diff
@@ -1,3 +1,13 @@
 apiVersion: v1
-entries: {}
-generated: 1970-01-01T00:00:00.000000000-00:00
+entries:
```
joejulian: please revert these changes to this and the staging/index.yaml

docs/staging/index.yaml
```diff
@@ -1,3 +1,3 @@
 apiVersion: v1
 entries: {}
-generated: 1970-01-01T00:00:00.000000000-00:00
+generated: 2019-07-02T15:02:08.769246+02:00
```
joejulian: revert

joejulian: And delete the tarball

docs/stable/index.yaml
```diff
@@ -1,3 +0,0 @@
-apiVersion: v1
```
joejulian:
```
apiVersion: v1
entries: {}
generated: 1970-01-01T00:00:00.000000000-00:00
```
for both docs/stable/index.yaml and docs/staging/index.yaml

This ensures that if either are unable to be generated (no charts, for instance) an empty index file will still exist.

Makefile
```diff
@@ -50,8 +52,7 @@ $(STABLE_TARGETS) $(STAGING_TARGETS): $(TMPDIR)/.helm/repository/local/index.yam
 %/index.yaml: $(STABLE_TARGETS) $(STAGING_TARGETS)
 %/index.yaml: $(TMPDIR)/.helm/repository/local/index.yaml
 	@mkdir -p $(patsubst %/index.yaml,%,$@)
-	$(HELM) --home $(TMPDIR)/.helm repo index $(patsubst %/index.yaml,%,$@)
```
gpaul: Could we get `mesosphere` from `git remote origin`? I'd like to use this on my fork. Something like:

```
--url=https://$(shell git remote get-url origin | cut -d':' -f2 | cut -d'/' -f1).github.io/charts/$(patsubst...)
```

gpaul: higher up in  this file, could we have:
```
-	@git remote add publish git@github.com:mesosphere/charts >/dev/null 2>&1 || true
+	@git remote add publish $(shell git remote get-url origin) >/dev/null 2>&1 || true
```

gpaul: I'll fix these in my next PR, no worries.

PR #11, opened 2019-07-03
charts/staging: add dex-k8s-authenticator repository

mhrabovcin:
This chart is copied from https://github.com/mintel/dex-k8s-authenticator which does not provide URL structure so that the chart can be directly added as a helm repo.

I've created `LICENSE` file which copies original license and adds a reference to the source repository.

### DIFF
--- a/staging/dex-k8s-authenticator/Chart.yaml
+++ b/staging/dex-k8s-authenticator/Chart.yaml
@@ -6,6 +6,9 @@ version: 1.1.1
 sources:
 - https://github.com/mintel/dex-k8s-authenticator
 maintainers:
+- name: Martin Hrabovcin
+  email: mhrabovcin.c@mesosphere.com
+# Original maintainers from `mintel/dex-k8s-authenticator` repository
 - name: Aaron Roydhouse
   email: aaron@roydhouse.com
 - name: Nick Badger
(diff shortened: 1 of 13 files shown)


staging/dex-k8s-authenticator/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+appVersion: "v1.1.0"
+description: "Authenticator for using Dex with Kubernetes"
+name: dex-k8s-authenticator
+version: 1.1.1
+sources:
+- https://github.com/mintel/dex-k8s-authenticator
+maintainers:
```
hectorj2f: Are they the maintainers ? @mhrabovcin Please, add yourself here :)

staging/dex-k8s-authenticator/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+appVersion: "v1.1.0"
+description: "Authenticator for using Dex with Kubernetes"
+name: dex-k8s-authenticator
+version: 1.1.1
```
hectorj2f: Note: This version doesn't need to be the same version supported in the upstream repo, but it helps to identify the version used in these scripts.

staging/dex-k8s-authenticator/Chart.yaml
mhrabovcin: Done @hectorj2f

PR #12, opened 2019-07-04
Port storage addons to charts repo from kubeaddons-config

gpaul:
This PR ports storage addons from manifest-based addons defined in kubeaddons-config to helm charts defined in the charts repository.


gpaul:
What does the "blocked" label mean in this case?

stable/awsebscsiprovisioner/templates/storageclass.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: awsebscsiprovisioner
+  annotations:
+    kubernetes.io/description: AWS EBS CSI provisioner StorageClass
+    storageclass.kubernetes.io/is-default-class: {{ .Values.storageclassdefault | default "true" | quote }}
+provisioner: ebs.csi.aws.com
+volumeBindingMode: WaitForFirstConsumer
```
hectorj2f: Please, add this property to be configurable and also `reclaimPolicy: Delete` too.

gpaul:
If we're going to do more iteration on this PR and it's associated ones in our other repositories, we could consider taking on board the changes suggested by @alejandroEsc in https://github.com/mesosphere/charts/pull/14

stable/localvolumeprovisioner/values.yaml
```diff
@@ -1 +1,4 @@
-storageclassdefault: true
+storageclass:
```
nfnt: This change isn't compatible with what is checked for in https://github.com/mesosphere/kubeaddons/blob/master/pkg/utils/supported_addons.go#L348

PR #13, opened 2019-07-05
Makefile: don't update system-wide known hosts file

gpaul:
The publish target currently modifies the system-wide `/etc/ssh/ssh_known_hosts` file when adding github keys. This is an invasive change that typically requires root privileges. This commit changes that to use the user's `$HOME/.ssh/known_hosts` file instead, which typically requires no extra permissions.

It also fixes an issue where running `make publish` multiple times adds multiple github keys to the known_hosts file. It does so by using `ssh-keygen -R` to remove all keys for github.com from the known_hosts file before using `ssh-keyscan` to add it back again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,12 @@ publish:
 	@make all
 	@git add .
 	@git commit -m "$(LAST_COMMIT_MESSAGE)"
-	@ssh-keyscan github.com >> /etc/ssh/ssh_known_hosts
-	@chmod 0644 /etc/ssh/ssh_known_hosts
+	@mkdir -p $(HOME)/.ssh
+	@chmod 0700 $(HOME)/.ssh
+	@touch $(HOME)/.ssh/known_hosts
+	@ssh-keygen -R github.com -f $(HOME)/.ssh/known_hosts
+	@ssh-keyscan github.com >> $(HOME)/.ssh/known_hosts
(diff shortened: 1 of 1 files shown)


gpaul:
Hmm - perhaps we can try the following?

```
GIT_SSH_COMMAND="ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" git push -f publish master
```
~ https://stackoverflow.com/a/28527476

joejulian:
That or set the UserKnownHostsFile to something specific and ssh-keyscan the key to that file.

gpaul:
Yeah that would work, too. Either way we score no security as far as I can tell.

joejulian:
It's true. I also don't care either way.

Makefile
```diff
@@ -46,8 +46,12 @@ publish:
 	@make all
 	@git add .
 	@git commit -m "$(LAST_COMMIT_MESSAGE)"
-	@ssh-keyscan github.com >> /etc/ssh/ssh_known_hosts
-	@chmod 0644 /etc/ssh/ssh_known_hosts
+	@mkdir -p $(HOME)/.ssh
+	@chmod 0700 $(HOME)/.ssh
```
hectorj2f: why do you need to change the permissions of $(HOME)/.ssh ?

joejulian:
This has been converted to use https so this PR is no longer relevant. Closing.

PR #15, opened 2019-07-08
charts: add kommander

hectorj2f:


stable/kommander/values.yaml
```diff
@@ -0,0 +1,5 @@
+kommander:
+  replicas: 1
+  image:
```
alejandroEsc: pull policy would be nice to have here.

alejandroEsc: I would suggest image would be a first level field

stable/kommander/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
+---
```
alejandroEsc: Can we move service account (roles and bindings) information to another file.

stable/kommander/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+apiVersion: v1
+kind: Namespace
```
alejandroEsc: Hmmm, i honestly forget if the namespace is created or automagically for you.

stable/kommander/values.yaml
hectorj2f: Moved

hectorj2f: ack

stable/kommander/templates/deployment.yaml
hectorj2f: Done

stable/kommander/templates/deployment.yaml
hectorj2f: It should get all created at the same time.

joejulian:
Rats... I was too slow... You don't normally create namespaces in charts, you use the `Release.Namespace` variable to reference the namespace in which this chart was installed.

PR #22, opened 2019-07-09
Use Release.namespace to pass the namespace name to kommander

hectorj2f:
Use Release.namespace instead of hardcoding the namespace

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.24.2"
 description: Kommander
 name: kommander
-version: 0.1.0
+version: 0.1.1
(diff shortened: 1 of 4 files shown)


joejulian:
tiller will create a namespace when you helm install to a namespace, ie: 
"helm install --namespace fubar stable/minecraft" will create the 
namespace "fubar" if it does not exist. We should not put namespace 
records in helm charts.

hectorj2f:
@joejulian It is a request from @DanielMSchmidt to be able to configure the namespace. Also, there are helm charts creating the namespace in the helm hub.

alejandroEsc:
I would love to see the reasons for namespace in those charts, this does break convention @hectorj2f, @DanielMSchmidt why do we need this to be explicitly part of the chart?

PR #24, opened 2019-07-09
dex-k8s-authenticator: add custom templates

mhrabovcin:
A new version of the chart that deploys authenticator version that does not expose OAuth client secret to the authenticated user.

To override HTML templates in the official container this PR introduces new `ConfigMap` that injects custom version of HTML templates into the authenticator container.

JIRA: https://jira.mesosphere.com/browse/DCOS-56105

### DIFF
--- a/staging/dex-k8s-authenticator/templates/configmap_templates.yaml
+++ b/staging/dex-k8s-authenticator/templates/configmap_templates.yaml
@@ -5,7 +5,7 @@ metadata:
   name: {{ template "dex-k8s-authenticator.fullname" . }}-html-templates
   labels:
     app: {{ template "dex-k8s-authenticator.fullname" . }}
-    env: {{ default "dev" .Values.global.deployEnv }}
+    env: {{ .Values.global.deployEnv }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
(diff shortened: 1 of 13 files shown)


mhrabovcin:
This PR enables https://github.com/mesosphere/kubeaddons-configs/pull/81

staging/dex-k8s-authenticator/templates/configmap_templates.yaml
```diff
@@ -0,0 +1,13 @@
+# A ConfigMap that stores customized `dex-k8s-authenticator` HTML templates.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "dex-k8s-authenticator.fullname" . }}-html-templates
+  labels:
+    app: {{ template "dex-k8s-authenticator.fullname" . }}
+    env: {{ default "dev" .Values.global.deployEnv }}
```
alejandroEsc: no need to have default here, values file should be enough.

mhrabovcin: @alejandroEsc this is consistent with what we have in other templates for this package. Would you prefer to go over all templates and fix default value for `env` label?

mhrabovcin:
@joejulian in the upstream we'd solve this probably with a configuration flag that would require modifications of app itself. Is that the preference? I can probably create a JIRA as this is not a requirement for the GA.

mhrabovcin:
https://jira.mesosphere.com/browse/DCOS-56260 for tracking upstream contribution.

staging/dex-k8s-authenticator/templates/configmap_templates.yaml
alejandroEsc: There are items we need to address in a lot of these charts. Let us try to correct things as we go. Please keep changes in PRs to one per chart. Thanks.

PR #26, opened 2019-07-11
awsebscsiprovisioner chart updated

alejandroEsc:
moving aws ebs csi provider to a chart


alejandroEsc:
This is WIP as i have to test it out.

stable/awsebscsidriver/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "0.3.0"
+description: AWS EBS CSI Driver
+name: awsebscsidriver
```
hectorj2f: why are we adding awsebscidriver as a chart ? It is part of the awsebscsiprovisioner at this moment.

hectorj2f: To simplify the addons, we provide fout storage addons, localpath, localvolume, ebscsi and ebs.

alejandroEsc: https://jira.mesosphere.com/browse/DCOS-56399

alejandroEsc: This is super confusing, not to mention the naming we are using for all of this. Provisioner vs. Driver is still a valid change. Let's settle this in standup.

hectorj2f: That was the decision taken during the storage meetings. But, yeah, let's discuss it during the standup.

stable/awsebscsiprovisioner/.helmignore
```diff
@@ -0,0 +1,22 @@
+# Patterns to ignore when building packages.
```
s12chung: I don't think this should be included...

alejandroEsc: ok, np.

stable/awsebscsiprovisioner/values.yaml
```diff
@@ -1,4 +1,60 @@
...
+# AWS key id and access key
+# these are optional and can be left as is
+secrets:
+  enabled: false
+  key_id: ""
+  access_key: ""
+
 storageclass:
-  isDefault: true
+  isDefault: false
```
hectorj2f: we wanted to be this true to be consistent with the other provisioners

alejandroEsc: Seems like an arbitrary decision. Ill set it true in another PR.

PR #27, opened 2019-07-11
opsportal chart

alejandroEsc:


stable/opsportal/templates/deployment-proxy.yaml
```diff
@@ -0,0 +1,33 @@
+---
```
joejulian: What about adding if logic around this and create a value to enable it; default the value to disabled and override that value in the Addon? That way when we have proper authorization we can stop using the bearerproxy.

stable/opsportal/templates/service-proxy.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
joejulian: This, too.

stable/opsportal/templates/roles.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
joejulian: This too.

stable/opsportal/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+    repository: "quay.io/joejulian/bearerproxy"
+    tag: "latest"
+    pullPolicy: "Always"
+
+opsportal:
+  ingress:
+    paths:
+      # kubernetes-dashboard
+      - backend:
+          serviceName: bearerproxy
```
joejulian: And this would need logic, too, to either select the bearerproxy if it's enabled, or opsportal-landing if it's not.

alejandroEsc: Sure thing.

stable/opsportal/values.yaml
```diff
@@ -0,0 +1,53 @@
...
...LTE3LjJoLTI2LjNoLTE0LjloLTI5LjMKCQkJYy0xOS42LDAtMTkuNiw1Mi4xLTEuMSw1NS4xbDM4LjEtMi40YzQuNiwwLDEzLjgtMTcuOCwxNS41LTE5LjJjMS43LTEuNCw3LjQtMS42LDkuNiwwczcuMywxNy40LDExLjksMTcuNGwzOC4xLDQuMgoJCQlDMzM5LjIsMzA4LjIsMzM0LjcsMjYxLjMsMzE4LjYsMjYxLjNMMzE4LjYsMjYxLjN6Ii8+CgkJPHBhdGggaWQ9InBhdGgzMjcyIiBpbmtzY2FwZTpjb25uZWN0b3ItY3VydmF0dXJlPSIwIiBjbGFzcz0ic3QxNSIgZD0iTTIzOC4zLDI3Mi4xYy0xMi0wLjMtMzAuOS0wLjItNDEuMi0wLjEKCQkJYy0xLjcsNS43LTMuNiwxMi44LTIuNywxOS44YzkuMywwLDI5LjcsMCw0NC4xLDBjOC43LDAsMTYsMS42LDIyLjgsMy40YzAuMi0wLjIsMC4zLTAuNCwwLjQtMC41YzEuNy0xLjQsNy40LTEuNiw5LjYsMAoJCQljMC42LDAuNCwxLjQsMS44LDIuMywzLjZjMy41LDAuNyw2LjksMS4yLDEwLjYsMS4zYzkuOCwwLjEsMzIuOSwwLjEsNDcuOCwwYzEuOS02LjMsMC41LTEzLjEsMC4xLTE5LjVjLTE0LjUsMC0zNy41LTAuMy00OC42LTAuMwoJCQlDMjY3LjksMjc5LjksMjU0LjMsMjcyLjQsMjM4LjMsMjcyLjFMMjM4LjMsMjcyLjF6Ii8+Cgk8L2c+Cgk8cGF0aCBpZD0icGF0aDMyNzQiIGlua3NjYXBlOmNvbm5lY3Rvci1jdXJ2YXR1cmU9IjAiIGQ9Ik0xNDEuOCwxMTAuM2MtMC4zLTcuMiwxMy40LTguMSwxNS0yLjEKCQlDMTU4LjQsMTE0LjIsMTQyLjUsMTE1LjYsMTQxLjgsMTEwLjNDMTQxLjMsMTA2LDE0MS44LDExMC4zLDE0MS44LDExMC4zeiIvPgo8L2c+Cjwvc3ZnPgo="> <p>Traefik Dashboard</p></a> </section> </div></main><footer class="footer">© 2019 Mesosphere, Inc. All Rights Reserved</footer>'
```
mhrabovcin: You could put this string into an actual file and load it as a template variable. I did the same thing for `dex-k8s-authenticator` - https://github.com/mesosphere/charts/blob/dev/staging/dex-k8s-authenticator/templates/configmap_templates.yaml#L13

alejandroEsc: not a bad idea, let me look at that, thanks!

PR #30, opened 2019-07-13
Add ingress to kommander

joejulian:
Add an ingress record so we can allow access to the kommander
application.

Changed to more closely follow chart design guidelines.

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -1,6 +1,38 @@
-kommander:
-  image:
-    repository: mesosphere/kommander
-    tag: 1.24.2
-    pullPolicy: IfNotPresent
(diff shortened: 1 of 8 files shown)


stable/kommander/values.yaml
```diff
@@ -1,6 +1,38 @@
...
+
+extraInitContainers:
+
+resources:
+  requests:
+    memory: "128Mi"
+    cpu: "1000m"
+  limits:
+    memory: "512Mi"
+    cpu: "4000m"
```
hectorj2f: this is really high for kommnader, let's put 2000 although I think we should check what it uses on regular basis..

joejulian: Changed.

PR #31, opened 2019-07-13
dex-k8s-authenticator: reduce timeout

jieyu:
Reduce some timeout values for readiness probe and liveness probe.


hectorj2f:
Did you test them ? Are you sure these values are good now ?

PR #33, opened 2019-07-15
stable: add X-Forwarded-User to opsportal and kommander

hectorj2f:
I have the feeling we want to change this too :)

### DIFF
--- a/stable/kommander/templates/ingress.yaml
+++ b/stable/kommander/templates/ingress.yaml
@@ -10,10 +10,7 @@ metadata:
     heritage: "{{ .Release.Service }}"
   annotations:
     kubernetes.io/ingress.class: traefik
-    traefik.frontend.rule.type: PathPrefixStrip
-    ingress.kubernetes.io/auth-secret: ops-portal-htpasswd
-    ingress.kubernetes.io/auth-type: basic
-    ingress.kubernetes.io/auth-header-field: X-Forwarded-User
+    traefik.frontend.rule.type: {{ .Values.ingress.traefikFrontendRuleType }}
     {{- with .Values.ingress.extraAnnotations }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
--- a/stable/kommander/templates/ingress.yaml
+++ b/stable/kommander/templates/ingress.yaml
@@ -13,7 +13,7 @@ metadata:
     traefik.frontend.rule.type: PathPrefixStrip
     ingress.kubernetes.io/auth-secret: ops-portal-htpasswd
     ingress.kubernetes.io/auth-type: basic
-    ingress.kubernetes.io/auth-header-field: X-WebAuth-User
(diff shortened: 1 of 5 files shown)


stable/kommander/templates/ingress.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     traefik.frontend.rule.type: PathPrefixStrip
     ingress.kubernetes.io/auth-secret: ops-portal-htpasswd
     ingress.kubernetes.io/auth-type: basic
-    ingress.kubernetes.io/auth-header-field: X-WebAuth-User
+    ingress.kubernetes.io/auth-header-field: X-Forwarded-User
```
jieyu: I don't think this will work if the `auth-type` is still `basic`.

jieyu:
@brandonc we probably don't need this change, but we need to make sure kommander service expect `X-Forwarded-User` instead of `X-WebAuth-User` from the proxy

stable/kommander/templates/ingress.yaml
jieyu: actually, scratch that. This is `auth-header-field` not `auth-responses-header`.

stable/opsportal/templates/ingress-opsportal.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     traefik.frontend.rule.type: PathPrefixStrip
     ingress.kubernetes.io/auth-secret: ops-portal-htpasswd
```
jieyu: This has changed. We probably need to remove all those labels and use forward auth instead

hectorj2f: done

stable/kommander/templates/ingress.yaml
```diff
@@ -11,9 +11,7 @@ metadata:
   annotations:
     kubernetes.io/ingress.class: traefik
     traefik.frontend.rule.type: PathPrefixStrip
```
jieyu: I think this should be configurable by values.yaml. For the konvoy ui page, this annotation should not be PathPrefixStrip.

stable/opsportal/templates/ingress-opsportal.yaml
```diff
@@ -5,9 +5,7 @@ metadata:
   annotations:
     kubernetes.io/ingress.class: traefik
     traefik.frontend.rule.type: PathPrefixStrip
-    ingress.kubernetes.io/auth-secret: ops-portal-htpasswd
-    ingress.kubernetes.io/auth-type: basic
-    ingress.kubernetes.io/auth-header-field: X-WebAuth-User
+    ingress.kubernetes.io/auth-header-field: X-Forwarded-User
```
jieyu: Opsportal is going to be replaced by konvoy-ui (based on kommander chart, yes, the same chart with different configurations). So I won't bother do any change in this chart.

PR #34, opened 2019-07-16
Add initContainers to traefik-forward-auth chart

shaneutt:
This PR adds `initContainer` support to the `treafik-forward-auth` chart.


s12chung:
just need to bump version :)

PR #38, opened 2019-07-17
make file needed to be fixed

alejandroEsc:
found an error in the makefile


alejandroEsc:
going to self merge because this breaks the CI process.

PR #40, opened 2019-07-17
stable/dex: add dex chart with initContainers support

mhrabovcin:
JIRA: https://jira.mesosphere.com/browse/COPS-4881

Adds support for `initContainers` to official `stable/dex` helm chart. This is necessary in order to migrate `extrasteps` from `kubeadsons` repository.

The chart version has been bumped to `1.5.2` from `1.5.1`.

### DIFF
--- a/stable/dex/templates/clusterrole.yaml
+++ b/stable/dex/templates/clusterrole.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: {{ template "dex.name" . }}
+    chart: {{ template "dex.chart" . }}
+    heritage: "{{ .Release.Service }}"
+    release: "{{ .Release.Name }}"
+  name: {{ template "dex.fullname" . }}
+rules:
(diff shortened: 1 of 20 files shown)


mhrabovcin:
Upstream PR: https://github.com/helm/charts/pull/15588

stable/dex/templates/clusterrole.yaml
```diff
@@ -0,0 +1,18 @@
...
+    heritage: "{{ .Release.Service }}"
+    release: "{{ .Release.Name }}"
+  name: {{ template "dex.fullname" . }}
+rules:
+- apiGroups: ["dex.coreos.com"] # API group created by dex
+  resources: ["*"]
+  verbs: ["*"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["create"] # To manage its own resources, dex must be able to create customresourcedefinitions
```
shaneutt: Just double checking: `view` is not needed?

shaneutt: It's like this [upstream](https://github.com/helm/charts/blob/master/stable/dex/templates/clusterrole.yaml) so it would appear to not be a problem.

PR #41, opened 2019-07-17
remove ssh cruft

joejulian:
Since we're now using the https uri for publishing, we no longer need these ssh commands.


alejandroEsc:
@joejulian please sign your commits.

PR #42, opened 2019-07-17
chore(kommander): update kommander to 1.40.0

DanielMSchmidt:
This release is aware of its namespace, therefore we need to add an
environment variable with the namespace passed to it

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -1,6 +1,6 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.24.2
+  tag: 1.40.0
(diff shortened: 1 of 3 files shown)


stable/kommander/values.yaml
```diff
@@ -1,6 +1,6 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.24.2
+  tag: 1.40.0
```
hectorj2f: that is a big change :D

alejandroEsc:
can we get this version also bumbed in kubeaddons-configs please?

PR #43, opened 2019-07-18
Added a chart for knative-eventing.

chhsia0:
The contents are retrieved from https://github.com/knative/eventing/releases/download/v0.7.1/eventing.yaml.
The YAML files are organized as in https://github.com/knative/eventing/tree/master/config

JIRA: https://jira.mesosphere.com/browse/DCOS-56868


joejulian:
please rebase against current dev

chhsia0:
@joejulian Rebased and updated the README.

shaneutt:
Seems that there are some issues similar to your other parallel PR @chhsia0:

https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2110470&_focus=204

Let us know if you need any assistance with this PR and we'll help out :+1:

chhsia0:
Fixed. PTAL

staging/knative/charts/eventing/templates/:vsp controller.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
```
hectorj2f: the name of this filename contains a typo

staging/knative/charts/eventing/values.yaml
```diff
@@ -0,0 +1,2 @@
+global:
+  serviceLabels: {}
```
hectorj2f: what about having more available variables such as like this helm chart did at that time https://github.com/simplytunde/charts/blob/master/incubator/knative/values.yaml ?

staging/knative/charts/eventing/templates/:vsp controller.yaml
shaneutt: A fellow vim user :+1:

chhsia0: Hmm I thought I fixed this... guessed the fix was on another draft branch :(

staging/knative/charts/eventing/values.yaml
chhsia0: The whole `build` section is not needed, since we're not bundling `knative-build` in favor of the tekton pipeline. I don't see why the `eventing` section is necessary unless we have very specific needs to change the official images for the controller and webhook and the arguments.

Then, the remaining meaningful configuration options are:
1. Controlling if RBAC should be installed, which I think it should always be.
2. Setting up the nemaspace for just eventing. This might be handy if we want to configure it otherwise in konvoy. But for now I prefer keeping the default, since there are 3 namespaces involved, not just eventing, in the package.

Also, it seems to me that it would be more meaningful to add values for public configuration options in `ConfigMap`s, e.g.: https://github.com/knative/serving/blob/master/install/CONFIG.md

WDYT?

chhsia0: FIY this is the `value.yaml` from the most recent helm chart PR for knative:
https://github.com/tomleah/helm-charts/blob/2afbc130a23ab9aae052967bb1bc20eb739a783a/incubator/knative/values.yaml
Unfortunately the PR became inactive so I created these charts here.

staging/knative/charts/eventing/templates/:vsp controller.yaml
chhsia0: Fixed. Forgot to push :(

staging/knative/charts/eventing-sources/charts/github/values.yaml
```diff
@@ -0,0 +1,2 @@
+global:
+  serviceLabels: {}
```
hectorj2f: by looking at the configuration options, this doesn't look much as a chart. It looks more like a bunch of yaml files that could be installed with a kubectl create command without any customization

PR #44, opened 2019-07-18
should not deploy ops-portal-secret

alejandroEsc:
I may be mistaken, but other tools will create this secret, we probably should not create it in production as it will break and fail to deploy.

### DIFF
--- a/staging/traefik-forward-auth/templates/secret.yaml
+++ b/staging/traefik-forward-auth/templates/secret.yaml
@@ -6,17 +6,3 @@ type: Opaque
 data:
   clientSecret: {{ required "traefikForwardAuth.clientSecret is required" .Values.traefikForwardAuth.clientSecret.value | b64enc | quote }}
   secret: {{ required "traefikForwardAuth.secret is required" .Values.traefikForwardAuth.secret | b64enc | quote }}
-
-{{- if .Values.traefikForwardAuth.opsPortalSecret.enable }}
----
-apiVersion: v1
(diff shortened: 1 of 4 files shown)


staging/traefik-forward-auth/templates/secret.yaml
```diff
@@ -6,6 +6,8 @@ type: Opaque
 data:
   clientSecret: {{ required "traefikForwardAuth.clientSecret is required" .Values.traefikForwardAuth.clientSecret.value | b64enc | quote }}
   secret: {{ required "traefikForwardAuth.secret is required" .Values.traefikForwardAuth.secret | b64enc | quote }}
+
+{{- if .Values.traefikForwardAuth.opsPortalSecret.enable }}
```
hectorj2f: @jongiddy @mhrabovcin this sounds like an appropriate change to me. Does it also seem to you ?

jongiddy: Putting knowledge of the Ops Portal into a chart for a different service seems strange to me. I expect a chart to be generic. Given that this has been done anyway, I don't object to this particular change.

hectorj2f: okay, I am thinking about it ☕️...  Let's delete this secret here and so the configuration. We don't need it here.

alejandroEsc:
@hectorj2f done. I agree. Also should this chart be in /staging?

PR #46, opened 2019-07-18
Update kommander to 1.45.0

DanielMSchmidt:
This introduces new logout behavior and the chart updates allow
kommander to be configured as Konvoy UI.


stable/kommander/values.yaml
```diff
@@ -1,8 +1,10 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.40.0
+  tag: 1.45.0
   pullPolicy: IfNotPresent
 replicas: 1
+isKonvoyUI: true
```
hectorj2f: should not we default this to false ?

DanielMSchmidt: yep, changed it, great catch 👍

stable/kommander/values.yaml
```diff
@@ -1,8 +1,10 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.40.0
+  tag: 1.45.0
   pullPolicy: IfNotPresent
 replicas: 1
+isKonvoyUI: false
+logoutRedirectPath: "/"
```
hectorj2f: also, is this a default for kommander or it is only used for konvoy-ui too ?

hectorj2f: @DanielMSchmidt ☝️

DanielMSchmidt: I am unsure, it's the default we have in the backend code as well, so I think it's a save bet. @natmegs or @brandonc might know more about ops portal related logout routing

brandonc: It's used in both

shaneutt:
LGTM normally I would wait for the other reviewers, but we're in crunch time to start testing kommander for our release tomorrow so I'm going to merge so we can start testing asap.

PR #50, opened 2019-07-18
install ct if not installed

joejulian:


joejulian:
Not to worry. I had already done a full check and merged in fixes before adding this. :)

PR #52, opened 2019-07-19
landing for ops portal and Traefik config

s12chung:
still need to bump docker image version for whenever we get it

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -3,6 +3,6 @@ name: kommander
 home: https://github.com/mesosphere/kommander
 appVersion: "1.46.1"
 description: Kommander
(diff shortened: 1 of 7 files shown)


joejulian:
Ah, right. Sorry, I shouldn't have enabled that until #50 was actually merged.

hectorj2f:
I am thinking to merge this PR but we are not bumping the docker image tag :/ so I bet this won't work

hectorj2f:
@s12chung I pinged @DanielMSchmidt, I assume this new tag is the one we want to use `1.47.2` https://mesosphere.slack.com/archives/CJQURP1FT/p1563531459404700?thread_ts=1563527967.402900&cid=CJQURP1FT. But I'll give it a try

PR #53, opened 2019-07-19
kommander: add conditional to install the CRD

hectorj2f:
We are seeing this error when installing konvoy-ui and kommander, we should install one time the CRD:
```
konvoy-ui                                                              [ERROR]

konvoy-ui: konvoy-ui failed: exit status 1 [stderr: Error: object is being deleted: customresourcedefinitions.apiextensions.k8s.io "observableclusters.stable.mesosphere.com" already exists
]
```

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -3,6 +3,6 @@ name: kommander
 home: https://github.com/mesosphere/kommander
 appVersion: "1.47.2"
 description: Kommander
-version: 0.1.10
+version: 0.1.11
 maintainers:
   - name: hectorj2f
--- a/stable/kommander/templates/crd.yaml
+++ b/stable/kommander/templates/crd.yaml
@@ -1,3 +1,4 @@
+{{- if (.Values.createObservableClusterCRD) }}
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
@@ -62,3 +63,4 @@ spec:
     - name: Age
       type: date
       JSONPath: .metadata.creationTimestamp
+{{ end }}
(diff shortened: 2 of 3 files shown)


jaxesn:
Im wondering why this is an issue... I was thinking we added the namespace support so that we could namespace the two instances separately so we aren't sharing the same CRD list between the two...

cc @DanielMSchmidt @natmegs

hectorj2f:
@jaxesn I think there is a confusion about the term CRD here. The kubernetes resource definition that creates a schema that will allow creating  custom resources that are `namespaced or cluster-wide`.

hectorj2f:
The CRD of Kubernetes doesn't belong to any namespace.

jaxesn:
ah gotcha. thanks for the explanation!

DanielMSchmidt:
We should not need to do this. When a cluster runs with the latest
kommander chart for both kommander and konvoy ui the problem. In this case
there should only be namespaced observable clusters. I fear this change
could mascarade a version mismatch that ee do not support.

Shane Utt <notifications@github.com> schrieb am Fr., 19. Juli 2019, 19:02:

s12chung:
@DanielMSchmidt I think you're mixing between the CRD and the CRD instance. we need this. it's pretty standard amongst helm charts.

jaxesn:
I was trying to digest @DanielMSchmidt's comment, not totally sure but he may be referring to a situation where we have two versions of kommander, one that is more stable to power the konvoy-ui and one that is more alpha for kommander itself.  In this case, its possible that the CRD def might be slightly different in the alpha version because we add/remove a field or something.  Not an issue for today, but maybe Daniel and you guys can chat about this next week when we are in the clear of the release and mwt and decide if we need to handle this differently.

s12chung:
sure, we can talk next week.

PR #54, opened 2019-07-19
fix: remove logo and brand link

natmegs:
Remove branded assets on landing page.


jaxesn:
nothing quite like reviewing minified html + css :shipit:

alejandroEsc:
What happened to the html resource file page i created and why did we go back to using a line in the chart directly in the configmap? Also can you add me as a maintainer of this chart?

s12chung:
@alejandroEsc I didn't see the `html resource file page` anywhere. this code is pretty much the same as the old manifest `kubeaddons-configs` code because it had some changes from the security team.

this code was written late last night and the landing is a temporary stopgap until we find a better solution.

hectorj2f:
Please, rebase against `dev` we merged some changes

PR #55, opened 2019-07-19
chore: fix ops portal and kommander charts

jieyu:
Fix the traefik related annotations. Some traefik forward auth related
configurations should be moved to kubeaddons-configs.


stable/kommander/templates/ingress.yaml
```diff
@@ -10,13 +10,6 @@ metadata:
     heritage: "{{ .Release.Service }}"
   annotations:
     kubernetes.io/ingress.class: traefik
-{{- if (eq "konvoy" .Values.mode) }}
```
s12chung: you forgot to delete the match `{{- end }}`

PR #57, opened 2019-07-19
change all .storageclass.isDefault

s12chung:
because it only works when the boolean is quoted


alejandroEsc:
Holy smokes, yeah i caught that some time ago, i should have made the changes then and there but completely forgot to return to this.

PR #58, opened 2019-07-20
fix: change logout redirect to '/landing' in kommander

natmegs:
Redirect to on logout `/` leads to 404. Redirect to `/landing` instead.

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -3,7 +3,7 @@ image:
   tag: 1.50.0
   pullPolicy: IfNotPresent
 replicas: 1
-logoutRedirectPath: /
+logoutRedirectPath: /ops/landing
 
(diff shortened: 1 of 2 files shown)


s12chung:
every time you update this branch, you'll need to bump the version here:
https://github.com/mesosphere/charts/blob/2865317b6b03ec3e9c1c6d140a7a889abce3490a/stable/kommander/Chart.yaml#L6

stable/kommander/values.yaml
```diff
@@ -3,7 +3,7 @@ image:
   tag: 1.50.0
   pullPolicy: IfNotPresent
 replicas: 1
-logoutRedirectPath: /
+logoutRedirectPath: /landing
```
s12chung: it's now `/ops/landing`

PR #60, opened 2019-07-22
dex-k8s-authenticator: avoid using subPath for the secret

hectorj2f:
The cleanup of the subpath for this secret resulted in a problem when terminating the pod. This change should work the same without using `subPath`.

```
Jul 22 12:01:19 out-darwin-9d9b-worker-pool0-1 kubelet[686]: E0722 12:01:19.110863     686 nestedpendingoperations.go:270] Operation for "\"kubernetes.io/secret/5729f513-0433-4c56-a8b6-aafa9cbd4289-dex-k8s-authenticator-kubeaddons-dex-root-ca\" (\"5729f513-0433-4c56-a8b6-aafa9cbd4289\")" failed. No retries permitted until 2019-07-22 12:02:23.1108117 +0000 UTC m=+4373.047820401 (durationBeforeRetry 1m4s). Error: "error cleaning subPath mounts for volume \"dex-k8s-authenticator-kubeaddons-dex-root-ca\" (UniqueName: \"kubernetes.io/secret/5729f513-0433-4c56-a8b6-aafa9cbd4289-dex-k8s-authenticator-kubeaddons-dex-root-ca\") pod \"5729f513-0433-4c56-a8b6-aafa9cbd4289\" (UID: \"5729f513-0433-4c56-a8b6-aafa9cbd4289\") : error processing /var/lib/kubelet/pods/5729f513-0433-4c56-a8b6-aafa9cbd4289/volume-subpaths/dex-k8s-authenticator-kubeaddons-dex-root-ca/dex-k8s-authenticator: error cleaning subpath mount /var/lib/kubelet/pods/5729f513-0433-4c56-a8b6-aafa9cbd4289/volume-subpaths/dex-k8s-authenticator-kubeaddons-dex-root-ca/dex-k8s-authenticator/7: Unmount failed: exit status 32\nUnmounting arguments: /var/lib/kubelet/pods/5729f513-0433-4c56-a8b6-aafa9cbd4289/volume-subpaths/dex-k8s-authenticator-kubeaddons-dex-root-ca/dex-k8s-authenticator/7\nOutput: umount: /var/lib/kubelet/pods/5729f513-0433-4c56-a8b6-aafa9cbd4289/volume-subpaths/dex-k8s-authenticator-kubeaddons-dex-root-ca/dex-k8s-authenticator/7: not mounted\n\n"
```

PR #62, opened 2019-07-23
dex-k8s-authenticator: stop using subPath in general for dex-k8s-auth

hectorj2f:
I proposed to stay away of subPath volumes in general to avoid issues when mounting/unmounting vols on multiple OS and environments.
related issues: 
https://github.com/kubernetes/kubernetes/issues/50345 
https://github.com/kubernetes/kubernetes/issues/72991#issuecomment-508132380
also this one #60


hectorj2f:
> Maybe we could contribute this upstream?

Yeah, I agree we can make this change upstream. Let's do it after GA :).

PR #63, opened 2019-07-23
stable/storage charts: set tolerations and critical priority

hectorj2f:
During our mwt tests, we found that the pods were not deployed in all the nodes. Besides I found that the priority class of the localvolumeprovisioner pod is not set as a priority class.


hectorj2f:
@jieyu yes, let's add them as variables

PR #66, opened 2019-07-23
Initial addition of Prometheus-Operator

alejandroEsc:
- Prometheus-Operator chart has been added


jieyu:
We've been using https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheus.yaml#L34 for the entire time. Given that MWT will be using this version, can we fork this particular version instead?

PR #67, opened 2019-07-23
Add branded assets back to landing page

natmegs:
Add D2iQ logo, D2iQ link, align button padding with login pages, add D2iQ favicon, update HTML title. Had issues with svg resizing so logos are base64 encoded PNGs.


hectorj2f:
We might want to bump opsportal chart version in kubeaddons-configs as well

PR #69, opened 2019-07-23
Add polling interval env vars for kommander

natmegs:
Add env vars to kommander deployment that represent cluster polling interval (for kommander backend) and client polling interval (for kommander frontend).


hectorj2f:
@natmegs I did rebased this PR against `dev` for you :). Let's get it merge asap.

PR #70, opened 2019-07-24
[prometheus-operator] post install hook created, spacing fixed

alejandroEsc:
- Post install hook created
- Fix indentation problem


alejandroEsc:
This has be tested manually using values found in my branch of kubeadons-configs. Dashboard default page is working as expected.

PR #71, opened 2019-07-24
Fix awsebscsiprovisioner chart values

gpaul:
The chart refactoring in https://github.com/mesosphere/charts/pull/26 got the image tag version wrong in a few places. I've gone through that PR and checked that the values in values.yaml all match the pre-refactoring values.

Fixes https://jira.mesosphere.com/browse/DCOS-56901 - further PRs required to update all the various upstream repositories.

### DIFF
--- a/stable/awsebscsiprovisioner/values.yaml
+++ b/stable/awsebscsiprovisioner/values.yaml
@@ -33,12 +33,12 @@ provisioner:
   enableVolumeScheduling: false
   image:
     repository: "quay.io/k8scsi/csi-provisioner"
-    tag: "v1.1.0"
+    tag: "v1.2.0"
 
 attacher:
   image:
     repository: "quay.io/k8scsi/csi-attacher"
-    tag: "v1.1.0"
+    tag: "v1.1.1"
 
 snapshotter:
   # True if enable volume snapshot
--- a/stable/awsebscsiprovisioner/Chart.yaml
+++ b/stable/awsebscsiprovisioner/Chart.yaml
@@ -5,7 +5,7 @@ name: awsebscsiprovisioner
 maintainers:
   - name: alejandroEsc
   - name: gpaul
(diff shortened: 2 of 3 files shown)


stable/awsebscsiprovisioner/values.yaml
```diff
@@ -33,12 +33,12 @@ provisioner:
   enableVolumeScheduling: false
   image:
     repository: "quay.io/k8scsi/csi-provisioner"
-    tag: "v1.1.0"
+    tag: "v1.2.0"
```
joejulian: Are these tags the versions that have been used in soak and the mini MWT today?

gpaul: That's a good question. I'll check - I got these values from the refactoring PRs diff.

alejandroEsc:
I got these values from the latest version in the official site/repo. Not saying things changed? but have they? You mind giving me an opportunity to check this PR before it gets merged?

stable/awsebscsiprovisioner/values.yaml
gpaul: According to https://github.com/mesosphere/konvoy-mwt/blob/6ee5b8b186aec7691389fd244ae46aca470db786/konvoymwt2-dr/cluster.yaml#L94 the existing tags were used in the mwt2 (i.e., are set in kubeaddons-configs v0.0.45).

From the comments on https://jira.mesosphere.com/browse/DCOS-56901, it would appear that the MWT susceptible to the same issue.

alejandroEsc:
We should probably make a note of this in some place so that we do not loose these facts again. Upstream the values are definitely not in lock-step.

hectorj2f:
@alejandroEsc the official repo for us is https://github.com/mesosphere/csi-driver-deployments :)

PR #74, opened 2019-07-24
tfa: create secret as hooks

dkoshkin:
Generating `extra-steps` secret with hooks.

### DIFF
--- a/staging/traefik-forward-auth/templates/hooks.yaml
+++ b/staging/traefik-forward-auth/templates/hooks.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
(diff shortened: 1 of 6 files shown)


staging/traefik-forward-auth/templates/hooks.yaml
```diff
@@ -0,0 +1,64 @@
+---
```
shaneutt: Since this is a staging chart, and we generally intend to upstream this do we (or don't we) want some of this functionality in this chart to be conditional rather than deploying the jobs by default?

dkoshkin: What is this password and secret used for? Do we expect users wanting to generate it?

PR #79, opened 2019-07-25
[fix] removing deps from ops-portal

alejandroEsc:
Goal is to remove last piece of dep on ops-portal-secret and allow CI to continue when removing this part. 

- Need to gather more information about how ops-portal, kommander, dex and konvoy-ui interact
- Should find a way to pass current username, generated either by ops-portal or konvoy-ui


alejandroEsc:
closing this as it will take a while to get back to it.

PR #80, opened 2019-07-25
stable/awsebscsiprovisioner: Fix topology support

nfnt:


stable/awsebscsiprovisioner/templates/statefulset.yaml
```diff
@@ -74,9 +74,7 @@ spec:
             - --provisioner=ebs.csi.aws.com
             - --csi-address=$(ADDRESS)
             - --v=5
-            {{- if .Values.enableVolumeScheduling }}
```
hectorj2f: this had to be `.Values.provisioner.enableVolumeScheduling`... that was the problem

nfnt: Makes sense. So it isn't any weird templating behavior. Great, I'll fix the value instead of removing it.

stable/awsebscsiprovisioner/templates/statefulset.yaml
```diff
@@ -74,7 +74,7 @@ spec:
...
+            {{- if .Values.provisioner.enableVolumeScheduling }}
```
gpaul: :man_facepalming: :laughing:

hectorj2f: yeah... a 💣

hectorj2f:
Please, rebase against the `dev` branch so we can get this in asap :)

nfnt:
Oh, I thought this was based on `dev`. My local branch was probably outdated. Fixed.

PR #84, opened 2019-07-26
localvolumeprovisioner: allow configuration of multiple storage classes

nfnt:
Multiple storage classes allow defining a volume topology. For each item
in the configuration a `localvolumeprovisioner-%name%` storage class
will get created. Volumes mounted at `/mnt/%name%` on each host will be
detected and added as persistent volumes under this storage class.

### DIFF
--- a/stable/localvolumeprovisioner/Chart.yaml
+++ b/stable/localvolumeprovisioner/Chart.yaml
@@ -3,6 +3,6 @@ name: localvolumeprovisioner
 home: https://github.com/mesosphere/charts
 appVersion: "1.0"
 description: Local persistent volume provisioner for konvoy
-version: 0.2.2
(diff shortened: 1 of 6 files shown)


nfnt:
Did not test this configuration yet, will add the `ready` label once that's done.

nfnt:
Tested locally (by installing this chart manually via `helm`).

Noticed
```
Warning  FailedCreate  70s (x16 over 3m54s)  daemonset-controller  Error creating: pods "local-volume-provisioner-" is forbidden: pods with system-node-critical priorityClass is not permitted in kubeaddons namespace
```
on Konvoy 0.6.0. I'm not in the loop regarding these changes, this chart's default values probably need to be updated, either here or in `kubeaddons-configs`.

PR #86, opened 2019-07-27
awsebscsiprovisioner: set allowed topologies values

hectorj2f:
Customers might want to configure the allowed topologies when deploying on a multi AZ cluster:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    kubernetes.io/description: AWS EBS CSI provisioner StorageClass
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2019-07-27T14:12:39Z"
  name: awsebscsiprovisioner
  resourceVersion: "2043"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/awsebscsiprovisioner
  uid: 684d5b49-6704-4ef9-98d3-01e9c137bb45
parameters:
  type: gp2
provisioner: ebs.csi.aws.com
allowedTopologies:
- matchLabelExpressions:
  - key: topology.ebs.csi.aws.com/zone
    values:
    - us-west-2a
    - us-west-2b
    - us-west-2c
reclaimPolicy: Delete
```

### DIFF
--- a/stable/awsebscsiprovisioner/Chart.yaml
+++ b/stable/awsebscsiprovisioner/Chart.yaml
@@ -5,7 +5,8 @@ name: awsebscsiprovisioner
 maintainers:
   - name: alejandroEsc
   - name: gpaul
-version: 0.2.8
+  - name: hectorj2f
+version: 0.2.9
 kubeVersion: ">=1.13.0"
 home: https://github.com/kubernetes-sigs/aws-ebs-csi-driver
 sources:
--- a/stable/awsebscsiprovisioner/templates/storageclass.yaml
+++ b/stable/awsebscsiprovisioner/templates/storageclass.yaml
@@ -21,3 +21,7 @@ parameters:
   {{- if .Values.storageclass.encrypted }}
   encrypted: {{ .Values.storageclass.encrypted | quote }}
   {{- end }}
+{{- with .Values.storageclass.allowedTopologies }}
+allowedTopologies:
+{{ toYaml . }}
+{{- end }}
--- a/stable/awsebscsiprovisioner/templates/storageclass.yaml
+++ b/stable/awsebscsiprovisioner/templates/storageclass.yaml
@@ -21,7 +21,9 @@ parameters:
   {{- if .Values.storageclass.encrypted }}
   encrypted: {{ .Values.storageclass.encrypted | quote }}
   {{- end }}
+{{- if .Values.storageclass.allowedTopologies }}
 {{- with .Values.storageclass.allowedTopologies }}
 allowedTopologies:
 {{ toYaml . }}
+{{- end }}
 {{- end }}
--- a/stable/awsebscsiprovisioner/values.yaml
+++ b/stable/awsebscsiprovisioner/values.yaml
@@ -66,3 +66,10 @@ storageclass:
   reclaimPolicy: Delete
   volumeBindingMode: WaitForFirstConsumer
   type: gp2
+  allowedTopologies: []
+  # - matchLabelExpressions:
+  #   - key: topology.ebs.csi.aws.com/zone
+  #     values:
+  #     - us-west-2a
+  #     - us-west-2b
+  #     - us-west-2c


stable/awsebscsiprovisioner/Chart.yaml
```diff
@@ -5,7 +5,8 @@ name: awsebscsiprovisioner
 maintainers:
   - name: alejandroEsc
   - name: gpaul
-version: 0.2.8
```
nfnt: Please rebase on version 0.3.0.

stable/awsebscsiprovisioner/templates/storageclass.yaml
```diff
@@ -21,3 +21,7 @@ parameters:
   {{- if .Values.storageclass.encrypted }}
   encrypted: {{ .Values.storageclass.encrypted | quote }}
   {{- end }}
+{{- with .Values.storageclass.allowedTopologies }}
+allowedTopologies:
+{{ toYaml . }}
```
gpaul: Just checking, should this line be indented?

stable/awsebscsiprovisioner/templates/storageclass.yaml
```diff
@@ -21,3 +21,7 @@ parameters:
   {{- if .Values.storageclass.encrypted }}
   encrypted: {{ .Values.storageclass.encrypted | quote }}
   {{- end }}
+{{- with .Values.storageclass.allowedTopologies }}
+allowedTopologies:
```
nfnt: Let's allow a user to not set any allowed topologies at all. This would be the default case for on-prem. I.e., surround this with an if-cluase
```
{-if .Values.storageclass.allowedTopologies }}
allowedTopologies:
  [...]
{{- end }}
```

stable/awsebscsiprovisioner/values.yaml
```diff
@@ -66,3 +66,10 @@ storageclass:
   reclaimPolicy: Delete
   volumeBindingMode: WaitForFirstConsumer
```
nfnt: I'm no expert in topologies, but according to https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies it isn't necessary to set allowed topologies in the `WaitForFirstConsumer` case. Is this added to support the `Immediate` case as well?

stable/awsebscsiprovisioner/Chart.yaml
hectorj2f: that is not truth https://github.com/mesosphere/charts/blob/dev/stable/awsebscsiprovisioner/Chart.yaml#L8 :)

nfnt: You're right, I looked at the wrong chart :)

stable/awsebscsiprovisioner/templates/storageclass.yaml
hectorj2f: It'll have the same behavior but I can still add it :)

stable/awsebscsiprovisioner/templates/storageclass.yaml
hectorj2f: The content will be an array of `topologySelectorTerms` which should not be indented.
``` 
allowedTopologies:
- matchLabelExpressions:
  - key: topology.ebs.csi.aws.com/zone
     values:
     - us-west-2a
     - us-west-2b
     - us-west-2c
- nodeSelector:
  ...
```

stable/awsebscsiprovisioner/values.yaml
hectorj2f: There is some confusion because one is not defining any restrictions and assumes the volume can be mounted whenever a pod asks for it, and or relying on whether the pod already defined restrictions.
 
Meanwhile when you use allowedTopologies, it is you who set these volume affinity policies to any pod that uses this storage class.  One doesn't exclude the other.

hectorj2f: As an example, WaitForFirstConsumer without topologies will set the AZ to the AZ of the instance that asked for the pod. However I might not want to have that behavior and restrict that scenario to a specific or multiple AZ different than the one defined by the instance that claimed that vol. Therefore you can add allowedTopologies and awsebssi will pick (that is very unfortunate) the [0] AZ from that list.

hectorj2f: So the pod will be forced to be scheduled in an AZ where it can use the volumes by specified in the allowedTopologies

stable/awsebscsiprovisioner/templates/storageclass.yaml
hectorj2f: Done

PR #90, opened 2019-07-30
localvolumeprovisioner: remove hardcoded ns and comment priorityclass

hectorj2f:


stable/localvolumeprovisioner/templates/daemonset.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       tolerations:
 {{ toYaml .Values.tolerations | indent 8 }}
 {{- end }}
-      priorityClassName: {{ .Values.priorityClassName }}
+      # priorityClassName: {{ .Values.priorityClassName }}
```
shaneutt: Should we leave this in rather than comment it out, just with the understanding for now that it needs to be `kube-system` deployed?

stable/localvolumeprovisioner/values.yaml
```diff
@@ -9,7 +9,8 @@ storageclasses:
     reclaimPolicy: Delete
     volumeBindingMode: WaitForFirstConsumer
 
-priorityClassName: system-node-critical
```
shaneutt: Should we leave this?

PR #94, opened 2019-07-31
opsportal: merge konvoy ui in opsportal

hectorj2f:
I am testing the usage of dependencies in the charts to merge ops-portal and kommander


hectorj2f:
It seems we need to add a charts directory with kommander as part of this chart 🤔.

hectorj2f:
@joejulian  it seems it requires the chart to be part of the helm charts. Otherwise it fails to run `make all`. That is what I get after running the dependency command.

```
/usr/local/bin/helm --home /var/folders/b3/gql3fhrn2_x4h91gnhh817nw0000gn/T/tmp.9PfcjRJQ/.helm package stable/opsportal -d docs/stable
Error: found in requirements.yaml, but missing in charts/ directory: kommander
```

hectorj2f:
@jieyu I find easier and better to maintain a tarball created based on the specified version which lives in our repository. IF we copy the templates we will need to maintain these templates in two places, we already make mistakes with one single set of templates. I cannot imagine when having these templates copied here too.

Every time we decided to use a newer version of kommander for konvoy that will be generated and the tarball will be updated. 

By the way, this tarball is generated and packaged by helm in order to use dependencies. It isn't me doing the tarball. You need to run `helm update dependencies`  and that will get the tarball available in the specified repository.

joejulian:
I just checked the upstream charts repo. Including the tarball is common.

jieyu:
@hectorj2f @joejulian ok, thanks for the explanation! sg

stable/opsportal/values.yaml
```diff
@@ -49,3 +49,55 @@ opsportal:
...
+clusterPollingInterval: 3000
+clientPollingInterval: 3000
+
+# Mode must be either production|konvoy
+mode: konvoy
+
+displayName: Konvoy Cluster
+
+# Konvoy ui should NOT create the CRD if Kommander already did it
+createObservableClusterCRD: true
```
s12chung: hmm... can we take some of the configs `konvoy-ui` in `kubeaddons-configs` to be the defaults here?

hectorj2f: I think I took everything it is there 🤔  ... please let me know if I missed any value

s12chung: `Konvoy ui should NOT create the CRD if Kommander already did it` this comment was misleading. it seems like you have all the values.

s12chung: let's remove that coment.

stable/opsportal/values.yaml
```diff
@@ -49,3 +49,55 @@ opsportal:
           serviceName: traefik-kubeaddons-dashboard
           servicePort: 80
         path: /ops/portal/traefik
+
+
+###############################################################################
+#  Konvoy UI
+###############################################################################
+
+image:
```
s12chung: just realized because it's a subchart, don't you need to put all `konvoy-ui` values like such?

```yaml
kommander:
  image:
    tag: 1.60.0
    pullPolicy: IfNotPresent
  replicas: 1
# ....

  mode: konvoy
# ....
```

s12chung: example: https://github.com/helm/charts/blob/9fa4db5f3f2981fc56d5e54b9f551a9c86fdfbe0/stable/prometheus-operator/values.yaml#L351

hectorj2f: but we want to have the konvoy-ui defaults set here

stable/opsportal/values.yaml
hectorj2f: I took the defaults from kubeaddons-configs

hectorj2f: is only the comment something we want to remove ? or there is anything else... please do a more specific review and point the values and properties. We want to have the defaults from kubeaddons-configs here, afaik.

stable/opsportal/values.yaml
s12chung:
```
# i think you need to put `kommander` here because the subchart is "kommader"
# it's basically saying, "apply the values under `kommander` to the `kommander` chart
kommander
  image:
    tag: 1.60.0
    pullPolicy: IfNotPresent
  replicas: 1
# ....

  mode: konvoy
# ....
```

in my example:
`grafana` is a subchart: https://github.com/helm/charts/blob/9fa4db5f3f2981fc56d5e54b9f551a9c86fdfbe0/stable/prometheus-operator/requirements.yaml#L13

all the values are under `grafana`: https://github.com/helm/charts/blob/9fa4db5f3f2981fc56d5e54b9f551a9c86fdfbe0/stable/prometheus-operator/values.yaml#L351

i don't know where `enabled` is from in the example, may need to investigate.

stable/opsportal/values.yaml
s12chung: think that's all we need to remove.

hectorj2f:
This chart is working as expected, I just tested with konvoy/kubeaddons/kubeaddons-configs . So it should be safe to merge it

stable/opsportal/values.yaml
```diff
@@ -49,3 +49,54 @@ opsportal:
...
+
+  extraInitContainers:
+
+  resources:
+    requests:
+      memory: "128Mi"
+      cpu: "100m"
+    limits:
+      memory: "512Mi"
+      cpu: "2000m"
```
shaneutt: nonblocking comment: do we have behavioral information that indicates it should be allowed to burst this high?

PR #97, opened 2019-07-31
opsportal: add pod resource:

dkoshkin:
Added ability to provide `resources:` in opsportal.


hectorj2f:
good catch! we might miss this in other charts

PR #101, opened 2019-08-02
chore: make kommander idempotent

alejandroEsc:
make kommander idempontent by recreating SA if reinstalling also make it obvious what items are hooks by moving that SA creation to its own file


alejandroEsc:
tested this on my local kind cluster, works as expected.

PR #103, opened 2019-08-02
[kommander] fix: removed the hook annotation to SA

alejandroEsc:
Making the SA a helm hook does not seem to have any advantage and keeps the SA from being deleted properly.

The chart changes have been tested manually.


alejandroEsc:
Closing this: The issue and related reasons can be found here: https://github.com/helm/helm/issues/4884 I will update the documentation to include these details. However, this work around misses some lifecycle implications including deletion of resources etc..

PR #104, opened 2019-08-06
Add config for service labels to velero for service monitor discovery

gracedo:
- Commented out pod annotations because it looks like [prometheus-operator does not encourage annotation-based discovery of services anymore](https://github.com/helm/charts/tree/master/stable/prometheus-operator#prometheusioscrape) (in other charts, these annotations are commented out so following the pattern here).
- Added a `(.Capabilities.APIVersions.Has "monitoring.coreos.com/v1")` check in servicemonitor.yaml to avoid the case where velero's service monitor is enabled but the prometheus-operator isn't or has not yet deployed the `ServiceMonitor` CRD.
- Added config option to add service labels to enable discovery by service monitors already defined in prometheus ([our prom service monitor defs here](https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheus.yaml#L46)) instead to reuse those monitors instead of defining a separate service monitor here. This is already a common pattern I've noticed in other charts, including some that we use e.g. [fluent-bit](https://github.com/helm/charts/blob/master/stable/fluent-bit/values.yaml#L21) - we use label matching on the service for prom discovery/scraping instead of enabling its own service monitor def

image

[these changes should be upstreamed]

### DIFF
--- a/staging/velero/values.yaml
+++ b/staging/velero/values.yaml
@@ -43,6 +43,8 @@ metrics:
     #   key: value
 
   # Pod annotations for Prometheus
+  # Commented out since the Prometheus Operator discourages annotation-based service discovery.
+  # See https://github.com/helm/charts/tree/master/stable/prometheus-operator#prometheusioscrape
   # podAnnotations:
   #   prometheus.io/scrape: "true"
   #   prometheus.io/port: "8085"
--- a/staging/velero/values.yaml
+++ b/staging/velero/values.yaml
@@ -38,11 +38,15 @@ metrics:
   enabled: false
   scrapeInterval: 30s
 
+  service:
+    # labels:
+    #   key: value
+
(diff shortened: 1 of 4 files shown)


staging/velero/values.yaml
```diff
@@ -38,11 +38,15 @@ metrics:
   enabled: false
   scrapeInterval: 30s
 
+  service:
+    # labels:
+    #   key: value
+
   # Pod annotations for Prometheus
-  podAnnotations:
```
shaneutt: We're planning on providing our staging changes to upstream charts in time. This will require some iterations on the current implementation in the future, so it would be helpful for future readers to add a comment here providing an explanation and the context from the PR as to why this is now commented out.

hectorj2f: +1

PR #105, opened 2019-08-07
Velero dashboard and alert rule

gracedo:
- Updated a few dashboards (removed unnecessary/broken links)
- Added dashboard for Velero
image

- Added an alert rule for Velero (60min look back window on rate of # backup failures  metric)
image

Lmk if there are any additional graphs or rules that could be useful

### DIFF
--- a/staging/prometheus-operator/templates/prometheus/rules/mesosphere-rules/velero-rules.yaml
+++ b/staging/prometheus-operator/templates/prometheus/rules/mesosphere-rules/velero-rules.yaml
@@ -1,0 +1,29 @@
+# Rules for alerting against velero
+{{- if and .Values.mesosphereResources.create .Values.mesosphereResources.rules.velero }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ printf "%s-%s" (include "prometheus-operator.fullname" .) "velero" | trunc 63 | trimSuffix "-" }}
+  labels:
+    app: {{ template "prometheus-operator.name" . }}
+{{ include "prometheus-operator.labels" . | indent 4 }}
+{{- if .Values.defaultRules.labels }}
+{{ toYaml .Values.defaultRules.labels | indent 4 }}
+{{- end }}
+{{- if .Values.defaultRules.annotations }}
+  annotations:
+{{ toYaml .Values.defaultRules.annotations | indent 4 }}
+{{- end }}
+spec:
+  groups:
+  - name: velero
+    rules:
+    # backup failed
+    - alert: veleroFailedBackup
+      annotations:
+        message: 'velero schedule "{{`{{ $labels.schedule }}`}}": backup failed.'
+      expr: rate(velero_backup_failure_total{schedule!=""}[60m]) > 0
+      for: 10m
+      labels:
+        severity: warning
+{{- end }}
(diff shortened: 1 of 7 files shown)


hectorj2f:
What would it happen if Velero os not enabled? Will we get this alert as a false positive?

staging/prometheus-operator/templates/prometheus/rules/mesosphere-rules/velero-rules.yaml
```diff
@@ -0,0 +1,29 @@
+# Rules for alerting against velero
+{{- if and .Values.mesosphereResources.create .Values.mesosphereResources.rules.velero }}
```
hectorj2f: Let's be sure, we don't fire this alert if Velero is not running in the cluster. Let's remember that customer could disable the Velero Addon at any time

staging/prometheus-operator/values.yaml
```diff
@@ -47,6 +47,7 @@ mesosphereResources:
   create: false
   rules:
     etcd: true
+    velero: true
```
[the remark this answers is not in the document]
s12chung: hmm........ if you're making that statement... then most of these should be defaulted to be false....... and then true on the kubeaddons-configs side---even better... would be extrasteps/kubeaddons to be handling this....

staging/prometheus-operator/templates/prometheus/rules/mesosphere-rules/velero-rules.yaml
hectorj2f: This should work even if velero is enabled at any time.

@shaneutt ideally, dashboards and alerts should be enabled if the addon is enabled at any time. But that is a bit unrealistic :/

s12chung: yeah... think it's unrealistic for now.....

hectorj2f: we could have an operator that watches changes to the addon list, and so it disable/enable these settings on prometheus :)

staging/prometheus-operator/values.yaml
```diff
@@ -60,6 +61,7 @@ mesosphereResources:
     prometheusoverview: true
     scheduler: true
     traefik: true
+    velero: true
```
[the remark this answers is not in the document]
gracedo: I think for dashboards, it's okay to default them to true since nothing will break if they're installed without the corresponding metrics. Until we figure out how to tie dashboards to addons, we have been defaulting all the dashboards to on for addons that are enabled in knovoy by default, which I believe velero falls under.

Different story for alert rules unfortunately, which may break/misfire if the metrics are missing :)

staging/prometheus-operator/templates/prometheus/rules/mesosphere-rules/velero-rules.yaml
gracedo: I've defaulted the rule to `false` here - I'm inclined to leave it defaulted to `false` on our end in kubeaddons-configs as well, and document somewhere that users can override it to `true` if they wish to turn on that rule for velero, with a caveat that they must turn it off if they ever disable velero.

Not ideal, but I think it's the best solution for now until we figure out how to dynamically configure dashboards/rules dependent on whether addons are enabled/disabled.

gracedo:
@hectorj2f @shaneutt mind taking another look? Per our discussion on slack, `default all addon dashboards to on; default all addon alert rules to off` is now reflected in these changes. I will leave the velero alert rule defaulted to off in kubeaddons-configs, and open up a PR against konvoy docs to add notes on how to enable it.

PR #108, opened 2019-08-22
update minio to use the official one

s12chung:
it's now available officially and fixes a bug
which required a fork earlier

### DIFF
--- a/staging/velero/templates/deployment.yaml
+++ b/staging/velero/templates/deployment.yaml
@@ -120,7 +120,7 @@ spec:
       initContainers:
         {{- with .Values.minioBackend }}
         - name: initialize-velero
-          image: mesosphere/kubeaddons-addon-initializer:v0.0.4
+          image: mesosphere/kubeaddons-addon-initializer:v0.0.5
           args: ["velero"]
           env:
           - name: "VELERO_MINIO_FALLBACK_SECRET_NAME"
--- a/staging/velero/templates/miniobackend.yaml
+++ b/staging/velero/templates/miniobackend.yaml
@@ -123,7 +123,7 @@ spec:
       serviceAccountName: minio-operator-sa
       containers:
       - name: minio-operator
-        image: mesosphere/kubeaddons-minio-operator:v0.0.2
+        image: minio/k8s-operator:1.0.1
 ---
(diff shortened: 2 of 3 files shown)


staging/velero/templates/miniobackend.yaml
```diff
@@ -123,7 +123,7 @@ spec:
       serviceAccountName: minio-operator-sa
       containers:
       - name: minio-operator
-        image: mesosphere/kubeaddons-minio-operator:v0.0.2
+        image: minio/k8s-operator:1.0.1
```
hectorj2f: can you share the bug that was fixed ? So we can understand why we are changing it :).

staging/velero/templates/deployment.yaml
```diff
@@ -120,7 +120,7 @@ spec:
       initContainers:
         {{- with .Values.minioBackend }}
         - name: initialize-velero
-          image: mesosphere/kubeaddons-addon-initializer:v0.0.4
+          image: mesosphere/kubeaddons-addon-initializer:v0.0.5
```
hectorj2f: what did it change at this side ?

s12chung: https://github.com/mesosphere/kubeaddons-extrasteps/pull/10 -- basically some fields in the new minio-opeator are gone now. and it updated names of some resources.

staging/velero/templates/miniobackend.yaml
s12chung: remember that tempfile problem we paired together on?

so i wrote this to copy the files to not make templfiles and make regular files instead:

https://github.com/s12chung/minio-operator/commit/50afb6c0b43bef4a89153c6087623bc17175b2b5#diff-3254677a7917c6c01f55212f86c57fbfL11

which points to this: https://github.com/s12chung/minio

staging/velero/templates/deployment.yaml
hectorj2f: ack

staging/velero/templates/miniobackend.yaml
hectorj2f: thanks, the explanation would help any reviewer to understand your changes ;)

PR #112, opened 2019-08-29
[staging/gatekeeper] Cert manager integration

jimmidyson:
Also adds more configurability of webhook rules.

### DIFF
--- a/staging/gatekeeper/README.md
+++ b/staging/gatekeeper/README.md
@@ -5,7 +5,7 @@ Kubernetes.
 
 ## Prerequisites
 
-- Kubernetes 1.15 (or newer) for validating and mutating webhook admission
+- Kubernetes 1.9 (or newer) for validating and mutating webhook admission
   controller support.
 
 ## Overview
--- a/staging/gatekeeper/README.md
+++ b/staging/gatekeeper/README.md
@@ -34,25 +34,23 @@ Gatekeeper.
 All configuration settings are contained and described in
 [values.yaml](values.yaml).
 
-| Parameter                                             | Description                                                                  | Default             |
-| ----------------------------------------------------- | ---------------------------------------------------------------------------- | ------------------- |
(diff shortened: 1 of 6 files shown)


staging/gatekeeper/README.md
```diff
@@ -5,7 +5,7 @@ Kubernetes.
 
 ## Prerequisites
 
-- Kubernetes 1.15 (or newer) for validating and mutating webhook admission
+- Kubernetes 1.9 (or newer) for validating and mutating webhook admission
```
hectorj2f: 1.9 ? do we support that version in any of our products 🤔 ?

staging/gatekeeper/README.md
```diff
@@ -34,25 +34,23 @@ Gatekeeper.
...
-| admissionControllerNamespaceSelector.matchExpressions | Admission controller namespace selector expressions                          | []                  |
-| admissionControllerObjectSelector.matchExpressions    | Admission controller object selector expressions                             | []                  |
-| admissionControllerObjectSelector.matchLabels         | Admission controller object label selector                                   | []                  |
-| webhook.apiGroups                                     | Webhook rules API groups                                                     | [""]                |
-| webhook.apiVersions                                   | Webhook rules API versions                                                   | ["*"]               |
-| webhook.operations                                    | Webhook rules operations                                                     | ["CREATE","UPDATE"] |
-| webhook.resources                                     | Webhook rules resources                                                      | ["pods"]            |
```
hectorj2f: are we removing these properties on purpose ?

staging/gatekeeper/README.md
jimmidyson: This was a request from helm upstream to support it because it does work with that.

staging/gatekeeper/README.md
jimmidyson: Yes, changed to be more flexible by allowing multiple rules to be defined.

jimmidyson:
Chart version bumped.

PR #115, opened 2019-08-30
localvolumeprovisioner: allow configuration of multiple storage classes

nfnt:
Multiple storage classes allow defining a volume topology. For each item
in the configuration a 'localvolumeprovisioner-%name%' storage class
will get created. Volumes mounted at '/mnt/%name%' on each host will be
detected and added as persistent volumes under this storage class.

### DIFF
--- a/stable/localvolumeprovisioner/templates/configmap.yaml
+++ b/stable/localvolumeprovisioner/templates/configmap.yaml
@@ -6,11 +6,13 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   storageClassMap: |
-    localvolumeprovisioner:
-       hostDir: /mnt/disks
-       mountDir: /mnt/disks
-       blockCleanerCommand:
-         - "/scripts/shred.sh"
-         - "2"
-       volumeMode: Filesystem
-       fsType: ext4
+    {{- range .Values.storageclasses }}
+    localvolumeprovisioner-{{ .name }}:
+      hostDir: /mnt/{{ .name }}
+      mountDir: /mnt/{{ .name }}
+      blockCleanerCommand:
+       - "/scripts/shred.sh"
+       - "2"
+      volumeMode: Filesystem
+      fsType: ext4
+    {{- end }}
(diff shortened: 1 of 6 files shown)


stable/localvolumeprovisioner/values.yaml
```diff
@@ -3,10 +3,11 @@ image:
   tag: "v2.3.2"
   pullPolicy: "IfNotPresent"
 
-storageclass:
-  isDefault: true
-  reclaimPolicy: Delete
-  volumeBindingMode: WaitForFirstConsumer
+storageclasses:
```
shaneutt: I feel like this would be a great place to add some comments documenting how this configuration can be used and what type of problems it's meant to solve (e.g. the ability to have more than one storageclass deployed by localvolumeprovisioner).

nfnt: Good point! I'll add a comment here.

stable/localvolumeprovisioner/templates/configmap.yaml
```diff
@@ -6,11 +6,13 @@ metadata:
...
   storageClassMap: |
-    localvolumeprovisioner:
-       hostDir: /mnt/disks
-       mountDir: /mnt/disks
-       blockCleanerCommand:
-         - "/scripts/shred.sh"
-         - "2"
-       volumeMode: Filesystem
-       fsType: ext4
+    {{- range .Values.storageclasses }}
```
hectorj2f: I understand we have tested this change. 
What would it happen with the old storage class `localvolumeprovisioner` ? what will it happen with the existing pods using the storage class ?

nfnt: Good point, I'll change this to make both provisioner name as well as mount dir name configurable. With that we can provide a default configuration with storage class `localvolumeprovisioner` and mount dir `/mnt/disks` that's backwards-compatible with existing deployments.

nfnt: Fixed in https://github.com/mesosphere/charts/pull/118.

PR #116, opened 2019-09-02
feat: remove dependency on observableCluster-CRD for konvoy-ui

pierrebeitz:
Now that Konvoy-UI can be built separately from Kommander, let's make use of that by giving it an own chart. There's no need for the ObservableClusters-CRD anymore in the context of Konvoy-UI. 🎉

@shaneutt, @hectorj2f please have a thorough look, as this is the first time i wrote a chart. There's definitely the need to update the `requirements.lock` of `opsportal`.

How to make sure this gets deployed? Is there anything the frontend-folks can do to test/debug this further?

### DIFF
--- a/stable/opsportal/requirements.yaml
+++ b/stable/opsportal/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
-- name: kommander
-  version: 0.1.22
+- name: konvoy-ui
+  version: 0.1.0
   repository: https://mesosphere.github.io/charts/stable
--- a/stable/opsportal/values.yaml
+++ b/stable/opsportal/values.yaml
@@ -70,20 +70,15 @@ opsportal:
 ###############################################################################
 kommander:
   image:
-    repository: mesosphere/kommander
-    tag: 1.60.0
+    repository: mesosphere/konvoy-ui
+    tag: 1.65.30
     pullPolicy: IfNotPresent
   replicas: 1
   logoutRedirectPath: /ops/landing
   clusterPollingInterval: 3000
   clientPollingInterval: 3000
 
-  # Mode must be either production|konvoy
-  mode: konvoy
-
   displayName: Konvoy Cluster
-
-  createObservableClusterCRD: true
 
   extraInitContainers:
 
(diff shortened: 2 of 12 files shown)


stable/opsportal/requirements.yaml
```diff
@@ -1,4 +1,4 @@
 dependencies:
-- name: kommander
-  version: 0.1.22
+- name: konvoy-ui
+  version: 0.0.1
```
hectorj2f: Please, set `0.1.0`

DanielMSchmidt:
Hey @hectorj2f So, the reason why we need this is that kommander and konvoy-ui are now to separate, but interconnected code bases instead of a single product with a feature flag turned on and off. I agree with you that we have quite some duplicated code here, our thinking was that we rather have some duplication than a wrong abstraction. In the future kommander and konvoy-ui could become vastly different in some aspects of their deployment. If you say we keep the charts together that is fine, but it could cause confusion down the road IMHO

stable/opsportal/requirements.yaml
DanielMSchmidt: Addressed 👍

pierrebeitz:
@hectorj2f what's the status here?

hectorj2f:
@DanielMSchmidt Thanks for the clarification. Let's then proceed with the changes. We need to test these new changes don't affect running clusters.

stable/opsportal/values.yaml
```diff
@@ -68,23 +68,18 @@ opsportal:
 ###############################################################################
 #  Konvoy UI
 ###############################################################################
-kommander:
```
hectorj2f: I have to say, this change might cause some issues to running clusters :/. 

Scenario: Imagine we have a cluster where the customer changed the values property for the opsportal addon.
```
 kommander:
   yada: 
   yada
```

With this rename, the new changes might not be applied to konvoy-ui. Can we double check that ?

hectorj2f: Otherwise, we should inform the customers that they need to update the `values: ` for opsportal from kommander to konvoy-ui.

hectorj2f:
please, rebase against dev

pierrebeitz:
thanks for the review! i rebased against `dev` and changed back to use `kommander`, so customers won't have issues.

pierrebeitz:
closing until it becomes urgent/important

PR #119, opened 2019-09-09
adding kubefed as a subchart to kommander

alejandroEsc:
- kubefed is created as a requirement for kommander
- made changes to include creating roles, rolebindings and addons federated type

to test run:
```helm install stable/kommander --name kommander --namespace kommander```

### CODE REVIEW

File: stable/kommander/templates/federated-addons.yaml [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+    pluralName: addons
+    scope: Namespaced
+    version: v1alpha1
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  generation: 1
+  name: federatedaddons.types.kubefed.io
```
joejulian: Should this be in our own domain space?

stable/kommander/templates/federated-rolebindings.yaml
```diff
@@ -0,0 +1,160 @@
...
+    kind: RoleBinding
+    pluralName: rolebindings
+    scope: Namespaced
+    version: v1
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  generation: 1
+  name: federatedrolebindings.types.kubefed.io
```
joejulian: Here, too?

stable/kommander/templates/federated-roles.yaml
```diff
@@ -0,0 +1,160 @@
...
+    kind: Role
+    pluralName: roles
+    scope: Namespaced
+    version: v1
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  generation: 1
+  name: federatedroles.types.kubefed.io
```
joejulian: "

stable/kommander/templates/federated-rolebindings.yaml
alejandroEsc: Sorry, not sure I understand?

stable/kommander/templates/federated-addons.yaml
joejulian: Typically, most CRDs are named for the organization that's responsible for maintaining it, which would imply that `federatedaddons.types.kubefed.io` would be maintained by kubernets/sig-multicluster. Since, afaict, that's our own CRD, it seems to me it should be suffixed with `mesosphere.io` (or maybe even `d2iq.io`. Not sure if the rest are ever going to be changed since that's a backward compatibility problem.). But then again, maybe there's a reason they have to be suffixed with `kubefed.io` that I'm not aware of.

alejandroEsc: Oh I understand now, thanks for clarifying! This isn't something we own, we are enabling a federated resource with these files, we aren't creating a crd, it is confusing admittedly.

stable/kommander/values.yaml
```diff
@@ -42,3 +42,40 @@ ingress:
...
+# in order to override values.
+kubefed:
+  controllermanager:
+    enabled: true
+    # override value is necessary due to a "featureGates":interface {}(nil)... spec.featureGates in body must be of type array: "null"
+    # bug found
+    featureGates:
+      CrossClusterServiceDiscovery: "Enabled"
+      FederatedIngress: "Enabled"
+      PushReconciler: "Enabled"
```
jimmidyson: I wonder if, for now, we just enable `PushReconciler` as that's all we're actually using?

stable/kommander/values.yaml
```diff
@@ -42,3 +42,40 @@ ingress:
...
+
+federate:
+  addons: true
+  roles: true
+  rolebindings: true
+
+# This name must match the name of the kubefed dep
+# in order to override values.
+kubefed:
+  controllermanager:
```
jimmidyson: Do we need to override all these values? Looking at https://github.com/kubernetes-sigs/kubefed/blob/v0.1.0-rc6/charts/kubefed/README.md#configuration, the only non-default I can see are the `featureGates` (if you agree we disable everything other than `PushReconciler`)?

s12chung: i agree we should not override.

stable/kommander/values.yaml
s12chung: i agree.

stable/kommander/templates/federated-addons.yaml
s12chung: not sure if it'll break kubefed if we change the domain space. these are generated via kubefed, which needs to follow kubefed specs. so i think it's ok to kubefed to own it.

stable/kommander/values.yaml
alejandroEsc: Should not? ... Anyway, the reason for overriding in cases is to expose these values to kubeaddons in the future, if we so wish. For the featureGates we have to override as there is a problem with how null results are interpreted in kubefed. I will override these falues and leave only pushreconciler to enabled, but the rest have to be set to false explicitly.

stable/kommander/values.yaml
alejandroEsc: sure, but not sure the benefit/drawbacks of this.

PR #120, opened 2019-09-10
Added Nvidia driver container and nvidia-container-runtime.

Gilbert88:
Added Nvidia Chart, which includes the Nvidia Driver and Nvidia device plugin. Both depend on Nvidia Container Runtime and Nvidia GPU hardware, while the driver has extra dependencies on some kernel modules.


shaneutt:
Hey @Gilbert88 just wanted to check in with you on this. Anything you need from us to assist or get you unblocked on this one? Let us know how we can help! :+1:

shaneutt:
Just checking in since it's been a while, need any help with anything @Gilbert88 ?

Gilbert88:
@shaneutt Thanks for following up with the PR. Sorry I missed your comments above. Was focusing on the low level runtime dependencies that we have to commit first on the platform.

Two quick questions:

- I saw the current charts are using some general stable image versions (e.g., with image tags as 1.0.3 etc). For Nvidia, the driver has to interact with some low-level kernel modules and build some hardware libraries during container runtime. So unavoidably, Nvidia has to build images per OS distribution (also per specific kernel patch version). Is there any place in the `Charts` repo that we can introduce extra logics (e.g., a script) for OS detection or kernel version check, before deciding which image tag to use?

- Post chart installation labeling. I expect to do some `kubectl label node ......` after the chart becomes running state, because the labeling information comes from the nvidia device plugin (e.g., identifying the GPU model, version etc). Probably I do not understand how to define a chart deep enough. Does the `charts` or `kubeaddons-config` supports this case?

Thank you!

cc @joejulian

shaneutt:
> I saw the current charts are using some general stable image versions (e.g., with image tags as 1.0.3 etc). For Nvidia, the driver has to interact with some low-level kernel modules and build some hardware libraries during container runtime. So unavoidably, Nvidia has to build images per OS distribution (also per specific kernel patch version). Is there any place in the Charts repo that we can introduce extra logics (e.g., a script) for OS detection or kernel version check, before deciding which image tag to use?

We can provide different configurations in the manifest templates based on end-user input: e.g. if you can provide those things up front we can provide logic that will generate the appropriate template.

However if you mean to ask if we can dynamically adjust for that post-deployment of the chart, the closest thing we have for that kind of functionality would be [extrasteps](https://github.com/mesosphere/kubeaddons-extrasteps) wherein we could run an init-container which could inspect the underlying system and self-adjust its own deployment based on it's findings.

This is a bit of an unusual case for extrasteps in that we generally use it to manage secrets post-deployment, but it might be our best bet in this case for a quick solution.

> Post chart installation labeling. I expect to do some kubectl label node ...... after the chart becomes running state, because the labeling information comes from the nvidia device plugin (e.g., identifying the GPU model, version etc). Probably I do not understand how to define a chart deep enough. Does the charts or kubeaddons-config supports this case?

This can be supported fairly cleanly by [extrapsteps](https://github.com/mesosphere/kubeaddons-extrasteps) :+1: 

Let us know if you need anything else!

staging/nvidia/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+```bash
+$ helm install staging/nvidia --name nvidia
+```
+
+## Uninstalling the Chart
+
+```bash
+$ helm delete nvidia
+```
```
shaneutt: docs++

staging/nvidia/charts/nvidia-device-plugin/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: nvidia-device-plugin
+home: https://github.com/mesosphere/charts
+appVersion: "1.0"
+description: Nvidia Device Plugin for konvoy
+version: 0.1.0
+maintainers:
+  - name: gilbert
```
shaneutt:
```suggestion
  - name: gilbert88
```

staging/nvidia/charts/nvidia-device-plugin/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+{{ include "nvidia-device-plugin.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args:  ['{{ include "nvidia-device-plugin.fullname" . }}:{{ .Values.service.port }}']
+  restartPolicy: Never
```
shaneutt: tests++

staging/nvidia/charts/nvidia-device-plugin/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+tolerations:
+  # Allow this pod to be rescheduled while the node is in "critical add-ons only" mode.
+  # This, along with the annotation above marks this pod as a critical add-on.
+  - key: CriticalAddonsOnly
+    operator: Exists
+  - key: nvidia.com/gpu
+    operator: Exists
+    effect: NoSchedule
+
+affinity: {}
```
shaneutt: I would recommend adding support for `initContainers` here and now so we don't have to make a PR for it later if needed, and to support more versatility for end-users if they have some other tasks they need to run in init before deploying the addon.

```suggestion
affinity: {}

initContainers: []
```

This will require a change in the `Daemonset` to support this as well.

staging/nvidia/charts/nvidia-driver/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+{{ include "nvidia-driver.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args:  ['{{ include "nvidia-driver.fullname" . }}:{{ .Values.service.port }}']
+  restartPolicy: Never
```
shaneutt: tests++ :tada:

staging/nvidia/charts/nvidia-driver/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+nodeSelector:
+  konvoy.mesosphere.com/gpu-provider: NVIDIA
+
+tolerations: []
+
+affinity: {}
```
shaneutt: Again here I think we should just support initContainers on day 1 as it can add a lot of value for us and end-users right out of the box:
```suggestion
affinity: {}

initContainers: []
```

And it would need to be added to the `Daemonset` here too.

staging/nvidia/README.md
Gilbert88: I have added the basic doc for this chart. However, more change on the document may be needed in the near future because Nvidia may release a new driver with additional features supported, as well as a controller.

staging/nvidia/charts/nvidia-device-plugin/templates/tests/test-connection.yaml
Gilbert88: I was following the test from gatekeeper, kube-oidc-proxy, and seems like some other chart (e.g., knative) does not come with a test yet. Could we add a test in a separate PR given the GPU functionality is tested with some CUDA image?

shaneutt: Oh sorry, these comments are just meant as :+1: type comments :joy:

staging/nvidia/charts/nvidia-device-plugin/templates/daemonset.yaml
```diff
@@ -0,0 +1,45 @@
...
+    metadata:
+      # Mark this pod as a critical add-on; when enabled, the critical add-on scheduler
+      # reserves resources for critical add-on pods so that they can be rescheduled after
+      # a failure.  This annotation works in tandem with the toleration below.
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ""
+      labels:
+        name: nvidia-device-plugin-ds
+    spec:
+{{- if .Values.initContainers }}
```
shaneutt: Awesome :+1: ty

staging/nvidia/charts/nvidia-device-plugin/values.yaml
```diff
@@ -0,0 +1,60 @@
...
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
+
+resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
```
hectorj2f: we should set some default resources

hectorj2f: so we can adjust them in the next MWT

Gilbert88: I was following examples in this repo. Were they misbehaving previously? Or we want to make this enforcement starting from now?

- Daemonsets like `awscsiprovisioner`, `localvolumeprovisioner`.
- Other deployments that haven't set up a default resource yet - `dex-k8s-authenticator`, `kube-oidc-proxy`, `prometheus-node-exporter` and `traefik-forward-auth`

I am happy to figure out what is the appropriate default resources for the Nvidia Chart. Not sure if we can address it together with other charts

Gilbert88:
> @Gilbert88 Why do we have a whole chart as part of the nvidia chart ?

Because there is a dependency between those two sub-charts and running any single of them does not provide any value as an addon.

> We might need to set some default resource usage that can be later adjusted in the next MWT

Please see my comment above. Happy to add a default resources but it may take me sometime, because the nvidia driver container depends heavily on kernel and kernel module, and re-compile the driver during container runtime before it is ready for services  :)

staging/nvidia/charts/nvidia-device-plugin/values.yaml
hectorj2f: @Gilbert88 hold on! is this a critical daemon ? you are comparing it with the awscsiprovisioner, if so you are missing some properties such as `priorityClassName: system-node-critical`

`prometheus` gets his resource definitions in kubeaddons-configs. https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheusadapter.yaml#L32 https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheus.yaml#L182
https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/prometheus.yaml#L193

`kube-oidc-proxy` has been removed....
`dex-k8s-authenticator` https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/dex-k8s-authenticator.yaml#L54


`traefik-forward-auth` https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/traefik-forward-auth.yaml#L23

hectorj2f: Please, add some estimations in the kubeaddons-configs PR. Don't set them here if you prefer.

hectorj2f:
> Because there is a dependency between those two sub-charts and running any single of them does not provide any value as an addon.

hectorj2f:
@Gilbert88 please, rebase against `dev` branch and set some resources in the kubeaddons-configs PR. The rest LGTM 👍

staging/nvidia/charts/nvidia-device-plugin/values.yaml
Gilbert88: Spent a couple hours to test out different combinations of default resources. I believe the value in https://github.com/mesosphere/kubeaddons-configs/pull/212 will be close to what we expect.

PR #121, opened 2019-09-10
Update CONTRIBUTING.md to include extrasteps reference

shaneutt:


shaneutt:
As this wasn't a requirement of [DCOS-58504](https://jira.mesosphere.com/browse/DCOS-58504) I would like to pass on this for this PR and instead created [DCOS-58587](https://jira.mesosphere.com/browse/DCOS-58587) to follow up on this. There's some sensitivity around the structure of our CONTRIBUTING.md rules right now as more people within the org start to contribute, so I just wanna double check on that rather than add it here.

PR #123, opened 2019-09-12
Add cert-manager-setup chart, make dex + traefik use cert-manager issued certificates

timaa2k:


staging/cert-manager-setup/templates/post-install-hook-job.yaml
```diff
@@ -0,0 +1,43 @@
...
+  template:
+    metadata:
+      name: "sleep"
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: {{ .Chart.Name }}
+        image: ubuntu:xenial
+        imagePullPolicy: IfNotPresent
+        command: ["sleep", "30"]
```
hectorj2f: this is ugly :/ ... we need to check which resource we need to query to be sure the apiserver is ready. This could cause race conditions based on certain scenarios.

s12chung: if we need a sleep, it'll be better to create an `extra-step` to check the service directly. since you're within the cluster, it should be simple.

hectorj2f: Let's stop to confuse him ... using an extra-step again.

hectorj2f: what does your comment have to do with my comment about the `["sleep", "30"]` ?

s12chung: i suggested:

> you can replace `["kubectl", "wait", "--for=condition=Available", "--timeout=300s", "apiservice", "v1beta1.webhook.certmanager.k8s.io"]` to run the extra step binary instead.

i think you're sleeping for `30` seconds because you're not sure if the webhook is installed properly?

staging/cert-manager-setup/templates/crds.yaml
```diff
@@ -0,0 +1,7514 @@
+
```
s12chung: i'm curious, why aren't use using the official chart and putting this as a subchart?

timaa2k:
The official chart (stable in Helm) is no longer updated. The jetstack repo now contains the official latest version. This latest version requires to install CRDs manually before deploying with Helm. I've automated this step using the `crd-install` hook.
https://docs.cert-manager.io/en/release-0.10/getting-started/install/kubernetes.html#steps

staging/cert-manager-setup/templates/crds.yaml
timaa2k: Answer below :)

staging/cert-manager-setup/templates/post-install-hook-job.yaml
timaa2k: It appears as if the original wait doesn't wait long enough. If I/you/we find a command that waits for the right thing the sleep could be removed again.

hectorj2f: please, file a jira ticket for that

timaa2k:
Thanks @hectorj2f I figured what was missing and it passes linting now :)

staging/cert-manager-setup/templates/post-install-hook-job.yaml
timaa2k: https://jira.mesosphere.com/browse/DCOS-58871

hectorj2f:
The rest looks good to me despite of the necessity of finding a better solution for the `sleep`

hectorj2f:
We're going to adapt this PR according to the changes agreed in the initial design doc.

stable/dex/templates/certificate.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: dex
+  namespace: kubeaddons
```
hectorj2f: Please, use `{{ .Release.Namespace }}`

stable/dex/templates/certificate.yaml
```diff
@@ -0,0 +1,19 @@
...
+  annotations:
+    "helm.sh/hook": pre-install
+spec:
+  secretName: dex
+  issuerRef:
+    name: kubernetes-ca
+    kind: ClusterIssuer
+  commonName: dex
+  organization:
+    - Mesosphere
```
hectorj2f: D2iQ ?

stable/dex/templates/certificate.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: dex
+  namespace: kubeaddons
+  annotations:
+    "helm.sh/hook": pre-install
+spec:
+  secretName: dex
```
hectorj2f: This should be configurable too

stable/dex/templates/certificate.yaml
```diff
@@ -0,0 +1,19 @@
...
+  issuerRef:
+    name: kubernetes-ca
+    kind: ClusterIssuer
+  commonName: dex
+  organization:
+    - Mesosphere
+  dnsNames:
+    - dex-kubeaddons
```
hectorj2f: This should be configurable, let's add it to `values.yaml`

staging/cert-manager-setup/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: read-apiservices-rolebinding
+  namespace: cert-manager
```
hectorj2f: I'd rely on `{{ .Release.Namespace }}`

staging/cert-manager-setup/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+subjects:
+- kind: ServiceAccount
+  namespace: cert-manager
```
hectorj2f: {{ .Release.Namespace }}

staging/cert-manager-setup/templates/issuers.yaml
```diff
@@ -0,0 +1,42 @@
+{{ if .Values.clusterissuer }}
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Issuer
+metadata:
+  name: kubernetes-root-issuer
+  namespace: cert-manager
```
hectorj2f: {{ .Release.Namespace }}

staging/traefik/templates/certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+  annotations:
+    "helm.sh/hook": pre-install
+spec:
+  secretName: traefik-certificate
+  issuerRef:
+    name: kubernetes-ca
+    kind: ClusterIssuer
+  commonName: traefik.localhost.localdomain
+  organization:
+  - Mesosphere
```
hectorj2f: D2iQ ?

staging/traefik/templates/certificate.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: traefik
+  namespace: kubeaddons
```
hectorj2f: the same as above: `namespace: {{ .Release.Namespace }}`

staging/traefik/templates/deployment.yaml
```diff
@@ -174,7 +174,7 @@ spec:
       {{- if and .Values.ssl.enabled (not .Values.ssl.upstream) }}
       - name: ssl
         secret:
-          secretName: {{ template "traefik.fullname" . }}-default-cert
+          secretName: traefik-certificate
```
hectorj2f: IMHO we should enable to define this secret-name in `values.yaml` but this one is not a blocker

stable/dex/templates/certificate.yaml
timaa2k: https://jira.mesosphere.com/browse/DCOS-59044

staging/traefik/templates/deployment.yaml
timaa2k: https://jira.mesosphere.com/browse/DCOS-59044

PR #129, opened 2019-09-20
fix for master not being fetched already

joejulian:


joejulian:
Sorry, when I tested this I already had a fully fetched git clone.

PR #131, opened 2019-09-20
[kommander-cluster ] initial contribution

alejandroEsc:
Initial contribution for a chart of the Kommander lifecycle github repo.

This is very prelim and there is much work yet to be done.

### DIFF
--- a/stable/kommander-cluster/Chart.yaml
+++ b/stable/kommander-cluster/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: kommander-cluster
 home: https://github.com/mesosphere/kommander-cluster-lifecycle
 appVersion: "0.0.1"
-description: Kommander
+description: Kommander Clusters from (kommander cluster lifecycle)
 version: 0.0.1
 maintainers:
   - name: alejandroEsc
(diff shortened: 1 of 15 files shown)


jimmidyson:
Thank you for starting this! Can we please get this moved to cluster lifecycle repo and ultimately added as chart dependency of kommander chart? What do you think?

alejandroEsc:
> Thank you for starting this! Can we please get this moved to cluster lifecycle repo and ultimately added as chart dependency of kommander chart? What do you think?

I am not sure why it belong in cluster-lifecycle? Not saying I wont do it, just whats the reasoning exactly?

jimmidyson:
@alejandroEsc This chart is specifically for deploying cluster lifecycle components. Cluster lifecycle currently uses kustomize for generating manifests and for deploying in both testing and development, which means we would have two different ways to deploy cluster lifecycle (kustomize and this chart) which would need to keep in sync, etc. By putting the chart in the cluster lifecycle repo, we can use the chart during development and testing, and ensure the chart remains up to date with changes as they are made by using controller-gen updates files, rather than having to manually make changes in this chart repo. When new versions of cluster lifecycle are released, the kommander chart dependency can be updated to use the latest version.

alejandroEsc:
closing per request.

PR #138, opened 2019-09-25
stable/awsebscsi: fix role not being created and add env to values

dkoshkin:
Merging the same PR to `dev` instead of `master`.


hectorj2f:
we need to sign the commits :)

PR #140, opened 2019-09-26
fix: use latest version of kommander, disable kubefed

alejandroEsc:
Should allow us to disable kubefed and its resources when using kommander chart for this chart

- Requires: https://github.com/mesosphere/charts/pull/139


stable/opsportal/values.yaml
```diff
@@ -114,3 +115,6 @@ kommander:
       traefik.ingress.kubernetes.io/auth-type: forward
       traefik.ingress.kubernetes.io/auth-url: http://traefik-forward-auth-kubeaddons.kubeaddons.svc.cluster.local:4181/
       traefik.ingress.kubernetes.io/auth-response-headers: X-Forwarded-User
+
+  kubefed:
```
hectorj2f: why does opsportal install kubefed ? shouldn't it be kommander chart who does it ?

hectorj2f: nvm! This is for the kommander chart.

stable/opsportal/values.yaml
```diff
@@ -114,3 +115,13 @@ kommander:
...
       traefik.ingress.kubernetes.io/auth-response-headers: X-Forwarded-User
+
+  # disable kubefed to keep all items related to kubefed from being created
+  kubefed:
+    enabled: false
+
+###############################################################################
+#  Kubefed: Leave disabled as only kommander should deploy
+###############################################################################
+kubefed:
```
s12chung: I don't think this is needed because you already have it above?

s12chung: had a slack discussion. seems a bit weird that both definitions are needed but.... it is what it is.

PR #142, opened 2019-09-27
Update certificate with load balancer address

timaa2k:


staging/traefik/templates/init-cert-job.yaml
```diff
@@ -0,0 +1,31 @@
...
+spec:
+  template:
+    metadata:
+      name: "update-traefik-certificate"
+    spec:
+      serviceAccountName: {{ template "traefik.fullname" . }}-cert-init
+      restartPolicy: Never
+      containers:
+      - name: {{ .Chart.Name }}
+        image: timaa2k/kubeaddons-addon-initializer:v0.0.8
```
hectorj2f: we should expose this image to bump it and/or change it on deman

PR #148, opened 2019-09-30
dex: bump cert duration to 10 years

jieyu:
The default value is 90 days, which is way too short.  Until we figure
out a good cert rotation story, it's better to use a longer duration to
be safe.


hectorj2f:
We need a rebase here

hectorj2f:
I will take care of it

PR #149, opened 2019-09-30
traefik: bump cert duration to 10 years

jieyu:
The default value is 90 days, which is way too short. Until we figure
out a good cert rotation story, it's better to use a longer duration to
be safe.


mhrabovcin:
We're going to integrate this change in https://github.com/mesosphere/charts/pull/142

timaa2k:
We're integrating this change into the next traefik PR that we're preparing

hectorj2f:
@mhrabovcin @timaa2k Gotcha!

hectorj2f:
I am closing this PR, the bump in the cert duration has been done in https://github.com/mesosphere/charts/pull/142

PR #150, opened 2019-09-30
help fix not adding extra annotations to ingress type

alejandroEsc:


alejandroEsc:
tested this part with templating, works if empty map is added.

PR #151, opened 2019-10-01
Make cert-manager installation conditional

timaa2k:


timaa2k:
@jimmidyson @jieyu this PR presents a (manually tested) quickfix for the "cert-manager already installed on a cluster". PTAL and share your thoughts.

staging/cert-manager-setup/values.yaml
```diff
@@ -10,3 +10,13 @@ clusterissuer: {}
   # spec:
   #   ca:
   #     secretName: kubernetes-intermediate-ca
+
+# When installing addons on a cluster that already has cert-manager
```
shaneutt: docs++ :+1:

jieyu:
Why we need to change the dependency?

PR #155, opened 2019-10-02
Fix the upgrade problems for dex and traefik

jieyu:
Hook resources won't be tracked by Helm. As a result, we'll have to add
a special annotation so that it gets re-created during upgrade. See
more details in https://github.com/helm/helm/issues/5482.


jimmidyson:
Why does it need to be recreated? Why not just ignore on upgrade (wrap in conditional to check if in upgrade mode)?

jieyu:
@jimmidyson we need to add pre-upgrade because this hook is newly added, and we have to support the upgrade path from not having the cert to having this cert.

PR #156, opened 2019-10-02
[catalog] chore: update crds and roles

alejandroEsc:
Here we: 
- only create the CRD if it is missing, using crd hook logic as well.
- Create the proper roles for the controller to work
- we also declare the metrics container port.

- Tested locally


staging/kubeaddons-catalog/templates/crd_addonrepository.yaml
```diff
@@ -1,9 +1,14 @@
+{{- if not (.Capabilities.APIVersions.Has "kubeaddons.mesosphere.io/v1alpha1") }}
```
s12chung: what would happen if `Addon` is already installed? would `.Capabilities.APIVersions` have `"kubeaddons.mesosphere.io/v1alpha1"`?

alejandroEsc: This is part of the problem with how we have catalog as part of Kubeaddons. They have the same api group and as such require that both these CRDs are installed.

s12chung: just want to confirm. if `Addon` is installed. then `not (.Capabilities.APIVersions.Has "kubeaddons.mesosphere.io/v1alpha1")` will return `true`?

i'd assume it would return `false`.

PR #160, opened 2019-10-03
dex: add dex-controller as a subchart

jieyu:
Allow declarative management of Dex configurations using Kubernetes API.


hectorj2f:
@jieyu please rebase against `dev` branch

PR #171, opened 2019-10-04
Add jenkins helm chart to staging

kvish:


staging/jenkins/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+version: 1.7.3
+appVersion: lts
+description: Open source continuous integration server. It supports multiple SCM tools
+  including CVS, Subversion and Git. It can execute Apache Ant and Apache Maven-based
+  projects as well as arbitrary scripts.
+sources:
+- https://github.com/jenkinsci/jenkins
+- https://github.com/jenkinsci/docker-jnlp-slave
+- https://github.com/maorfr/kube-tasks
+- https://github.com/jenkinsci/configuration-as-code-plugin
```
vinodkone: do we want to put in some konvoy specific sources too here since we are adding things on top of the upstream chart?

staging/jenkins/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+name: jenkins
+home: https://jenkins.io/
+version: 1.7.3
```
vinodkone: is this the convention we are using in mesosphere charts, ie our charts have the same version as the upstream version?

staging/jenkins/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+appVersion: lts
```
vinodkone: "lts" seems very broad, dont we want to pin to a more explicit version?

staging/jenkins/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: Open source continuous integration server. It supports multiple SCM tools
```
vinodkone: This should probably say this is D2IQ's version of Jenkins which has been adapted to work on Konvoy?

staging/jenkins/Chart.yaml
vinodkone: i'm not sure if dependent charts are expected (as convention) to include sources of their parent charts.

staging/jenkins/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
shaneutt: I think it's important to note that `staging/` is meant to be used for charts that are going to be worked on here and then PR'd against upstream implementations. I just wanted to double check: is that the goal here?

staging/jenkins/Chart.yaml
shaneutt: No convention: we should increase the patch version here for our changes:
```suggestion
version: 1.7.4
```

staging/jenkins/Chart.yaml
shaneutt: :heavy_check_mark: if there's a semver compat release version available let's use that instead.

staging/jenkins/Chart.yaml
shaneutt: :heavy_plus_sign::heavy_plus_sign:

staging/jenkins/values.yaml
```diff
@@ -0,0 +1,19 @@
+master:
+  installPlugins: ["prometheus:2.0.6"]
+  adminPassword: "deleteme"
```
shaneutt: This is a good place to add a brief comment explaining how you expect the end-user to handle this password field.

For Kubeaddons one of the goals is for Addon resources to be "ready to go" without modification, while still being secure. For other services, we've used [initContainers](https://github.com/mesosphere/kubeaddons-extrasteps) as a mechanism by which we can generate strong persistent passwords for services pre-startup, would that be appropriate here as well?

staging/jenkins/Chart.yaml
kvish: we have no changes on the base tech, Shouldnt we keep the versions the same?

kvish:
@shaneutt the tar ball approach seems to follow helm conventions.

staging/jenkins/Chart.yaml
shaneutt: :thinking: CI is hardcoded to require the chart version be updated as far as upstream we'll just basically be "diverged" until we merge something up there.

shaneutt:
> @shaneutt the tar ball approach seems to follow helm conventions.

I'm not going to block on that :+1:

staging/jenkins/Chart.yaml
joejulian: That list comes from upstream: https://github.com/helm/charts/blob/master/stable/jenkins/Chart.yaml

joejulian:
Please expand the tarball, see https://github.com/mesosphere/charts/tree/80b1850e8aa8061959470cffb5215c3731a3a96d/staging/prometheus-operator/charts

staging/jenkins/Chart.yaml
joejulian: https://github.com/mesosphere/charts#repository-structure
staging is for temporary forks where there is a pending PR upstream.

hectorj2f:
@kvish what is the state of this PR ?

vinodkone:
After the latest discussion on this approach, I think we decided to close this PR and open a new one that adds Jenkins to kubeaddons. cc @kaiwalyajoshi  @takirala

PR #178, opened 2019-10-09
traefik: add annotations to job to delete the job once succeeded

hectorj2f:
Add annotation to delete the pod once it succeed.


staging/traefik/templates/init-cert-job.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
   # an upgrade. See more details in:
   # https://github.com/helm/helm/issues/5482
   name: {{ include "traefik.fullname" . }}-{{ .Chart.Version }}
+  annotations:
+    "helm.sh/hook-delete-policy": "hook-succeeded"
```
jieyu: i would be more defensive here:
```yaml
"helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
```

hectorj2f: okay

PR #179, opened 2019-10-09
Grant Prometheus access to namespaces and configmaps

gracedo:
This grants the Prometheus pod access to "get" namespaces and "get"/"create" configmaps. This is needed so that a container running in the pod can get the kube-system namespace's UID and then create a configmap containing that UID (`cluster_id: <uid>`) -- [see pr here](https://github.com/mesosphere/kubeaddons-extrasteps/pull/22). That configmap is then used in the `prometheus-config-reloader` container to set env variables, which is used to populate prometheus config `externalLabels`'s `cluster` label.

### DIFF
--- a/staging/prometheus-operator/templates/prometheus/clusterrole.yaml
+++ b/staging/prometheus-operator/templates/prometheus/clusterrole.yaml
@@ -25,6 +25,14 @@ rules:
   - endpoints
   - pods
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources:
+  - namespaces
+  verbs: ["get"]
+- apiGroups: [""]
+  resources:
+  - configmaps
+  verbs: ["get", "create"]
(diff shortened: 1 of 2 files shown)


staging/prometheus-operator/templates/prometheus/clusterrole.yaml
```diff
@@ -25,6 +25,14 @@ rules:
   - endpoints
   - pods
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
```
shaneutt: While we have an opportunity to, could we add the context about why we need these permissions in a comment here? It could be helpful for future readers to be able to tell exactly why we need these perms at a glance just by looking at the helm chart.

gracedo:
I'm going to close this and open up a new PR with the changes to instead add a values option to add additional RBAC rules

PR #184, opened 2019-10-11
prometheus: Option to add additional rbac rules

gracedo:
Instead of hardcoding the additional rbac rules needed for Prometheus to run the extrasteps command (https://github.com/mesosphere/kubeaddons-extrasteps/pull/22), we should instead add an option to define additional RBAC rules, and set those rules in kubeaddons-configs (https://github.com/mesosphere/kubeaddons-configs/pull/268).


staging/prometheus-operator/templates/prometheus/clusterrole.yaml
```diff
@@ -32,4 +32,18 @@ rules:
...
+  {{- end }}
+  resources:
+  {{- range .resources }}
+  - {{ . }}
+  {{- end }}
+  verbs:
+  {{- range .verbs }}
+  - {{ . }}
+  {{- end }}
+{{- end }}
```
s12chung: think you can do something like:

`{{- toYaml .Values.mesosphereResources.additionalPrometheusRBACRules | nindent 4 }}`

instead of all this code. might need to play around with `nindent` a bit.

joejulian:
Since this is a security related change, I'd like @mesosphere/mesosphere-security to review. I feel like statically defining permissions is better because the permission requirements should not require user intervention, but I'll leave that to the security team.

gracedo:
I think that's a legitimate concern - I was thinking maybe we could instead use a boolean `enableAdditionalPrometheusRBACRules` to turn on the set of statically defined rules in the chart. I'll wait for @mesosphere/mesosphere-security to weigh in :)

gracedo:
Thanks for weighing in @jongiddy !

I'll still make the change to put statically defined rules behind a flag; that set of rules is meant to be static so I think it's a good idea to enforce it.

staging/prometheus-operator/values.yaml
```diff
@@ -64,6 +64,9 @@ mesosphereResources:
     scheduler: true
     traefik: true
     velero: true
+  ## Additional permissions to grant to a container running in the Prometheus pod.
+  ## See: https://github.com/mesosphere/kubeaddons-extrasteps/tree/master/pkg/prometheus/prometheus.go
+  enableAdditionalPrometheusRBACRules: false
```
gracedo: I defaulted this to false since it should only be run when the extra container is specified/defined in kubeaddons-configs - will flip it to `true` in kubeaddons-configs, lmk if this makes sense

gracedo:
@shaneutt mind merging this one? :)

PR #186, opened 2019-10-14
kommander: Add kommander system namespace federated to all clusters

jimmidyson:
Fixes https://jira.mesosphere.com/browse/DCOS-59829.


jieyu:
@jimmidyson any specific reason we don't create a system "Project" here? I feel we should not have too much specialty about this namespace than any other user facing Project namespaces. We can solve the visibility issue later using labels on the Project.

This is non blocking. Just want to hear your reasoning on this.

PR #187, opened 2019-10-15
kommander: Update KCL dependency to 0.1.6

jimmidyson:
This supports Konvoy 1.3 beta.


jimmidyson:
Self-merging as this was just a bump of chart version.

PR #189, opened 2019-10-16
adding kudo as a chart

kensipe:
This is a tested version of a kudo chart for v0.7.4.   

**Notes:**
1. install `helm install --name kudo  ./kudo`
2. kudo currently does NOT support being in a namespace other than `kudo-system`.  This chart creates that ns and all manifests directly use that ns (until some future time)
3. it is possible to use a different version... to do so pass a different image tag `helm install --name kudo  ./kudo --set image.tag=v0.7.3`
4. it is possible to install without CRDs... pass `--set installCRD=false`

Fixes: https://github.com/kudobuilder/kudo/issues/920

### DIFF
--- a/staging/kudo/Chart.yaml
+++ b/staging/kudo/Chart.yaml
@@ -1,11 +1,12 @@
 apiVersion: v1
-appVersion: "1.0"
+appVersion: "0.7.4"
 description: "Kubernetes Universal Declarative Operator"
 name: kudo
-version: 0.7.4
+version: 0.1.0
 home: https://kudo.dev
 sources:
 - https://github.com/kudobuilder/kudo
+icon: https://raw.githubusercontent.com/kudobuilder/www/master/content/.vuepress/public/favicon.ico
 maintainers:
 - name: kensipe
   email: kensipe@gmail.com
--- a/staging/kudo/Chart.yaml
+++ b/staging/kudo/Chart.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+appVersion: "1.0"
+description: "Kubernetes Universal Declarative Operator"
+name: kudo
+version: 0.7.4
+home: https://kudo.dev
+sources:
+- https://github.com/kudobuilder/kudo
+maintainers:
+- name: kensipe
+  email: kensipe@gmail.com
+- name: Alena Varkockova
+  email: avarkockova@mesosphere.com
+- name: Gerred Dillon
+  email: gdillon@d2iq.com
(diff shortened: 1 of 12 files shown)


staging/kudo/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+appVersion: "1.0"
+description: "Kubernetes Universal Declarative Operator"
+name: kudo
+version: 0.7.4
```
shaneutt: It seels like this and the `appVersion` should be swapped, no? `version` is the Chart version, and since this is the first version `0.1.0` or something like that seems reasonable

staging/kudo/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
+
```
shaneutt:
```suggestion
```

staging/kudo/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - name: SECRET_NAME
+          value: kudo-webhook-server-secret
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        name: manager
+        ports:
+        - containerPort: 9876
+          name: webhook-server
+          protocol: TCP
+        resources:
```
shaneutt: Do we want to set limits here as well? As written currently it will be pretty easy to get the controller scheduled, but do we want to set some limits above this to give some burst room?

staging/kudo/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+crds:
+  forceDeploy: false
+
+resources:
+  requests:
+    cpu: 100m
+    memory: 50Mi
+
+# When installing addons on a cluster that already has kudo crds
```
shaneutt: comments++ :+1:

staging/kudo/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+image:
+  repository: kudobuilder/controller
+  tag: v0.7.4
+  pullPolicy: IfNotPresent
+
+
+crds:
+  forceDeploy: false
+
+resources:
```
shaneutt: Similar to the above: do we want to set some default limits in the values and have running with no limits be an opt-in config?

staging/kudo/Chart.yaml
kensipe: good point

PR #190, opened 2019-10-17
Add config checksum annotation

jr0d:
add rolebinding

remove hooks

    TODO: make hooks work so we can drop privileges after the init container
    completes


jr0d:
closing, we are considering another approach

PR #191, opened 2019-10-21
updated the azuredisk-controller

sebbrandt87:
- to be able to provision always successful


stable/azuredisk-csi-driver/templates/crd-csi-node-info.yaml
```diff
@@ -1,38 +0,0 @@
...
-                  type: string
-                type: array
-          type: array
-  version: v1alpha1
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
```
s12chung: curious, how will this work without the CRDs?

sebbrandt87: since 1.14 the crds are in as it the api is in beta state
and the chart is anyway defined as Kubernetes > 1.15.0 :)

PR #192, opened 2019-10-22
Maintainers is array, not object

alenkacz:


s12chung:
ah, need to bump up chart version

PR #193, opened 2019-10-22
dex-controller: bump to 0.2.1

jieyu:
This removes the leader election flag to workaround a potential bug in
controller runtime that might cause leader election to stuck.

### DIFF
--- a/stable/dex-controller/templates/deployments.yaml
+++ b/stable/dex-controller/templates/deployments.yaml
@@ -40,7 +40,6 @@ spec:
           {{- end }}
       - args:
         - --metrics-addr=127.0.0.1:8080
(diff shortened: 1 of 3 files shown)


stable/dex-controller/templates/deployments.yaml
```diff
@@ -40,7 +40,6 @@ spec:
           {{- end }}
       - args:
         - --metrics-addr=127.0.0.1:8080
-        - --enable-leader-election
```
samba: Curious, why removing this?

jieyu: @samba see https://github.com/mesosphere/dex-controller/pull/32

PR #194, opened 2019-10-22
use apps/v1

joejulian:
With the release of kubernetes 1.16, extensions/v1beta1, apps/v1beta1,
and apps/v1beta2 APIs were retired. This allows these charts to use the
apps/v1 api which is available in all supported versions.


samba:
How will we ensure these only get deployed on compatible versions of Kubernetes?
Do we have a suitable branching strategy in this repo to support this?

joejulian:
https://github.com/mesosphere/charts/pull/194#issue-331166727

joejulian:
`apps/v1` has been available since 1.9.

stable/dex/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: dex
-version: 1.6.6
+version: 1.6.7
```
jieyu: Can you rebase. This needs to be 1.6.8

joejulian: done

PR #195, opened 2019-10-23
Add Thanos Querier to Kommander

branden:
https://jira.mesosphere.com/browse/DCOS-59074

This updates the Kommander chart to deploy Thanos Querier on the cluster where Kommander is running, and create a federated mTLS proxy for the Thanos sidecar that runs on on each connected Konvoy cluster. This PR does not include a controller that updates Thanos Querier config with new clusters as they're added to Kommander; this will be coming later. In the meantime, clusters can be added manually by editing the Thanos deployment and adding a `--store` argument to the container.

This PR adds two new charts. The first is `mtls-proxy`, which deploys a ghostunnel mTLS proxy for a given target. The second is `kommander-thanos`, which deploys the `thanos` helm chart, a `mtls-proxy` `FederatedAddon` that's configured to proxy the Thanos sidecar, and the certificate secrets necessary to secure them. (I chose to deploy the mTLS proxy as a single `FederatedAddon` rather than separate federated resources so that it can be reused in an upcoming PR.) This PR then adds `kommander-thanos` as a subchart to `kommander`.

The `kommander-thanos` chart generates certificates using Helm template builtins because it was the simplest implementation I could come up with. I'm happy to reimplement that if there's a preferred way to generate certificates.


branden:
It looks like CI is failing because it doesn't have the repo for the Thanos chart:

https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2377768&_focus=623
```
[Step 1/1] Error: no repository definition for https://kubernetes-charts.banzaicloud.com. Please add the missing repos via 'helm repo add'
```

alejandroEsc:
> It looks like CI is failing because it doesn't have the repo for the Thanos chart:
> 
> https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2377768&_focus=623
> 
> ```
> [Step 1/1] Error: no repository definition for https://kubernetes-charts.banzaicloud.com. Please add the missing repos via 'helm repo add'
> ```

Right, you have to add it as part of your PR. Check:

```
/test/ct.yaml:
...
Chart-repos:
  - mesosphere-staging=https://mesosphere.github.io/charts/staging
  - mesosphere-stable=https://mesosphere.github.io/charts/stable
  - kubefed-charts=https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
  - jetstack-charts=https://charts.jetstack.io
  - kommander-cluster-lifecycle=https://mesosphere.github.io/kommander-cluster-lifecycle/charts
```

stable/kommander-thanos/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+    extraArgs:
+    - "--grpc-client-tls-secure"
+    - "--grpc-client-tls-cert=/etc/certs/tls.crt"
+    - "--grpc-client-tls-key=/etc/certs/tls.key"
+    - "--grpc-client-tls-ca=/etc/certs/ca.crt"
+    - "--grpc-client-server-name=server.thanos.localhost.localdomain"
+    certSecretName: kommander-thanos-client-tls
+    http:
+      ingress:
+        enabled: true
```
gracedo: maybe default to `false` since we're not exposing this service out to users? it is useful for debugging though, so we can easily override the value when deploying kommander.

branden: Would we want to keep for customer support? We don't have to link to it anywhere, but it might be handy for us or the support team if we can browse to a URL and get a list of stores and their connection status, without needing to redeploy or reconfigure Kommander. I'm not sure if there's another way to get that information out of Querier.

stable/kommander-thanos/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# Internal address for the cluster's Thanos gRPC service.
+# thanosAddress: "HOST:PORT"
+thanosAddress: "prometheus-kubeaddons-prom-prometheus.kubeaddons.svc.cluster.local:10901"
+
+federate:
+  systemNamespace:
+    name: kommander-system
```
branden: Is there a way to set this value using the kommander chart's `federate.systemNamespace.name`? I don't like duplicating it here, but I couldn't figure out how else to set the value.

stable/kommander-thanos/templates/secret.yaml
```diff
@@ -0,0 +1,33 @@
+{{ $thanosCa := genCA "Kommander Thanos CA" 3650 }}
+{{ $thanosServer := genSignedCert "server.thanos.localhost.localdomain" nil nil 3650 $thanosCa }}
+{{ $thanosClient := genSignedCert "client.thanos.localhost.localdomain" nil nil 3650 $thanosCa }}
+---
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: kommander-thanos-client-tls
```
branden: I'm using a fixed name for this secret because I need to pass it to the thanos subchart (`thanos.query.certSecretName`), and I wasn't sure how to use Helm-generated values in `values.yaml`.

stable/kommander/requirements.lock
```diff
@@ -5,5 +5,8 @@ dependencies:
 - name: kommander-cluster-lifecycle
   repository: https://mesosphere.github.io/kommander-cluster-lifecycle/charts
   version: v0.1.9
-digest: sha256:553d3c5c6ae5dda10093e5149170e92a0313958d64718add8716f514e912e211
-generated: "2019-10-25T10:30:23.462935-07:00"
+- name: kommander-thanos
+  repository: https://branden.github.io/charts/stable
```
hectorj2f: this should be mesosphere repo :/. That tells me we need two PRs one with kommander-thanos and another using it.

branden:
Thanks for reviewing, @hectorj2f. I removed the subchart addition from this PR, it's now in https://github.com/mesosphere/charts/pull/213.

stable/kommander-thanos/values.yaml
jimmidyson: This should happen through kubeaddons, somehow... This will be set by kommander chart (assuming this is going to be a dependency of kommander chart?) and so will be passed down from there to this chart when installing kommander chart.

hectorj2f: yeah, this should be defined by kubeaddons-config properties

shaneutt: the `Addon` will take care of the namespace, I believe this field can be removed and the release-namespace can be used instead.

s12chung: > the Addon will take care of the namespace, I believe this field can be removed and the release-namespace can be used instead.

if it's a sub chart, that's only if the `kommander-system` matches the release namespace of the commander addon, which it doesn't:

https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/kommander.yaml#L14

hectorj2f: okay, I understand the reason. We can keep it as it is because the addon will set a diff namespace value, it isn't very clean but it is valid for this scenario.

s12chung: took a look for an alternative, but: https://github.com/helm/helm/issues/2813

PR #205, opened 2019-10-24
kommander: increase timeout for liveness, env maintenance

brandonc:
I noticed during MWT that when Kommander was very busy polling large numbers of clusters the default liveness timeout of 1 second was sometimes not enough time for kommander to respond and this would send the pod into a termination loop.

I also added and removed some env variables to reflect the actual env config.

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -1,13 +1,13 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.141.4
+  tag: 1.145.6
   pullPolicy: IfNotPresent
 replicas: 1
 logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
-clientPollingInterval: 3000
 hostPollingInterval: 3000
-kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.4","konvoyVersion":"v1.3.0-kommander-beta2","kubeaddonsVersion":"kommander-beta2"},{"kubernetesVersion":"1.15.3","konvoyVersion":"v1.1.5","kubeaddonsVersion":"stable-1.15.3-4"}]'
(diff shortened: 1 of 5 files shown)


jaxesn:
Make sure and bump the chart version.  Lets also wait until your code is merged and built and update the app version as well.  There are two places where the app version is so make sure and update both.

brandonc:
👍 ill add that change after the other merges

stable/kommander/values.yaml
```diff
@@ -5,8 +5,8 @@ image:
 replicas: 1
 logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
-clientPollingInterval: 3000
 hostPollingInterval: 3000
+clusterPollingThrottleFactor: 1
 kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.4","konvoyVersion":"v1.3.0-kommander-beta2","kubeaddonsVersion":"kommander-beta2"},{"kubernetesVersion":"1.15.3","konvoyVersion":"v1.1.5","kubeaddonsVersion":"stable-1.15.3-4"}]'
```
jaxesn: can you `v1.3.0-kommander-beta2` -> `v1.3.0-kommander-beta3` and `kommander-beta2` -> `kommander-beta3` while you are editing here?  Also, lets wait for the catalog changes to be merged and use that image tag.

brandonc: Yes

brandonc:
@jaxesn I guess I'll rely on you to tell me which kommander version to set the app version to?

jaxesn:
When we do update the version lets update it here too.

https://github.com/mesosphere/charts/blob/dev/stable/opsportal/values.yaml#L75

PR #211, opened 2019-10-28
kommander: enable federated CRDs

alejandroEsc:
as part of work in: https://jira.mesosphere.com/browse/DCOS-60108


stable/kommander/values.yaml
```diff
@@ -50,9 +50,12 @@ ingress:
 
 federate:
   addons: true
+  clusteraddons: true
+  clusterrolebindings: true
+  customresourcedefinitions: true
   roles: true
   rolebindings: true
-  clusterrolebindings: true
```
s12chung: why change this?

alejandroEsc: just moved it, not changed.

s12chung: ah, whops.

PR #212, opened 2019-10-29
kommander: enable federation of kubeaddons controller

alejandroEsc:
change allow federating kubeaddons controller

Tested locally, we assume by default we are deploying on a konvoy cluster, to avoid deleting the kubeaddons namespace when removing the chart.

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -53,9 +53,12 @@ federate:
   clusteraddons: true
   clusterrolebindings: true
   customresourcedefinitions: true
+  kubeaddons: true
   roles: true
   rolebindings: true
 
+  # set to true if you wish to deploy on a non konvoy cluster
+  createKubeaddonsNamespace: false
 
   systemNamespace:
     enable: true
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -57,7 +57,7 @@ federate:
   roles: true
   rolebindings: true
 
-  image:
+  kubeaddonsInitializer:
     repository: "mesosphere/kubeaddons-addon-initializer"
     tag: "v0.1.1"
     pullPolicy: IfNotPresent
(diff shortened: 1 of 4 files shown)


stable/kommander/values.yaml
```diff
@@ -53,9 +53,12 @@ federate:
   clusteraddons: true
   clusterrolebindings: true
   customresourcedefinitions: true
+  kubeaddons: true
   roles: true
   rolebindings: true
 
+  # set to true if you wish to deploy on a non konvoy cluster
```
jimmidyson: My one concern is that now we need to know whether we're deploying on Konvoy or non-Konvoy and can't use this as same resource(s) to federate to all joined clusters. Is there a way round this? As in, perhaps a pre-install hook that only creates kubeaddons namespace if required?

alejandroEsc: alright, wasnt sure how we wanted to handle that. I will test a bit with a pre-install hook and get back to you. The word is that helm should not delete the resource in this case, which is what i would expect, but let me verify.

s12chung: could add it as an init container for kommander?

alejandroEsc: @s12chung yep, digging into that possibility now.

stable/kommander/values.yaml
```diff
@@ -53,9 +53,15 @@ federate:
   clusteraddons: true
   clusterrolebindings: true
   customresourcedefinitions: true
+  kubeaddons: true
   roles: true
   rolebindings: true
 
+  image:
+    repository: "mesosphere/kubeaddons-addon-initializer"
```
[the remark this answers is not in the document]
alejandroEsc: This is the image that is responsible for creating the namespace if necessary.

stable/kommander/values.yaml
```diff
@@ -53,9 +53,15 @@ federate:
   clusteraddons: true
   clusterrolebindings: true
   customresourcedefinitions: true
+  kubeaddons: true
   roles: true
   rolebindings: true
 
+  image:
```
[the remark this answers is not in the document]
alejandroEsc: image is what we name docker images  normally.

PR #213, opened 2019-10-29
kommander: Add Thanos Querier

branden:
https://jira.mesosphere.com/browse/DCOS-59074

This depends on #195.

This PR adds the `kommander-thanos` chart from #195 as a subchart of `kommander`.

PR #215, opened 2019-10-29
[kommander] Add Grafana as a subchart

gracedo:
This adds Grafana as a subchart to Kommander. This Grafana is specific to Kommander (as opposed to the Grafanas deployed as part of prometheus-operator in all the Konvoy clusters) and is meant to only have dashboards specific to the Kommander view. The datasource should point to the Thanos Query http service endpoint, which is currently in review (#195 #213). When those two PRs are merged, I will update this chart's version and datasource if necessary. Dashboards are in progress and will be added in a future PR.


gracedo:
@alejandroEsc This is not meant to "replace" the currently running Grafana installed with prometheus-operator, we want to leave that Grafana up and running since it's meant for that cluster's metrics. This additional Grafana is deployed with Kommander and meant to only contain dashboards specific to the Kommander view - e.g. dashboards on kommander-only metrics (like future Thanos Querier and Karma metrics, both components only running on kommander), dashboards with an overview of all connected clusters (metrics from all connected clusters are pulled in via Thanos Querier).

alejandroEsc:
> @alejandroEsc This is not meant to "replace" the currently running Grafana installed with prometheus-operator, we want to leave that Grafana up and running since it's meant for that cluster's metrics. This additional Grafana is deployed with Kommander and meant to only contain dashboards specific to the Kommander view - e.g. dashboards on kommander-only metrics (like future Thanos Querier and Karma metrics, both components only running on kommander), dashboards with an overview of all connected clusters (metrics from all connected clusters are pulled in via Thanos Querier).
> 
> I have tested this pretty thoroughly, but will do another round of testing once #195 and #213 are merged in!

Thanks so much for taking the time to reply, i totally appreciate!

gracedo:
image

PR #216, opened 2019-10-30
Add karma chart with certSecretName parameter

branden:
https://jira.mesosphere.com/browse/DCOS-59076

This PR copies the karma chart from https://github.com/helm/charts/tree/master/stable/karma, and includes the `certSecretName` parameter from https://github.com/helm/charts/pull/18433. Copying the chart to this repo is only necessary to pull in the changes from that PR; it can be removed once the PR is merged.


jimmidyson:
As this is using a chart from upstream with hopefully potentially accepted PR, could you please move this to staging as detailed in https://github.com/mesosphere/charts#repository-structure. I'll review as-is anyway and we can just get it moved before merging.

jimmidyson:
Let me review the upstream PR instead, I'll assume that this chart is the same as upstream with the PR changes applied.

branden:
Pushed a change that moves the chart to staging, and reflects the current state of the upstream PR.

PR #217, opened 2019-10-30
Add kommander-karma chart

branden:
https://jira.mesosphere.com/browse/DCOS-59076

This PR depends on the `mtls-proxy` chart from #195. It also depends on the karma chart from #216.

This is part of a project to add support to Kommander for centralized monitoring of Konvoy clusters.

Konvoy's Prometheus addon provides an Alertmanager that handles alerts on that cluster. This PR adds a new chart that deploys [Karma](https://github.com/prymitive/karma), which provides a centralized alert dashboard for multiple Alertmanagers, and declares a federated mTLS proxy for the Alertmanager on each managed Konvoy. Karma consumes each Alertmanager's API through its respective proxy, and provides an alert dashboard that reflects the state of all managed Konvoys' Alertmanagers.

This does not include a controller for service discovery, which will be added later. In the meantime, clusters can be added manually by editing the Karma configmap and adding an entry to `alertmanagers` with the proxy's external address and TLS config. See the testing section below for an example.

This PR adds the `kommander-karma` chart, which deploys the Karma Helm chart, a `mtls-proxy` `FederatedAddon` that's configured to proxy Alertmanager, and the certificate secrets necessary to secure their communication. The `kommander-karma` chart will be added as a subchart of `kommander` in a later PR.

## Testing

This is how I tested the PR:

1. Deploy a Konvoy cluster.
1. Join the cluster to the kubefed federation.
    * `kubefedctl join kommander --kubefed-namespace=kommander --host-cluster-context=kommander-admin@kommander --host-cluster-name=kommander --cluster-context=kommander-admin@kommander`
1. Deploy the kommander-karma helm chart.
    * `helm install stable/kommander-karma --name kommander-kommander-karma --namespace kommander --set alertmanagerAddress="prometheus-kubeaddons-prom-alertmanager.kubeaddons.svc.cluster.local:9093"`
1. Retrieve the external hostname for the proxy's load balancer.
    * `kubectl -n kommander-system get service kommander-kommander-karma-proxy-kubeaddons-mtls-proxy -o jsonpath="{ .status.loadBalancer.ingress[0].hostname}"`
1. Add the proxy to Karma's configuration as an Alertmanager.
    1. `kubectl -n kommander edit configmap kommander-kommander-karma-config`
    1. Remove the placeholder in `karma.conf`'s `alertmanager.servers` and replace it with this entry:
    ```yaml
      - name: kommander
        uri: https://<proxy-hostname>:443  # Replace with the actual proxy hostname from the previous step.
[...]

PR #218, opened 2019-10-30
kommander: Add Karma

branden:
https://jira.mesosphere.com/browse/DCOS-59076

This depends on #217.

This PR adds the `kommander-karma` chart from #217 as a subchart of `kommander`.


branden:
@alejandroEsc This will be deployed by default when opsportal bumps kommander. That can be prevented by setting `kommander-karma.enabled=false`. LMK if you'd like me to submit a PR to add that to opsportal's values.

stable/kommander/values.yaml
```diff
@@ -97,6 +97,10 @@ kommander-thanos:
   enabled: true
   thanosAddress: "prometheus-kubeaddons-prom-prometheus.kubeaddons.svc.cluster.local:10901"
 
+kommander-karma:
+  enabled: true
```
hectorj2f: I am not sure if we want this enabled by default. I suggest you set it to false in kubeaddons-configs.

hectorj2f: @branden ^

branden: My understanding is that we want Karma enabled by default when Kommander is deployed.

branden: /cc @gracedo @philipnrmn

gracedo: That's also my understanding - that Karma should be enabled by default

PR #234, opened 2019-11-05
Kubefed chart

s12chung:
first 2 commits = from upstream + moving files

### DIFF
--- a/staging/kubefed/charts/controllermanager/templates/install-crds.yaml
+++ b/staging/kubefed/charts/controllermanager/templates/install-crds.yaml
@@ -1,0 +1,86 @@
+{{ if or (not .Values.global.scope) (eq .Values.global.scope "Cluster") }}
+{{if not (.Capabilities.APIVersions.Has "core.kubefed.io/v1beta1") }}
+---
+{{.Files.Get "files/crds.yaml"}}
+{{ end }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: kubefed-controllermanager-crds
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+data:
+  crds.yaml: |-
+{{.Files.Get "files/crds.yaml" | printf "%s" | indent 4}}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: kubefed-controllermanager-crds
+  annotations:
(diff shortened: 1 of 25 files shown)


staging/kubefed/charts/controllermanager/templates/install-crds.yaml
```diff
@@ -0,0 +1,84 @@
+{{ if (or (or (not .Values.global.scope) (eq .Values.global.scope "Cluster")) (not (.Capabilities.APIVersions.Has "core.kubefed.io/v1beta1"))) }}
```
joejulian: Is this the right logic? 
Even if the cluster has this capability, we still want it to run. This will allow the upgrading of installed CRDs.

s12chung: good catch. i changed one of the conditions, but forgot the other....

s12chung: changed.

staging/kubefed/charts/controllermanager/templates/webhook.yaml
```diff
@@ -0,0 +1,139 @@
+{{- $ca := genCA "kubefed-admission-webhook-ca" 3650 }}
```
joejulian: This file will have the same problem as CRDs. Can we apply the crd hook to ValidatingWebhookConfigurations?

staging/kubefed/charts/controllermanager/templates/install-crds.yaml
```diff
@@ -0,0 +1,85 @@
+{{ if or (not .Values.global.scope) (eq .Values.global.scope "Cluster") }}
+{{if not (.Capabilities.APIVersions.Has "core.kubefed.io/v1beta1") }}
```
joejulian: We want this file to happen regardless of whether this capability exists otherwise we won't ever be able to upgrade the CRDs.

staging/kubefed/templates/install-crds.yaml
```diff
@@ -0,0 +1,86 @@
+{{ if or (not .Values.global.scope) (eq .Values.global.scope "Cluster") }}
+{{if not (.Capabilities.APIVersions.Has "types.kubefed.io/v1beta1") }}
```
joejulian: Here, too, we don't want to test for the capability

alejandroEsc: ? are the end closures correct? he should check the first time if installed or not, and the upgrade path should always happen. I think this is ok?

staging/kubefed/charts/controllermanager/templates/webhook.yaml
joejulian: Since this may not be a problem, we'll skip worrying about it until it is.

PR #237, opened 2019-11-06
[kommander] crd fix (part 1)

alejandroEsc:
This PR is following the solution to fixing crd installation problems encountered when trying to deploy/upgrade a chart given existing crds. It also functionally attempts to update crds if so required. 

part 1 of 2. Second part would include pulling in a new kubefed chart that contains fixes for those CRDs.


jimmidyson:
We can get rid of observable cluster, confirmed with UI team.

alejandroEsc:
> We can get rid of observable cluster, confirmed with UI team.

awesome, if that is the case let me update the chart.

jimmidyson:
I think we can also get rid of the separate enablement of different federated resources, they should always all be enabled.

alejandroEsc:
> I think we can also get rid of the separate enablement of different federated resources, they should always all be enabled.

sounds ok, i will do so in a follow pr

PR #241, opened 2019-11-06
cert-manager-setup: use crd pattern

s12chung:


staging/cert-manager-setup/files/crds.yaml
```diff
@@ -7512,4 +7511,3 @@ status:
   conditions: []
```
jieyu: Can we remove `"helm.sh/hook": crd-install` annotation?

jieyu: oh, i hit approval accidentally

staging/cert-manager-setup/templates/install-crds.yaml
```diff
@@ -0,0 +1,86 @@
...
+      restartPolicy: OnFailure
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cert-manager-setup-crds
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
```
jieyu: Can you also add "before-hook-creation" just to be safe?

"helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation

jieyu: Please apply this also to other places (even in kommander chart)

staging/cert-manager-setup/files/crds.yaml
jimmidyson: No the crd-install hook still needs to be used here.

s12chung: yeah, still needs to be here.

staging/cert-manager-setup/templates/install-crds.yaml
s12chung: done for this chart, there's many places where this needs to be applied. will do it in another PR

staging/cert-manager-setup/files/crds.yaml
jieyu: Hum? Why?

s12chung: i believe helm 2 uses it to check whether it's a CRD or not for validations.

staging/cert-manager-setup/templates/install-crds.yaml
s12chung: other PR here: https://github.com/mesosphere/charts/pull/242

staging/cert-manager-setup/files/crds.yaml
jieyu: oh, right. OK.

PR #244, opened 2019-11-06
Adds a chart which can deploy an ingress

jr0d:
This is needed for  https://github.com/mesosphere/kommander-cluster-lifecycle/pull/206 , once we update to traefik2, this can go away unless we need it for something else.


timaa2k:
https://jira.mesosphere.com/browse/DCOS-60887

PR #245, opened 2019-11-07
kommander/kommander-ui: splitting the kommander chart

alejandroEsc:
This PR splits the UI parts/pieces from kommander and creates a kommander-ui chart.

```
$ helm list
NAME        	REVISION	UPDATED                 	STATUS  	CHART               	APP VERSION	NAMESPACE
cert-manager	1       	Wed Nov  6 21:27:14 2019	DEPLOYED	cert-manager-v0.10.1	v0.10.1    	cert-manager
kommander   	1       	Wed Nov  6 21:28:16 2019	DEPLOYED	kommander-0.2.3     	1.149.0    	kommander
kommander-ui	1       	Wed Nov  6 21:03:17 2019	DEPLOYED	kommander-ui-0.2.0  	1.149.0    	kommander
```

```
k get pods -n kommander
NAME                                          READY   STATUS             RESTARTS   AGE
kcl-cm-8fb66b498-p6xhh                        1/2     CrashLoopBackOff   5          5m1s
kcl-tfcb-5686c45448-wpkdl                     1/1     Running            0          5m1s
kcl-utility-apiserver-6b6d65ddcc-jrbcd        1/1     Running            0          5m1s
kcl-webhook-56bb557f8f-r6nqp                  1/1     Running            0          5m1s
kommander-grafana-84568f7547-28rvl            2/2     Running            0          5m1s
kommander-karma-6cbbf454db-bz9tn              1/1     Running            0          5m1s
kommander-thanos-query-85fdcf5b4c-m7t55       1/1     Running            0          5m1s
kommander-ui-79c65f8945-ggzqr                 1/1     Running            0          31m
kubefed-admission-webhook-7fdcc5b56f-jnptx    1/1     Running            0          5m1s
kubefed-controller-manager-75bdd8c787-7vcmk   1/1     Running            0          5m1s
kubefed-controller-manager-75bdd8c787-m8qrw   1/1     Running            0          5m1s
```

```
$k -n kommander logs -f kommander-ui-79c65f8945-ggzqr
info: Server is running on http://localhost:4000
info: We are serving the frontend from "/app/kommander/backend/node_modules/@kommander/frontend/dist/" - Kommander Mode - Production
info: Initializing store...
warn: Could not fetch ClusterLogins from host: Error: Request failed with status code 404
...
warn: Could not fetch ClusterLogins from host: Error: Request failed with status code 404
warn: Could not fetch ClusterLogins from host: Error: Request failed with status code 404
warn: Could not fetch ClusterLogins from host: Error: Request failed with status code 404
warn: Could not fetch ClusterLogins from host: Error: Request failed with status code 404
info: Subscribing to cluster data for konvoy-host-cluster - Kommander Host @ https://kubernetes.default.svc
info: Kommander cluster polling interval updated to 600000ms
info: Fetched Capabilities for konvoy-host-cluster-Kommander Host. version: v1.15.0 - APIs: {"metrics":false}
```


stable/kommander-ui/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+  initialDelaySeconds: 15
+  periodSeconds: 10
+livenessProbe:
+  initialDelaySeconds: 15
+  periodSeconds: 10
+  timeoutSeconds: 3
+
+### This must match the serviceName set in the ingress backend below
+service:
+  name: kommander
```
s12chung: change this to `kommander-ui`, won't make a diff except for naming niceness.

stable/kommander-ui/templates/hooks-roles.yaml
```diff
@@ -0,0 +1,29 @@
+---
```
shaneutt: Will this SA always be needed, or should we have it's creation flagged?

alejandroEsc: this may be an existing problem with the chart? but also not in the scope of this PR. Could we address this on a later pr?

shaneutt: :thinking: It seems like this is something we may wanna work on in this scope yeah? Set the baseline that it's optional?

alejandroEsc: I will check, but doesnt seem like something i want to touch with this many moving parts.

alejandroEsc: this is required as far as i can tell, by the deployment.

stable/kommander-ui/templates/hooks-roles.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "kommander.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "kommander.labels" . | indent 4 }}
+  annotations:
+    helm.sh/hook: "pre-install"
```
jimmidyson: Why do these need to be pre install hooks?

alejandroEsc: I did not write this. It seems like it was some lifecycle reason?

alejandroEsc:
> In general I'm OK with the file changes, but have we run this through some manual testing that we can simply give a 👍 for?

Steve and myself have run through manual tests for the equivalent of a few days now, But that was a week ago. Ill revert some changes to keep things at the minimum so people are more confident with the changes.

jaxesn:
I want to understand how we are going to handle upgrading 1.2->1.3 here.  We discussed this on slack and there are a number of ideas, but the big thing is we are changing the definition of the kommander chart now where it will be the backend whereas before it was the ui.  When a user upgrades, they need to enable the kommander-ui addon or they will have no kommander or ops-portal ui.  I think @samba mentioned maybe we add a way in konvoy to replace addons on new versions?  either way lets get aligned on how that will work before we move forward with this.  We should lock this down early next week though. cc @dkoshkin

alejandroEsc:
@jaxesn I suspect that upgrades include retrieving a new list of "addons" from the version tag that is being upgraded from. That would leviate the question of having to download the konvoy-ui pieces, as this should be automatic. We have in place a mechanism for upgrading single charts/addons, we can leverage this and have that moving from 1.2 -> 1.3 we "delete" the previous version and upgrade will install the new version. This would have some loss so we should decide if this is viable.

alejandroEsc:
closing this as others are working on this particular case.

PR #255, opened 2019-11-09
Couple of fixes in istio charts

GoelDeepak:
1. Fix typo in grafana dashboard namespace
2. Creates a destination rule to disabled mTLS for prometheus as it is outside of istio
3. Fixes issue in citadel chart


GoelDeepak:
cc @gpaul

staging/istio/charts/prometheus/templates/destinatiorule.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: networking.istio.io/v1alpha3
```
joejulian: typo in filename

GoelDeepak:
Thanks, @shaneutt for taking care of it

gpaul:
Looks like this can be merged?

PR #261, opened 2019-11-15
flagger(,-loadtester): Add chart

gpaul:
This PR adds flagger.

In particular it,

- adds a flagger chart, with a flagger-loadtester subchart
- adds a grafana dashboard for tracking flagger deployments. The dashboard is called "Istio Canary" (that's the default name.) It is deployed to the `kubeaddons` namespace, regardless of where flagger is installed.
- adds a job to label the namespace in which flagger gets installed with `istio-injection=enabled` as flagger relies on istio.
- for the job to execute, we annotate the job, serviceaccount, clusterrole, clusterrolebinding with `helm.sh/hook: pre-install` so those actions are performed before pods are created.

The plan is to install flagger into a `kubeaddons-flagger` namespace, which canthen be marked with `istio-injection=enabled`. We don't want all pods in the `kubeaddons` namespace to be in the istio mesh, but flagger/flagger-loadtester needs to be in the mesh, hence the separate namespace.

### DIFF
--- a/staging/flagger/.helmignore
+++ b/staging/flagger/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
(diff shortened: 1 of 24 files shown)


gpaul:
This chart is ready to be merged, my manual tests passed.

PR #264, opened 2019-11-15
WORKAROUND: Fix Dispatch repository controller ingress lookup

jbarrick-mesosphere:


jbarrick-mesosphere:
This is not needed. I was confused because before our tekton dashboard ingress overwrote the hook and we only had one when we actually needed two.

PR #267, opened 2019-11-16
chore: bump kommander/opsportal to image tag 1.149.77

natmegs:


jaxesn:
I'm.pretty sure the version mapping is still being used so it would need updating to beta6.

stable/kommander/values.yaml
```diff
@@ -1,13 +1,13 @@
...
-  tag: 1.149.63
+  tag: 1.149.77
   pullPolicy: IfNotPresent
 replicas: 1
 logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
 hostPollingInterval: 3000
 clusterPollingThrottleFactor: 1
-kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.5","konvoyVersion":"v1.3.0-beta5","kubeaddonsVersion":"kommander-beta5"}]'
+kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.5","konvoyVersion":"v1.3.0-beta6","kubeaddonsVersion":"kommander-beta6"}]'
```
natmegs: Can someone confirm that these are the right versions for beta6?

s12chung: you probably know more than me. it changes this:

https://github.com/mesosphere/kommander/blob/bd720f9cea98b1dca728ac70f5c3fa22e8a76553/kommander/backend/src/config.ts#L27-L52

alejandroEsc: https://github.com/mesosphere/konvoy/pull/1000/files seems like we are opting for `kommander-beta6-4` which i am not sure of. @jaxesn can i get some clarity here?

jaxesn: Why do we need the -4? Can we just stick with beta6 to make this easy

shaneutt: Yes we can just use it without the -4

PR #270, opened 2019-11-17
kcl: bump dependency

hectorj2f:
Bump KCL containing the fix for the http no proxy settings.


jimmidyson:
Why does `helm dependency update` update the dependencies unrelated to this change too?

hectorj2f:
That is `helm dep update` will pull down the latest charts that satisfy the dependencies, and clean up old dependencies in a blind manner. It doesn't allow to specify which dependency to update.

jimmidyson:
Right, but my question is why have those ones changed? The version hasn't changed so I'm confused why they have changed rather than just the one we changed the version for. Anyway, not a blocker, but confusing to me.

hectorj2f:
I assume the reason might be that the tag of those can remain the same, but the code might change in case anyone backports any fixes to these versions. Otherwise I cannot see another reason, perhaps looking at the helm dep update code, because the documentation doesn't give me any reason.

PR #275, opened 2019-11-20
dispatch: use apps/v1 api

dkoshkin:
Part of https://jira.mesosphere.com/browse/DCOS-61056

Needed for Kubernetes v1.16.


dkoshkin:
Good catch CI :)

justinbarrick:
We need to be careful about making changes here. They’ll get overwritten next time we update the chart from our repository.

dkoshkin:
@justinbarrick I didn't realize you had the chart in a different repo, will PR the same change there.

PR #280, opened 2019-11-22
remove bearerproxy

jr0d:
* Removes bearer proxy
* Adds kubernetes-dashboard chart for 2.0beta6
* Modifies opsportal chart to allow the creation of an admin cluster role binding
* TFA now uses a dedicated serviceaccount with impersonation privileges


jaxesn:
What is the plan here, like what do we do while we wait for upstream to merge/release?  Are we going to fork for a bit?

jr0d:
@jaxesn I am hoping it gets reviewed and approved over the weekend, monday at the latest. If it takes longer, we can always just roll an image and pull it from mesosphere/kubernetes-dashboad.

staging/traefik-forward-auth/Chart.yaml
```diff
@@ -3,12 +3,14 @@ apiVersion: v1
...
+version: 0.2.9
 keywords:
   - traefik-forward-auth
   - traefik
   - oauth
 home: https://github.com/mesosphere/traefik-forward-auth
 maintainers:
   - name: jongiddy
-    email: jgiddy@mesosphere.com
+    email: jgiddy@md2iq.com
```
joejulian: typo

stable/opsportal/templates/hooks-roles.yaml
```diff
@@ -1,43 +1,61 @@
...
+# running an elevated job that can create the ops-portal super user
 {{- if .Values.secrets.create }}
 ---
-apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+apiVersion: v1
+kind: ServiceAccount
 metadata:
-  name: {{ template "opsportal.fullname" . }}-secrets
+  name: "{{.Release.Name}}-service-account-pre-install"
```
joejulian: This runs a risk of being too long. Consider defining a template in _helpers.tpl to generate the name the same way fullname is generated.

jr0d: Good catch, I'll go back to using the truncated name in the template.

jr0d: For some reason, changing this to `{{ template "opsportal.fullname" . }}-admin` breaks the chart.

joejulian: I'll offer a patch when I get to a computer.

PR #282, opened 2019-11-22
awsebs: switch to io1 as default vol types

hectorj2f:
In relation to this change https://github.com/mesosphere/konvoy/pull/1009, we should adapt the chart to use `io1` instead of `gp2`.

### DIFF
--- a/stable/awsebscsiprovisioner/values.yaml
+++ b/stable/awsebscsiprovisioner/values.yaml
@@ -64,9 +64,10 @@ storageclass:
   isDefault: false
   reclaimPolicy: Delete
   volumeBindingMode: WaitForFirstConsumer
-  type: gp2
+  type: io1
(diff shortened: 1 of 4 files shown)


stable/awsebscsiprovisioner/values.yaml
```diff
@@ -64,7 +64,7 @@ storageclass:
   isDefault: false
   reclaimPolicy: Delete
   volumeBindingMode: WaitForFirstConsumer
-  type: gp2
+  type: io1
```
shaneutt: Could we put a comment here with our reasoning for using io1? Might just be a nice breadcrumb for future code readers to better understand our configuration.

hectorj2f:
Closing this PR, we don't need it at this point. The default is only for the control plane.

PR #285, opened 2019-11-26
Added Nvidia GPU monitoring support.

Gilbert88:
This PR mainly contains two subcharts under Nvidia:
1. nvidia-dcgm-exporter: it is used to expose the dcgm GPU metrics to Prometheus
2. grafana: the GPU dashboard in grafana, captured by label in configmap

Screen Shot 2019-11-26 at 3 27 32 PM
Screen Shot 2019-11-26 at 3 29 20 PM


staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
```diff
@@ -0,0 +1,68 @@
...
+        resources:
+{{ toYaml .Values.nvidiaDcgmExporter.resources | indent 12 }}
+        volumeMounts:
+        - name: device-metrics
+          mountPath: /run/prometheus
+{{- if .Values.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 8 }}
+{{- end }}
+
```
alejandroEsc: nit space here.

staging/nvidia/charts/grafana/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+description: A Helm chart for Kubernetes
```
gracedo: Would you mind adding details to the description as to what this subchart does 😄

staging/nvidia/charts/nvidia-dcgm-exporter/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
gracedo: same here re adding details to the description

hectorj2f:
@Gilbert88 why do we need a grafana subchart here? Why don't we rely on a centralized grafana?

hectorj2f:
I would add the Nvidia dcgm exporter as an optional subchart or addon..There might be customers that don't want it installed by default.

staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
Gilbert88: Fixed. Thanks!

staging/nvidia/charts/grafana/Chart.yaml
```diff
@@ -0,0 +1,5 @@
...
+name: grafana
+version: 0.1.2
```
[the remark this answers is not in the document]
Gilbert88: Yea, this is just a config map, which is orthogonal to the grafana version. Updated it to 0.1.0

Gilbert88:
> @Gilbert88 why do we need a grafana subchart here? Why don't we rely on a centralized grafana?

It relies on centralized grafana. I don't want to introduce an empty grafana dashboard when the Nvidia addon is not enabled. This subchart only contains a configMap which represents the GPU dashboard. Grafana captures it by label.

Please see Istio for example. It also introduce dashboard by the same way.

Gilbert88:
> I would add the Nvidia dcgm exporter as an optional subchart or addon..There might be customers that don't want it installed by default.

There is a configurable option to disable this exporter:
https://github.com/mesosphere/charts/pull/285/files#diff-c8fbfea42e3504ca63bf01c1ed86301aR5

Customers can disable it from the cluster.yaml. By default, the GPU exporter is enabled.

I would avoid adding it as a separate addon because it depends on the Nvidia driver and Nvidia device plugin. We should avoid introducing more dependencies between addons.

staging/nvidia/charts/grafana/templates/configmap-dashboards.yaml
```diff
@@ -0,0 +1,1125 @@
+# Generated from 'nodes' from https://raw.githubusercontent.com/coreos/kube-prometheus/master/manifests/grafana-dashboardDefinitions.yaml
+# Do not change in-place! In order to change this file first read following link:
+# https://github.com/helm/charts/tree/master/stable/prometheus-operator/hack
+{{- if .Values.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nvidia-grafana-gpu-dashboard
```
joejulian: Best practice is to have this name generated by the template.

staging/nvidia/charts/grafana/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
joejulian: Just to make sure... these are Apache licensed because they're copied from upstream?

staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
```diff
@@ -0,0 +1,67 @@
+{{- if .Values.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nvidia-dcgm-exporter
```
joejulian: Again, should come from the template

staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
```diff
@@ -0,0 +1,67 @@
...
+  namespace: {{ $.Values.namespace }}
+spec:
+  selector:
+    matchLabels:
+      name: nvidia-dcgm-exporter
```
joejulian: same templated name as above

staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
```diff
@@ -0,0 +1,67 @@
...
+  template:
+    metadata:
+      labels:
+        name: nvidia-dcgm-exporter
```
joejulian: same templated name as above

staging/nvidia/charts/nvidia-dcgm-exporter/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "nvidia-dcgm-exporter.fullname" . }}-test-connection"
```
joejulian: This should be generated in _helpers.tpl. The maximum name length is 64 characters. fullname is trucated at 64 characters, but `-test-connection` adds another 16 characters making it possible for this to be 80 characters and be rejected by the api.

staging/nvidia/charts/grafana/templates/configmap-dashboards.yaml
Gilbert88: The main purpose was to make it explicit and it follows the example of Istio config map.

staging/nvidia/charts/grafana/LICENSE
Gilbert88: no, they are not copied from anywhere. We created this sub-chart for grafana dashboard. Added this license because other Nvidia charts are under apache 2.0. Should we remove it?

staging/nvidia/charts/nvidia-dcgm-exporter/templates/daemonset.yaml
Gilbert88: @joejulian , seems like most of the comments here are about using template for the metadata.name. However, I just looked though most of the charts in the repo (e.g., KUDO, Knative, Istio etc). None of them use the template for the metadata.name.

Could you please explain why that is regarded as best practice? Just for shortening the length of the metadata.name? I would prefer the name to be more explicit instead of requiring people to check the template.

If this is a hard requirement to use the template, could you point me to an existing example in the repo?

staging/nvidia/charts/grafana/LICENSE
Gilbert88: Let's keep the liscense because we developed this dashboard based on top of some contents in this repo https://github.com/NVIDIA/gpu-monitoring-tools

PR #286, opened 2019-11-26
kommander-thanos: Bump thanos subchart, add file SD configmaps

gracedo:
This updates the Thanos subchart dependency to the latest version v0.3.9, which includes the new config option that allows us to set up Thanos Query file service discovery with configmaps (https://github.com/banzaicloud/banzai-charts/pull/977). We use this new `serviceDiscoveryFileConfigMaps` option to specify a configmap `kommander-thanos-query-stores`, which is created in the `kommander-thanos` chart and consists of a yaml file containing an empty set of stores/targets. 

Once this is merged:
1. Another mesosphere/charts PR needs to be opened which bumps the kommander-thanos subchart dependency (to v1.0.5) within the kommander chart
2. Then a kubeaddons-configs PR which bumps the kommander chart.

The merging of this and abovementioned PRs would unblock https://github.com/mesosphere/kommander-cluster-lifecycle/pull/231 which refactors the `ThanosQueryController` to update this configmap with managed clusters' stores as they are joined to the management cluster.

## Testing
This was tested by building and self hosting kommander-thanos, then kommander - and creating a kubeaddons-configs branch `gracedo/test_kommander_thanos_charts_cms` which points to my hosted kommander chart. You may test this on a konvoy cluster by editing your `cluster.yaml`, `ClusterConfiguration` section to point `spec.addons.configVersion` to `gracedo/test_kommander_thanos_charts_cms`.

image


stable/kommander-thanos/templates/query-stores-configmap.yaml
```diff
@@ -0,0 +1,15 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ . }}
+  labels:
+{{ $labels }}
+data:
+  stores.yaml: |-
+    - targets: []
```
jimmidyson: This will overwrite the targets on upgrade. Will that be updated by the controller? Alternatively, could we leave this configmap out and let the controller create it if necessary? This would mean thanos wouldn't start up until the configmap existed (created by controller on startup) but I think that's ok. What do you think?

hectorj2f: yeah, we should test an upgrade here, it could wipe out the stores :S

branden: I don't think the controller would be able to recover from losing all the configured stores in its current form, since it only acts on changes to individual `KommanderCluster` resources. For that to work, we'd need to be able to reconcile changes to this configmap by querying all existing `KommanderCluster`s, retrieving their alertmanager service endpoints, and then overwriting1 the configmap with those endpoints. Would that be a sensible thing to implement?

The alternative of giving the controller total ownership of the configmap's lifecycle should work as well. I think we'd want to expose an option in the KCL chart to define that configmap's name, so we don't hardcode it, and document that this chart will fail to deploy if the configmaps in `serviceDiscoveryFileConfigMaps` don't exist. But that seems reasonable.

gracedo: The only issue I found with giving the controller ownership over the configmap is that due to thanos failing to start up, addons deployment times out (`Error: failed to deploy the cluster: deployment timed out! the following addons were not resolved: kommander`), which isn't great for user experience.

Adding a pre-install hook to the configmap should avoid overwriting on upgrades. I tested this in a cluster by starting a cluster on a version of kommander-thanos, manually modifying the configmap, then redeploying addons with an updated version of kommander which points to a bumped version+1 of kommander-thanos, and seeing that the configmap is not overwritten upon upgrade. I got the idea from [this thread](https://github.com/helm/charts/issues/5167#issuecomment-388833226), only caveat is that the cm may not get deleted upon uninstallation of the chart, but I think that's fine.

This solution would not require any changes to the [KCL refactor PR](https://github.com/mesosphere/kommander-cluster-lifecycle/pull/231). Lmk what you think @hectorj2f @jimmidyson

jimmidyson: If you're adding a pre-install hook, you can add a job to delete it in a post-remove hook perhaps? I'm OK with that, but I'd really rather not leave anything lying around after uninstall.

gracedo: @jimmidyson @hectorj2f I added a post-delete hook job, take a look and let me know what you think, thanks!

stable/kommander-thanos/templates/query-stores-configmap.yaml
```diff
@@ -0,0 +1,48 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    metadata:
+      name: {{ . }}-configmap-cleanup
+    spec:
+      serviceAccountName: {{ $sa }}
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
```
jimmidyson: nit (not blocker): Why such an old version of kubectl?

gracedo: I checked what version other jobs were running; they all ran 1.14.1 so I stuck with it also just to be safe!

PR #290, opened 2019-12-03
kommander-karma: Use upstream Karma chart

branden:
This updates the kommander-karma chart to pull in the upstream karma chart, and removes the copy of the karma chart in this repo.


alejandroEsc:
I labeled this ready but change it as you see fit. Could you paste here a sample deployment or whatever tests you ran that this works?

branden:
Closing this because it looks like we have a few changes we need to make in the karma chart in the near future.

PR #293, opened 2019-12-04
chore: bump kommander/opsportal to image 1.162.0

natmegs:

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -1,13 +1,13 @@
 image:
   repository: mesosphere/kommander
-  tag: 1.153.1
+  tag: 1.163.2
   pullPolicy: IfNotPresent
 replicas: 1
 logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
 hostPollingInterval: 3000
 clusterPollingThrottleFactor: 1
-kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.5","konvoyVersion":"v1.3.0-beta6","kubeaddonsVersion":"kommander-beta6"}]'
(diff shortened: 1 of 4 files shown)


stable/kommander/values.yaml
```diff
@@ -1,13 +1,13 @@
...
-  tag: 1.153.1
+  tag: 1.162.0
   pullPolicy: IfNotPresent
 replicas: 1
 logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
 hostPollingInterval: 3000
 clusterPollingThrottleFactor: 1
-kubernetesVersionsSelection: '[{"kubernetesVersion":"1.15.5","konvoyVersion":"v1.3.0-beta6","kubeaddonsVersion":"kommander-beta6"}]'
+kubernetesVersionsSelection: '["1.15.5", "1.15.4"]'
```
natmegs: @DanielMSchmidt Can we confirm that this is the expected format and these are the correct values?

DanielMSchmidt: Yes, that is the new format 👍

natmegs: @jimmidyson can you confirm that these are the supported versions in KCL? They are an assumption based on https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/kommander.yaml#L49

jimmidyson: For now, yeah, although we do also support 1.16.1, we need a konvoy release to support that properly.

natmegs: @jimmidyson will that support be available in the beta7 konvoy release?

jimmidyson:
Yes it should be

stable/kommander/values.yaml
jimmidyson: Yes

PR #294, opened 2019-12-04
[example] use templating to create pre-install serviceaccount name

joejulian:


alejandroEsc:
@joejulian what is the context here?

PR #300, opened 2019-12-05
fix: remove incorrect catalog env var from kommander chart

natmegs:


natmegs:
@TattdCodeMonkey Initially I removed it entirely, but then I changed my mind and added it back in in case the service deployment changes again it's not a breaking issue but something that can be fixed in the deployment?

@shaneutt Totally, is it possible to get the service deployment namespace from this chart? Or do you think `.Release.Namespace` is sufficient?

shaneutt:
> @TattdCodeMonkey Initially I removed it entirely, but then I changed my mind and added it back in in case the service deployment changes again it's not a breaking issue but something that can be fixed in the deployment?
> 
> @shaneutt Totally, is it possible to get the catalog service namespace from this chart? Or do you think `.Release.Namespace` is sufficient?

If things work like I think they work (@alejandroEsc can maybe double check me) the subchart for the kubeaddons catalog uses the same release namespace so it should in fact be only the `.Release.Namespace` you need to deal with I think? :thinking:

natmegs:
Updated. This feels like an imperfect solution since all the stuff around the service name and namespace are still hardcoded, but lmk!

natmegs:
It will also take a bit more digging to figure out what to change if the catalog service is ever deployed in a different namespace (i.e. when it changed from kubeaddons to kommander ns)

PR #301, opened 2019-12-05
[argo-cd] argo-cd chart added

alejandroEsc:
should have our crd management scheme in place


alejandroEsc:
> 👍 but would like @gpaul to give it a ✔️ because technically he's been added as the maintainer here so it seems only fair he check off xD

I did so since he was the original dispatch owner. For expediency i have done this, but would like to confirm if he would like this or not, either way i can add myself as a code-owner at a later date.

PR #307, opened 2019-12-09
charts: added gcpdisk-csi-driver

sebbrandt87:


sebbrandt87:
it is @hectorj2f 
the image is not available, as there is nothing done in the image as it looks like. they just changed the upstream images for csi-provisionier, what I did anyway ;)

sebbrandt87:
we did the same for azure, I do not see a difference here ;)

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -0,0 +1,116 @@
...
+      serviceAccountName: csi-gce-pd-controller-sa
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      priorityClassName: system-cluster-critical
+      containers:
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:v1.4.0
+          args:
+            - --csi-address=$(ADDRESS)
+            - --timeout=30s
```
hectorj2f: perhaps this should be configurable :/  I can imagine it failing under certain load

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - --leader-election-type=leases
+          env:
+            - name: ADDRESS
+              value: /csi/csi.sock
+          imagePullPolicy: Always
+          volumeMounts:
+            - mountPath: /csi
+              name: socket-dir
+        - name: cluster-driver-registrar
+          image: quay.io/k8scsi/csi-cluster-driver-registrar:v1.0.1
```
hectorj2f: we should expose this versions and tags :/

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: the timeout?

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: then we should do it for azure as well
we can do it afterwords I would say.
as simply changing these tags does not mean that it will work.

sebbrandt87:
I can just after that create an issue to update it to use variables everywhere, but also for azure then.

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: and if so, it would retry ;)

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
hectorj2f: Yes we should do it there too, so we can bump versions without changing the chart that often

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
hectorj2f: I meant to have an args property in values.yaml where you have these settings. Perhaps in a bigger cluster they need to be customize due to higher timeouts. I can do it in a follow-up pr

sebbrandt87: I can do it as well, this is not the point ;)
but we do not need it right now to get it running

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: sure

stable/gcpdisk-csi-driver/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+appVersion: 0.6.0
+description: Google Compute Engine Persistent Disk (GCE PD) Container Storage Interface (CSI) Storage Plugin
+name: gcpdisk-csi-driver
+maintainers:
+  - name: sebbrandt87
+version: 0.6.0
+kubeVersion: ">=1.16.0"
```
hectorj2f: cannot it be used with 1.15 🤔  ? we didn't face this restriction for our existing addons :/

sebbrandt87: we did for azure as well, pointed it to 0.15.0
we should not have it backward compatible out of my perspective, as we implement it right now and not in the past.

hectorj2f: we didn't https://github.com/mesosphere/kubeaddons-configs/blob/master/templates/azuredisk-csi-driver.yaml#L15

hectorj2f: Konvoy 1.3 supports 1.15 and 1.16(default), if this addon is part of 1.16 we need to leverage a logic to avoid deploying it on 1.15 cluster. At this moment, this logic is not in place, so that will be a blocker for this addon. IF you confirm it only works on 1.16, if it can work on 1.15 I don't see why to set this value here.

hectorj2f: I'd also be interested on see what it is blocking this chart to run on 1.15 🤔

sebbrandt87: nothing blocks it, changed it

PR #310, opened 2019-12-11
[kommander] feat: add default folder for kommander

natmegs:
When kommander is installed, add a default folder (namespace).

Disclaimer I'm not sure this is the correct way to do this! Goal is to have this yaml file applied at install time.

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -170,6 +170,16 @@ grafana:
     #   hosts:
     #   - grafana.example.com
 
+  folders: 
+    enabled: true
+    default: 
+      namespace: kommander-default-folder
+      labels:
+        kommanderType: folder
+      annotations:
+        kommander.d2iq.io/description: Default folder
+        kommander.d2iq.io/display-name: Default
+
   ## Configure grafana datasources
   ## ref: http://docs.grafana.org/administration/provisioning/#datasources
   ##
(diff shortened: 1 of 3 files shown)


stable/kommander/values.yaml
```diff
@@ -170,6 +170,16 @@ grafana:
     #   hosts:
     #   - grafana.example.com
 
+  folders: 
+    enabled: true
```
natmegs: Added this in case we need to feature flag this for some reason, but I think adding the namespace is pretty harmless for now

stable/kommander/values.yaml
```diff
@@ -170,6 +170,16 @@ grafana:
...
+    default: 
+      namespace: default-folder
```
brandonc: Since namespaces are cluster scope, I'm wondering if this should be more product-specific like kommander-default-folder ?

alejandroEsc:
fyi: ```[01:30:33][Step 1/1] stable/kommander/values.yaml
[01:30:33][Step 1/1]   173:11    error    trailing spaces  (trailing-spaces)
[01:30:33][Step 1/1]   175:13    error    trailing spaces  (trailing-spaces)```

DanielMSchmidt:
Added do not merge as we have not merged our folders feature branch

juliangieseke:
@natmegs i think this is obsolete now that we have everything in our repo.

makkes:
I'm closing this now. Please feel free to re-open if needed.

PR #312, opened 2019-12-12
allow federating kube-system resources

jr0d:
We have need to override the ingress record for the dashboard when a cluster joins federation. The dashboard is currently being deployed to the kube-system namespace. If it is not acceptable to federate objects in kube-system, perhaps we could migrate the dashboard to kubeaddons.


jaxesn:
This makes sense to me, based on where the dashboard is running.  My only concern is whether this is too heavy a solution for a namespace like kube-system?  I assume nothing is really affected other than the federated types within that namespace, but just want to be sure.

PR #320, opened 2019-12-17
traefik: add support for custom certificate

jayapriya90:


jayapriya90:
Merging this change to make it to the release

PR #321, opened 2019-12-20
kommander-karma: Use upstream karma chart, handle cm during upgrade/uninstall

gracedo:
Use the latest version of the upstream karma chart and remove our fork. Add pre-install annotations to the karma configmap so that it isn't overwritten during upgrades (KCL controller updates the configmap whenever a cluster is provisioned), and add a pre-delete hook job to delete the configmap at uninstallation.

Fixes https://jira.mesosphere.com/browse/DCOS-62056

Once this is merged:
1. Another mesosphere/charts PR needs to be opened which bumps the kommander-karma subchart dependency (to v0.3.0) within the kommander chart
1. Then a kubeaddons-configs PR which bumps the kommander chart.

## Testing
This was tested by building and self hosting the kommander-karma chart, then doing the same with kommander. I `konvoy up`-ed using a kubeaddons-configs branch pointing to my self-hosted kommander chart.

image

To test that the cleanup job works, disable the kommander addon and redeploy addons. Check that the configmap no longer exists. You should also be able to find logs that the cleanup job ran and deleted the cm:

image


stable/kommander-karma/templates/cleanup-configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    metadata:
+      name: karma-configmap-cleanup
+    spec:
+      serviceAccountName: {{ .Values.kommanderServiceAccount }}
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
```
hectorj2f: we should update this docker image we are using 1.16 or 1.15.5

stable/kommander-karma/Chart.yaml
```diff
@@ -3,6 +3,7 @@ appVersion: "1.0"
 description: Kommander Karma
 name: kommander-karma
 home: https://github.com/mesosphere/charts
-version: 0.2.0
+version: 0.3.0
```
hectorj2f: is kommander-karma by any other chart _? Perhaps we should also update the chart that was using staging/karma

stable/kommander-karma/templates/cleanup-configmap.yaml
gracedo: should I make that update everywhere this is used?

stable/kommander-karma/Chart.yaml
gracedo: yep, it'll be used by the kommander chart - but it depends on this pr getting merged so I was going to open up a new pr for kommander once this is out!

PR #322, opened 2019-12-23
kudo: add subresources status in instance CRD

zmalik:


alenkacz:
Btw. how are charts release? Are they immutable? I just wonder if we don't have to bump the version as well

zmalik:
> Btw. how are charts release? Are they immutable?

CI failure seems to indicate that we need to bump the chart version

PR #327, opened 2020-01-03
kommander: Add centralized 1.16 clusters dashboard

gracedo:
Due to the changes in metric names in kubernetes 1.16, we are adding a new grafana dashboard to the centralized grafana running on Kommander in management clusters. This dashboard is the same as the kubernetes-mixin cluster dashboard, with queries adapted to accommodate multiple clusters instead of just one. I also went ahead and replaced the previous dashboard with this one, as I saw that the cpu metric `namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate` was missing from both 1.15 and 1.16 clusters. On the other hand, the new 1.16 dashboard showed metrics from both 1.15 and 1.16 managed clusters.

Let me know if anything should be omitted or any other dashboards that could be useful.

## Testing
I have built this chart and hosted it on my personal GH. You may test these changes by setting the `configVersion` in your cluster.yaml to `gracedo/test_kommander_1.16_dash`, which points to that hosted chart. Run `konvoy up` and [provision some clusters](https://github.com/mesosphere/kommander-cluster-lifecycle#creating-konvoy-clusters). Go to `/kommander/monitoring/grafana` and check out the updated `Kubernetes / Compute Resources / Clusters` dashboard.

The below screenshots show data coming from 3 managed clusters.
image
image
image
image
image
image
image


alejandroEsc:
Just realized, this change would not be backwards compatible right? So would only apply to newer versions of konvoy right? Not a big deal just checking.

gracedo:
@alejandroEsc yeah, it would apply to managed clusters running newer versions of konvoy/k8s - I imagine with GA only newer versions would be used?

alejandroEsc:
> @alejandroEsc yeah, it would apply to managed clusters running newer versions of konvoy/k8s - I imagine with GA only newer versions would be used?

Yep, but the base addons repo should be versioned so we should be ok, just keeping this in mind that's all.

PR #330, opened 2020-01-06
[kommander] remove older versions of kubernetes

jaxesn:
These versions fail to install, I think because they are using older versions of konvoy configured [here](https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/kommander/1.163.x/kommander-1.yaml#L51).  We could maybe change that to use the latest konvoy? I'm not sure we need these versions so I am ok just removing as options.

thoughts @jimmidyson ?

### DIFF
--- a/stable/kommander/values.yaml
+++ b/stable/kommander/values.yaml
@@ -7,7 +7,7 @@ logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
 hostPollingInterval: 3000
 clusterPollingThrottleFactor: 1
-kubernetesVersionsSelection: '["1.16.3", "1.15.6", "1.15.5", "1.15.4"]'
+kubernetesVersionsSelection: '["1.16.4", "1.16.3", "1.15.6"]'
 kubeaddonsAddress: kommander-kubeaddons-kubeaddons-catalog
 
 # Mode must be either production|konvoy
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 - name: alejandroEsc
 - name: jimmidyson
 name: kommander
-version: 0.3.8
+version: 0.3.9


stable/kommander/values.yaml
```diff
@@ -7,7 +7,7 @@ logoutRedirectPath: /ops/landing
 clusterPollingInterval: 3000
 hostPollingInterval: 3000
 clusterPollingThrottleFactor: 1
-kubernetesVersionsSelection: '["1.16.3", "1.15.6", "1.15.5", "1.15.4"]'
```
alejandroEsc: What's the process we are using to vet these items and how do we maintain this list?

jaxesn: The end goal is for this list to go away and kcl to be querying dockerhub to get released versions of konvoy + kubernetes and generate it dynamically.  In the meantime, we have just be manually updating this list.

The UI currently has the ability to query an api in kcl to get this list dynamically, however, kcl itself is using a mapping between kubernetes version to config tag.  I believe this mapping exists because at the time of adding the default handling in kcl, the tagging scheme in kubeaddon-configs was not one that supported automatically figuring out which tag to use for a given kubernetes version.  I also believe that one of the goals of adding the new repo was to support this use, tho I dont know how.

Because the mapping still exist in the kommander addon config, I requested that the UI team not switch to the dynamic api in hopes to avoid some kind of mismatch where there is a new konvoy version, but its not listed in the kcl mapping so if the user chooses it, it will fail out.

That said, what we should probably do is move this configuration to the addon itself so its closer to the kcl mapping and therefore more likely to be in sync. (if this would be better, @DanielMSchmidt please take this over)

This was not really the question you asked, but I've had a few people ask for the background on this and I figured this was as good a place as any to document it. cc @DanielMSchmidt @juliangieseke @jimmidyson

jimmidyson:
needs a rebase...

jaxesn:
@dkoshkin i went ahead and added the new 1.16.4 to the mapping for the kommander UI. We'll also need to update this mapping as well before we cut the RC release.

https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/kommander/1.173.x/kommander-1.yaml#L53

PR #331, opened 2020-01-08
[traefik-forward-auth] reuse secret for session key

jr0d:
TFA now uses sessions to store group claims


jr0d:
https://github.com/mesosphere/kubernetes-base-addons/pull/56

jkoelker:
Does the `image.tag` also need to be bumped in `values.yaml`?

PR #334, opened 2020-01-08
Fix grafana internal urls

br-lewis:
Addresses: https://jira.mesosphere.com/browse/DCOS-53598

### DIFF
--- a/staging/prometheus-operator/values.yaml
+++ b/staging/prometheus-operator/values.yaml
@@ -448,7 +448,7 @@ grafana:
     hosts: []
 
     ## Path for grafana ingress
-    path: /
+    path: /ops/portal/grafana
 
     ## TLS configuration for grafana Ingress
     ## Secret must be manually created in the namespace
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.3.3
+version: 8.3.4
 appVersion: 0.34.0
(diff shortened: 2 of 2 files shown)


staging/prometheus-operator/values.yaml
```diff
@@ -448,7 +448,7 @@ grafana:
     hosts: []
 
     ## Path for grafana ingress
-    path: /
+    path: /ops/portal/grafana
```
alejandroEsc: Could I get more info why this change was made and why here? 

This does not seem like the right place to do this as it is the values file. Did you mean to make this change in the Kubernetes-base-addons repo?

br-lewis: This change addressed a bug where grafana would create incorrect urls to other dashboards within itself: https://jira.mesosphere.com/browse/DCOS-53598. I made the commit in this repo because I didn’t think about the kubernetes-base-addons repo. In retrospect I could see it belonging there instead.

jimmidyson: The less changes to the forked prometheus-operator chart the better. Overriding in base addons is better IMO.

gracedo: +1 on fewer changes to the forked chart. Also, this ingress is actually disabled so it doesn't look like we're actually adding it. Doesn't look like we set is in [kubernetes-base-addons](https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/0.34.x/prometheus-1.yaml#L251) either, so just checking if this would actually fix the bug? Looks like the ingress is actually configured in the [`opsportal` chart](https://github.com/mesosphere/charts/blob/dev/stable/opsportal/values.yaml#L31)

jimmidyson: So should we revert this PR since it was merged a few hours ago... @br-lewis thoughts?

PR #337, opened 2020-01-09
Revert "Fix grafana internal urls"

br-lewis:
This reverts commit 97d023b2c998bb0976a58907ead4591da7ce2c6b.
Reverted because changes were made in the wrong repo.


br-lewis:
The CI check fails because this is dropping the version from 8.3.4 to 8.3.3 which already exists, but it's a revert so the end result is the same as 8.3.3. Should we ignore the check? I think that would make more sense than adding 8.3.5 which is just 8.3.3 again.

PR #340, opened 2020-01-09
kommander: bump kcl and clean up chart

makkes:
Since a lot of the resources in this chart have been moved to the kcl chart itself, I removed them from this chart.


jimmidyson:
I think there will be some stuff in values to clean up too?

makkes:
> ?

Fixed.

jimmidyson:
https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2534551&_focus=442

> [ERROR] templates/: render error in "kommander/templates/hooks-kubeaddons.yaml": template: kommander/templates/hooks-kubeaddons.yaml:1:14: executing "kommander/templates/hooks-kubeaddons.yaml" at <.Values.kubefed.enab...>: can't evaluate field enabled in type interface {}

stable/kommander/requirements.yaml
```diff
@@ -1,14 +1,10 @@
 dependencies:
-- name: kubefed
```
alejandroEsc: So kubefed is being removed? Where is this applied now? and wouldnt some crds need to be removed now? what are the assumptions here?

makkes: That stuff has been moved to the kcl repohttps://github.com/mesosphere/kommander-cluster-lifecycle/pull/313

PR #342, opened 2020-01-10
[test] Add chart install and upgrade testing

jkoelker:
* Adds a testing script that tests the charts using `ct lint-and-install --upgrade`.
* Backports a fix from upstream `dex` chart to allow install and upgrade to succeed (https://github.com/helm/charts/pull/18380)
* Clean up EOL whitespace

### DIFF
--- a/stable/dex/templates/config-openssl.yaml
+++ b/stable/dex/templates/config-openssl.yaml
@@ -8,10 +8,6 @@ metadata:
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
   name: {{ template "dex.fullname" . }}-openssl-config
-  annotations:
-    "helm.sh/hook": post-install
-    "helm.sh/hook-weight": "1"
-    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
 data:
   openssl.conf: |
 {{ .Files.Get "config/openssl.conf" | indent 4 }}
--- a/staging/cert-manager-setup/Chart.yaml
+++ b/staging/cert-manager-setup/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.1.6
+version: 0.1.7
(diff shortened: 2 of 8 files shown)


test/ct-e2e.yaml
```diff
@@ -0,0 +1,38 @@
+debug: true
+target-branch: dev
+chart-dirs:
+  - stable
+  - staging
+excluded-charts:
+  - common
+  - dex-controller # Moved to a different helm repo
+  - azuredisk-csi-driver  # DCOS-62804
```
jkoelker: Its my hope that once all the tickets are closed we can remove the special *e2e* variant of the config and both linting and install and upgrade tests can share just the one config.

stable/dex/templates/config-openssl.yaml
```diff
@@ -8,10 +8,6 @@ metadata:
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
   name: {{ template "dex.fullname" . }}-openssl-config
-  annotations:
```
alejandroEsc: Why these changes to dex? Could you provide some context, just want to know why we are no longer using this hook, its a fairly large change so want to be careful.

jkoelker: The context is in the commit message for this commit. Basically this causes the configmap to be be deleted immediatly after creation, and the `grpc-tls` and `web-tls` jobs which depend on it fail, causing the install to fail. This is the backport from https://github.com/helm/charts/pull/18380, but since the charts have diverged enough I couldn't just `cherry-pick` it ;(.

staging/cert-manager-setup/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.1.6
+version: 0.1.7
```
jkoelker: Technically not really a change to the chart, but `ct` lint detects the new `ci/test-values.yaml` that the e2e test script uses as a change, so...

PR #343, opened 2020-01-11
[dispatch] Update version

cprovencher:


alejandroEsc:
Are we sure we are trying chart version to app version?

GoelDeepak:
Did we do the upgrade testing and verify it passes?

dkoshkin:
This is outdated as there is 0.5.2 already. Is this needed if the chart is being used in kubernetes-base-addons is coming from the dispatch repo?

jbarrick-mesosphere:
This is not needed as we maintain our chart here: https://github.com/mesosphere/dispatch/tree/master/charts/dispatch

This chart is only still here in case there are still people using Konvoy 1.2.

PR #344, opened 2020-01-13
stable/opsportal: bump nginx container to 1.17-alpine

mhrabovcin:
JIRA: https://jira.mesosphere.com/browse/DCOS-62720

Upgrade from `1.15-alpine` which contains critical security issues.

### DIFF
--- a/stable/opsportal/templates/landing.yaml
+++ b/stable/opsportal/templates/landing.yaml
@@ -51,7 +51,7 @@ spec:
     spec:
       containers:
         - name: opsportal-landing
-          image: nginx:1.15-alpine
+          image: nginx:1.17-alpine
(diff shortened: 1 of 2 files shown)


stable/opsportal/templates/landing.yaml
```diff
@@ -51,7 +51,7 @@ spec:
     spec:
       containers:
         - name: opsportal-landing
-          image: nginx:1.15-alpine
+          image: nginx:1.17-alpine
```
sreis: There's a known issue with [Alpine based container images](https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues) but I guess this is fine because it's only used to serve static content.

PR #353, opened 2020-01-14
traefik: add objects depending on useCertManager

armandgrillet:
Tested via `helm install /Users/armand/Code/charts/staging/traefik --dry-run --values=/Users/armand/Desktop/traefik.yaml`.

### DIFF
--- a/staging/traefik/Chart.yaml
+++ b/staging/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 1.72.10
+version: 1.72.11
 appVersion: 1.7.12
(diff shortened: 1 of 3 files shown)


br-lewis:
Shouldn't that be if not using a cert manager rather than only initializing when a cert manager is in use?

timaa2k:
This only relates to cert-manager so it will only work if cert-manager is running @br-lewis 
I'm not entirely sure if we captured everything, there may be the temporary secret somewhere in the charts that could/should also be conditional for cert-manager.

timaa2k:
Ah no nvm, this is good because the temporary secret is generated by cert-manager itself :)

PR #347, opened 2020-01-13
[kommander-karma] [kommander-thanos] Modify UI links

gracedo:
https://jira.mesosphere.com/browse/DCOS-62153
We are modifying karma and thanos (and grafana, done in followup PR) links to be behind `/ops/portal/kommander/monitoring/<>` instead of `/kommander/monitoring/<>`. Those ingress changes are done here along with chart bumps. After this PR is merged, I will open up a followup PR to do the following:

1. bump the thanos/karma dependencies in the Kommander chart
1. change the Kommander ui to point to `/ops/portal/kommander/ui` (see https://jira.mesosphere.com/browse/DCOS-62690 and we can close this PR https://github.com/mesosphere/charts/pull/338)
1. modify the grafana ingress to point to the new url
1. bump kommander chart version

Following the above PR, I will close out this ticket by bumping kommander to the updated version^ in `kubernetes-base-addons`.

## Testing
This was tested by hosting kommander-thanos and kommander-karma charts on my personal GH, then pointing kommander to my hosted charts as well as self hosting the kommander chart. I changed kommander to point to my hosted kommander chart in a branch of `kubernetes-base-addons` and used that to deploy a konvoy cluster. You can test this by using that branch: `gracedo/test_kommander_path_changes`. See below for the scrnshts of new urls working as expected:

image
image
image
image

PR #354, opened 2020-01-14
kommander-karma: Remove trailing slash from ingress path

gracedo:
My bad, the trailing slash shouldn't have been added - fixing this here and will follow the same protocol as https://github.com/mesosphere/charts/pull/347

### DIFF
--- a/stable/kommander-karma/README.md
+++ b/stable/kommander-karma/README.md
@@ -32,7 +32,7 @@ karma:
       traefik.ingress.kubernetes.io/auth-type: "forward"
       traefik.ingress.kubernetes.io/auth-url: "http://traefik-forward-auth-kubeaddons.kubeaddons.svc.cluster.local:4181/"
       traefik.ingress.kubernetes.io/priority: "2"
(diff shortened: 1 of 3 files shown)


stable/kommander-karma/README.md
```diff
@@ -32,7 +32,7 @@ karma:
       traefik.ingress.kubernetes.io/auth-type: "forward"
       traefik.ingress.kubernetes.io/auth-url: "http://traefik-forward-auth-kubeaddons.kubeaddons.svc.cluster.local:4181/"
       traefik.ingress.kubernetes.io/priority: "2"
-    path: "/ops/portal/kommander/monitoring/karma/"
+    path: "/ops/portal/kommander/monitoring/karma"
```
samvantran: wasn't this needed bc there was a bug of some sort? or this was due to the kommander-ui hijacking?

gracedo: there is another bug with the slash but we were able to sidestep it via our traefik/karma settings

PR #357, opened 2020-01-15
[nvidia] chore: chart improvements

alejandroEsc:
Working on improving this chart

- allow to enable/disable different subcharts explicitly
- remove unnecessary files
- clean values files
- changes to grafana chart (image below)

helm install:
```
$ helm install --name nvidia-kubeaddons --namespace kubeaddons ./nvidia/
NAME:   nvidia-kubeaddons
LAST DEPLOYED: Tue Jan 14 21:55:22 2020
NAMESPACE: kubeaddons
STATUS: DEPLOYED

RESOURCES:
==> v1/ConfigMap
NAME                       AGE
nvidia-kubeaddons-grafana  0s

==> v1/DaemonSet
NAME                                    AGE
nvidia-kubeaddons-nvidia-dcgm-exporter  0s
nvidia-kubeaddons-nvidia-device-plugin  0s
nvidia-kubeaddons-nvidia-driver         0s

==> v1/Pod(related)
NAME                                          AGE
nvidia-kubeaddons-nvidia-dcgm-exporter-82bw4  0s
nvidia-kubeaddons-nvidia-device-plugin-rkdtc  0s
nvidia-kubeaddons-nvidia-driver-gfjbz         0s
```

helm list:
```
$ helm list
NAME                 	REVISION	UPDATED                 	STATUS  	CHART                    	APP VERSION	NAMESPACE
nvidia-kubeaddons    	1       	Tue Jan 14 21:55:22 2020	DEPLOYED	nvidia-0.3.0             	0.2.0      	kubeaddons
opsportal-kubeaddons 	1       	Mon Jan 13 14:26:43 2020	DEPLOYED	opsportal-0.1.25         	1.0.0      	kubeaddons
prometheus-kubeaddons	2       	Mon Jan 13 15:36:11 2020	DEPLOYED	prometheus-operator-8.3.7	0.34.0     	kubeaddons
```

Screen Shot 2020-01-14 at 9 41 24 PM


staging/nvidia/charts/grafana/templates/configmap-dashboards.yaml
```diff
@@ -1,12 +1,8 @@
-# Generated from 'nodes' from https://raw.githubusercontent.com/coreos/kube-prometheus/master/manifests/grafana-dashboardDefinitions.yaml
-# Do not change in-place! In order to change this file first read following link:
-# https://github.com/helm/charts/tree/master/stable/prometheus-operator/hack
-{{- if .Values.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "grafana.fullname" . }}
-  namespace: {{ $.Values.namespace }}
+  namespace: {{ .Release.Namespace }}
```
Gilbert88: Could you educate me what ".Release.Namespace" represents?

we expect both still in kubeaddons namespace so that they can work with Prometheus and grafana

staging/nvidia/charts/grafana/values.yaml
```diff
@@ -1,7 +1,2 @@
-#
-# grafana dashboards
-#
-enabled: true
```
Gilbert88: are users still able to disable this chart in cluster.yaml?

staging/nvidia/values.yaml
```diff
@@ -1,4 +1,23 @@
...
-  domain: example.com
+grafana:
+  enabled: true
+
+nvidia-dcgm-exporter:
+  enabled: true
+  resources: {}
+  nodeSelector: {}
+  tolerations: []
+  initContainers: []
```
Gilbert88: could you educate me what is the relation between these fields and the corresponding ones in subchart's values.yaml?

staging/nvidia/charts/grafana/templates/configmap-dashboards.yaml
alejandroEsc: Usually you do not want to explicitly control placement of deployment like this, via value files. Since we are deploying the exporter and grafana charts in the same namespaces of kubeaddons (or where ever prometheus is deployed), this would presumably where you would tell helm to deploy this chart. You can then leverage helm's parameters to deploy these resources in this manner.

staging/nvidia/charts/grafana/values.yaml
alejandroEsc: yes, if you look at the updated values file ou will see that the grafana yaml stanza has the enabled field. You can set this to false and it will not deploy this chart. You may test it out by running helm install/template

staging/nvidia/values.yaml
alejandroEsc: You will notice that each stanza corresponds to each subchart. This is how you override values for those subcharts, and I have chose to make this explicitly shown. The enabled fields along with the requirement file allow you then to disable subcharts.

PR #360, opened 2020-01-15
kommander: Update kommander-karma dependency

gracedo:
I'm not sure what version of kommander-karma chart was pulled in on the previous PR (https://github.com/mesosphere/charts/pull/356) 🤔 but it was definitely an old version that was still setting the karma ingress path to `/kommander/monitoring/karma` instead of `ops/portal/kommander/monitoring/karma`. I re-ran `helm dependency update` here and tested that this should be the correct version of kommander-karma - tested by hosting the kommander chart on my GH and installing that version to a konvoy cluster. This will unblock https://github.com/mesosphere/kubernetes-base-addons/pull/68 and we'll update kommander to 0.3.13 there


alejandroEsc:
I see that an old version is deleted, but i do not see the new binary in this PR. @gracedo can you check as you need to explicitly add this binary.

gracedo:
@alejandroEsc I think it looks like it was deleted, but it was actually modified (see file here in branch https://github.com/gracedo/charts/blob/gracedo/kommander_karma_ingress_bump_fix/stable/kommander/charts/kommander-karma-0.3.2.tgz)

alejandroEsc:
just tested it out, yes i see what you mean. OK LGTM.

juliangieseke:
hm, I _think_ I've merged this PR while it being out of date, chart version already was at `0.3.14` on `dev` - we should require PRs to be up-to-date to be able to merge them. cc @joejulian @alejandroEsc  (I've changed that in settings - feel free to revert if you disagree)

PR #364, opened 2020-01-16
chore(kommander): bump KCL to v0.3.4

jimmidyson:


stable/kommander/requirements.lock
```diff
@@ -4,7 +4,7 @@ dependencies:
   version: 0.1.4
 - name: kommander-cluster-lifecycle
   repository: https://mesosphere.github.io/kommander-cluster-lifecycle/charts
-  version: v0.3.3
+  version: 0.3.4
```
hectorj2f: v0.3.4 ?

stable/kommander/values.yaml
```diff
@@ -52,31 +52,9 @@ federate:
...
 
-  certificates:
-    issuer:
-      name:
-      kind: Issuer
-      selfSigned: false
-  konvoy:
-    allowUnofficialReleases: false
-  kubeaddonsRepository:
-    versionStrategy: repository-tag  # One of: `repository-tag`, `mapped-kubernetes-version`, `master`,
```
hectorj2f: I'll miss these comments. Perhaps we have some documentation where they are explained :)

stable/kommander/requirements.lock
jimmidyson: This is what `helm dependency update` did so I trust it, plus I've installed the chart and it works fine.

stable/kommander/values.yaml
jimmidyson: They are in the original chart.

stable/kommander/requirements.lock
jimmidyson: All other dependencies are without the `v` prefix.

makkes: Yeah, I stumbled upon that every time. 🤷‍♂

PR #369, opened 2020-01-17
[custom certs] Updates charts to support custom certificate

GoelDeepak:
This patch is to improve the flow around using custom certificate with Konvoy. 

We recently added a feature in Konvoy in which it could create extra Kubernetes objects if an operator places them in `extras/kubernetes` directory.  For a custom certificate, we will use this feature of Konvoy. The expectation is that the operator would create a secret from the custom certificate files and then specify the secret name `caSecretName` as part of the values for each addon in `cluster.yaml` 

This PR changes the chart of 4 addons to use `caSecretName` if it is specified 

Additionally, this PR also displays the custom certificate on the `/token` output. Prior to this PR, it use to always display Kubernetes CA even if an operator is using custom certificate


GoelDeepak:
I have manually carried out two tests:
1. Functional testing - Verified everything is working as expected
2. Upgrade testing - Verified that charts are upgradable

staging/dex-k8s-authenticator/values.yaml
```diff
@@ -73,6 +73,8 @@ resources: {}
 
 caCerts:
   enabled: false
```
hectorj2f: `enabled: false ?`

staging/traefik/values.yaml
```diff
@@ -136,7 +136,7 @@ ssl:
   # tlsMinVersion: VersionTLS12
   # https://docs.traefik.io/configuration/entrypoints/#strict-sni-checking
   # sniStrict: false
-  customCert: false
```
hectorj2f: what about doing something like https://github.com/mesosphere/charts/pull/369/files#diff-84491f379c2eed7f94384116a9e49e45R74 with a boolean and a secret name ?

staging/dex-k8s-authenticator/html-templates/linux-mac-common.html
```diff
@@ -90,8 +90,8 @@ <h3>Run configuration commands</h3>
       <img class="clippy" width="13" src="{{ .Web_Path_Prefix }}static/clippy.svg" alt="">
     </button>
     <pre><code class="hljs">kubectl config set-context {{ .Username }}-{{ .ClusterName }} \
-    --cluster={{ .ClusterName }} \
-    --user={{ .Username}}-{{.ClusterName }}</code></pre>
+    --cluster={{ .Username }}-{{ .ClusterName }} \
+    --user={{ .Username }}-{{.ClusterName }}</code></pre>
```
hectorj2f: `-{{ .ClusterName }}`

staging/dex-k8s-authenticator/html-templates/linux-mac-common.html
```diff
@@ -8,7 +8,7 @@ <h3>Copy IDP CA Certificate From URL</h3>
       <button class="btn" style="float:right" data-clipboard-snippet="">
         <img class="clippy" width="13" src="{{ .Web_Path_Prefix }}static/clippy.svg" alt=""/>
       </button>
-      <pre><code>curl --create-dirs -s {{ .IDPCaURI }} -o ${HOME}/.kube/certs/{{ .ClusterName }}/idp-ca.crt</code></pre>
+      <pre><code>curl --create-dirs -s {{ .IDPCaURI }} -o ${HOME}/.kube/certs/{{ .Username }}-{{ .ClusterName }}/idp-ca.crt</code></pre>
```
hectorj2f: question: where is the Username coming from ?

hectorj2f: Would that work when attaching existing Kubernetes clusters like GKE and deploying addons there ?

staging/dex-k8s-authenticator/values.yaml
GoelDeepak: yeah, the default is false. It is enabled by the operator if they want to use custom certs

staging/traefik/values.yaml
GoelDeepak: Sorry, didn't understand your comment. This variable was introduced by us and this patch is taking it out

alejandroEsc:
Going to merge as we need this for tuesday and we need to test changes against these asap.

staging/traefik/values.yaml
hectorj2f: I meant, why don't you respect the same structure and have a complex object with a boolean and a secretName as done in https://github.com/mesosphere/charts/pull/369/files#diff-84491f379c2eed7f94384116a9e49e45R74

staging/dex-k8s-authenticator/html-templates/linux-mac-common.html
hectorj2f: @alejandroEsc ^

staging/traefik/values.yaml
GoelDeepak: What is the benefit of that structure?

PR #372, opened 2020-01-20
chore: Cleanup master branch

jimmidyson:
In conjunction with #373, this will keep the master branch clean.


jimmidyson:
The failing check is because it shouldn't run against PRs to master (which are not common but necessary for this PR).

PR #373, opened 2020-01-20
chore: Do not add chart sources to master

jimmidyson:
This works in combination with the master cleanup (see #372) to keep a clean master branch, with only chart repo packages and no chart sources.

Please merge #372 first so I don't have to rebase that, although that wouldn't be too hard anyway.


Makefile
```diff
@@ -61,9 +61,9 @@ publish:
 	rm -rf $(NON_DOCS_FILES)
 	git checkout $(GIT_REF) -- $(NON_DOCS_FILES)
 	$(MAKE) GIT_REF=$(GIT_REF) all
+	git add -A docs
```
jimmidyson: Moved this here, other the `git clean` was deleting newly created charts... oops!

PR #377, opened 2020-01-23
Remove duplicate Grafana dashboards

branden:
This PR fixes an error that causes the prometheus-operator chart to generate configmaps with conflicting names.

This removes dashboards that are duplicated by these dashboards introduced with #309:

* https://github.com/mesosphere/charts/blob/2e5603b4ff1c2490923a2673ff4580a11a7a0a55/staging/prometheus-operator/templates/grafana/dashboards-1.14/apiserver.yaml
* https://github.com/mesosphere/charts/blob/2e5603b4ff1c2490923a2673ff4580a11a7a0a55/staging/prometheus-operator/templates/grafana/dashboards-1.14/controller-manager.yaml
* https://github.com/mesosphere/charts/blob/2e5603b4ff1c2490923a2673ff4580a11a7a0a55/staging/prometheus-operator/templates/grafana/dashboards-1.14/kubelet.yaml
* https://github.com/mesosphere/charts/blob/2e5603b4ff1c2490923a2673ff4580a11a7a0a55/staging/prometheus-operator/templates/grafana/dashboards-1.14/scheduler.yaml

## Testing

I tested this by deploying a Konvoy cluster with the prometheus addon enabled, then deployed another cluster with the prometheus addon disabled, and deployed this modified chart to the second cluster with the addon's chart values. I then compared the affected dashboards on both clusters and observed that they looked the same.

PR #378, opened 2020-01-23
Retry requests when setting Grafana's home dashboard

branden:
https://jira.mesosphere.com/browse/DCOS-62216

This fixes an issue that may cause the Grafana home dashboard to be empty. If the job that sets Grafana's home dashboard runs before Grafana's API is ready, it will fail after one attempt. This PR changes the job to retry requests that fail with transient errors.

## Testing

I tested this PR by deploying a Konvoy cluster with the prometheus and prometheusadapter addons disabled, deploying the prometheus-operator chart from this branch, browsing to `/ops/portal/grafana`, and observing that the home dashboard is set to "Kubernetes / Compute Resources / Cluster".

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.3.8
+version: 8.3.9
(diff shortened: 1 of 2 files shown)


branden:
CI is giving this PR the same spurious error as https://github.com/mesosphere/charts/pull/377#issuecomment-577478076. Likewise, `make ct.lint` passes when I run it locally.

branden:
The CI error is fixed by https://github.com/mesosphere/charts/pull/379

PR #381, opened 2020-01-24
chore: bump kommander ui

juliangieseke:


stable/kommander/requirements.lock
```diff
@@ -4,7 +4,7 @@ dependencies:
   version: 0.1.4
 - name: kommander-cluster-lifecycle
   repository: https://mesosphere.github.io/kommander-cluster-lifecycle/charts
-  version: 0.3.4
+  version: v0.3.5
```
juliangieseke:
```suggestion
  version: 0.3.5
```

sebbrandt87: well that is what `helm dependency update` did :P 
and it is correct https://github.com/mesosphere/kommander-cluster-lifecycle/releases

sebbrandt87:
Releases manual checked on a cluster. 
Working! 8)

natmegs:
We tested and confirmed that we can provision aws clusters with both static credentials and roles-based credentials 👍

PR #383, opened 2020-01-25
[opsportal] remove application ingress records

alejandroEsc:
Removed ingress records of apps from opsportal, these will be now produced in each application. Added deprecation notice, and remove ingress records.

https://jira.d2iq.com/browse/DCOS-63429

### DIFF
--- a/stable/opsportal/Chart.yaml
+++ b/stable/opsportal/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.0.0
 home: https://github.com/mesosphere/charts
 description: OpsPortal Chart
 name: opsportal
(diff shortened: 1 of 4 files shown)


makkes:
@alejandroEsc, would you mind linking a JIRA ticket here? I don't quite understand the reasoning behind this change (not that I oppose it, I'd just like to understand).

alejandroEsc:
> @alejandroEsc, would you mind linking a JIRA ticket here? I don't quite understand the reasoning behind this change (not that I oppose it, I'd just like to understand).

https://jira.d2iq.com/browse/DCOS-63429

PR #385, opened 2020-01-28
[tfa] add rbac option

jr0d:
Adds option to enable rbac, passthrough. Also, this makes impersonation optional. 

relates to https://github.com/mesosphere/traefik-forward-auth/pull/12

ticket: https://jira.d2iq.com/browse/D2IQ-63101


GoelDeepak:
@jr0d Could you please provide a Jira and a small description of this change?

staging/traefik-forward-auth/templates/clusterrole.yaml
```diff
@@ -35,3 +35,12 @@ rules:
 - apiGroups: [""]
   resources: ["services"]
   verbs: ["get", "list", "watch"]
```
samvantran: you didn't touch this by why is this listed differently than the others in this yaml?

alejandroEsc: @jr0d could you bring the above to the same style as what you added in? please and thanks!

staging/traefik-forward-auth/templates/deployment.yaml
```diff
@@ -105,13 +110,7 @@ spec:
...
                   name: {{ template "traefik-forward-auth.fullname" . }}-secret
                   key: secret
-            {{- if .Values.traefikForwardAuth.allowedUser.valueFrom.secretKeyRef }}
-            - name: WHITELIST
-              valueFrom:
-                secretKeyRef:
-                  name: {{ required "traefikForwardAuth.allowedUser.valueFrom.secretKeyRef.name is required" .Values.traefikForwardAuth.allowedUser.valueFrom.secretKeyRef.name }}
-                  key: {{ required "traefikForwardAuth.allowedUser.valueFrom.secretKeyRef.key is required" .Values.traefikForwardAuth.allowedUser.valueFrom.secretKeyRef.key }}
-            {{- else if gt (len .Values.traefikForwardAuth.whitelist) 0 }}
+            {{- if gt (len .Values.traefikForwardAuth.whitelist) 0 }}
```
alejandroEsc: os if secretKeyRef is not defined, the whitelist will not be defined either? before there was an else if, is this what you intend?

staging/traefik-forward-auth/templates/clusterrole.yaml
jr0d: 👍

staging/traefik-forward-auth/templates/deployment.yaml
jr0d: yes, the ops-portal user no longer needs to be added to the whitelist since the user is also bound to the cluster-admin role and will be permitted by the default access policy.

samvantran:
Can you look into setting up the install/upgrade test for tfa? see [here](https://github.com/mesosphere/charts/blob/master/test/ct-e2e.yaml#L28). 

From what i can tell `make test` (run in our CI) actually runs the [e2e-kind tests](https://github.com/mesosphere/charts/blob/master/test/e2e-kind.sh) so maybe that's the place to add testing. If it's too much work, let's look into it soon after merging this

alejandroEsc:
> Can you look into setting up the install/upgrade test for tfa? see [here](https://github.com/mesosphere/charts/blob/master/test/ct-e2e.yaml#L28).
> 
> From what i can tell `make test` (run in our CI) actually runs the [e2e-kind tests](https://github.com/mesosphere/charts/blob/master/test/e2e-kind.sh) so maybe that's the place to add testing. If it's too much work, let's look into it soon after merging this PR

I am ok putting this off with a jira ticket capturing this work for now.

PR #384, opened 2020-01-28
kommander-thanos/karma: Cleanup job hook weights

gracedo:
Change the cleanup job hook weights for kommander-thanos and kommander-karma, so that sa/clusterroles are all set up prior to these cleanup scripts to avoid permission errors.

## Testing
Please see the Testing section of https://github.com/mesosphere/charts/pull/386

PR #386, opened 2020-01-28
kommander: Set Grafana home dashboard in Kommander Grafana

gracedo:
This creates a Job to set the Kommander-specific Grafana home dashboard to the Clusters summary dashboard. I was unable to get a post-install hook of this Job working - the job wasn't getting triggered at all (if anybody has any ideas why this is happening, let me know!). This job has retry logic so it is given plenty of time (10m) to complete (dependent on Grafana service coming up). This job was adapted from our [post-install-hook job in prometheus-operator](https://github.com/mesosphere/charts/blob/master/staging/prometheus-operator/templates/mesosphere-hooks/post-install-hooks.yaml) which sets the home dashboard for the kubeaddons grafana instance. I also added in cleanup pre-delete hook jobs for all of the resources that are created in the pre-install hooks (cm with script to set the home dashboard, ops-portal secret that the script depends on), which requires clusterroles setup so that they have the correct permissions to handle these resource types.

This work is dependent on https://github.com/mesosphere/charts/pull/384 so that the karma/thanos cleanup hooks occur after these clusterroles are created. Once that PR is merged, I will bump the karma/thanos dependencies here.

Once this is merged, I will open a PR with base-kubeaddons to bump the Kommander chart.

## Testing
I tested this by building https://github.com/mesosphere/charts/pull/384 on my personal GH repo and pointing kommander to those versions of kommander-thanos and kommander-karma. Then I built/published that version of kommander on my personal repo and created an addons branch referencing my repo to install kommander. You can use that branch to test on a Konvoy cluster: `gracedo/test_kommander_grafana_home_dash_hook`

Here is the `grafana-home-dashboard` job run that sets the home dashboard:
image

```
$ kubectl -n kommander get cm grafana-home-dashboard -oyaml
apiVersion: v1
data:
  run.sh: |-
    #!/bin/bash
    set -o nounset
    set -o errexit
    set -o pipefail
    CURL="curl --verbose --fail --max-time 30 --retry 20 --retry-connrefused"
    DASHBOARD_ID=$($CURL -H "X-Forwarded-User: $X_FORWARDED_USER" http://kommander-kubeaddons-grafana.kommander/api/dashboards/uid/efa86fd1d0c121a26444b636a3f509a8 | jq '.dashboard.id')
    echo "setting home dashboard to ID" $DASHBOARD_ID
    $CURL -X PUT -H "Content-Type: application/json" -H "X-Forwarded-User: $X_FORWARDED_USER" -d '{"homeDashboardId":'"$DASHBOARD_ID"'}' http://kommander-kubeaddons-grafana.kommander/api/org/preferences
kind: ConfigMap
metadata:
  creationTimestamp: "2020-01-28T22:34:52Z"
  name: grafana-home-dashboard
  namespace: kommander
  resourceVersion: "71396"
  selfLink: /api/v1/namespaces/kommander/configmaps/grafana-home-dashboard
  uid: 0af52c3e-b4b7-456c-9eec-a8a82a55f9ba

$ kubectl -n kommander get secrets ops-portal-credentials
NAME                     TYPE     DATA   AGE
ops-portal-credentials   Opaque   2      14m
```

I also verified that going to `<cluster_url>/ops/portal/kommander/monitoring/grafana` redirects to the home dashboard `Kubernetes / Compute Resources / Clusters`.

At deletion, all of these resources are cleaned up:
```
$ kubectl -n kommander get cm grafana-home-dashboard -oyaml
Error from server (NotFound): configmaps "grafana-home-dashboard" not found
$ kubectl -n kommander get secrets ops-portal-credentials
Error from server (NotFound): secrets "ops-portal-credentials" not found
```


stable/kommander/values.yaml
```diff
@@ -78,6 +78,14 @@ kubeaddons-catalog:
 grafana:
   enabled: true
 
+  hooks:
+    jobName: grafana-home-dashboard
+    image: dwdraju/alpine-curl-jq
+    secretKeyRef: ops-portal-credentials
+    serviceURL: http://kommander-kubeaddons-grafana.kommander
+    homeDashboardUID: efa86fd1d0c121a26444b636a3f509a8
```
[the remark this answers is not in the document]
samvantran: Answered my own question [here](https://github.com/mesosphere/charts/blob/ac3a752f90d80dfb81fe3fe1909e20214067f631/stable/kommander/templates/grafana/dashboards/k8s-resources-clusters.yaml#L2550) but i'm wondering if we ever update the chart does the uid change or its set in stone forever. If it changes, we should query by name but if not, then this is fine.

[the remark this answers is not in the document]
gracedo: The UID would only change if a dev went in there and changed that field - but it's the same with the name, they're [both fields set on the same json](https://github.com/mesosphere/charts/blob/ac3a752f90d80dfb81fe3fe1909e20214067f631/stable/kommander/templates/grafana/dashboards/k8s-resources-clusters.yaml#L2549-L2550). Personally it felt cleaner to me to grab a dashboard by the UID that we set on the dashboard (since we are in charge of creating that dashboard and we know it exists) rather than querying for a name/string (in dcos-monitoring, we grab dashboard by UID). The api call itself also just looks cleaner to me `/api/search/?query=Kubernetes+%2F+Compute+Resources+%2F+Cluster` vs `/api/dashboards/uid/{{ .Values.grafana.hooks.homeDashboardUID`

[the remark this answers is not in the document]
branden: I'm ok with querying for the dashboard by its UID. I think the only downside is that you can't tell at a glance which dashboard we're setting. How about a comment here that mentions the name of the dashboard?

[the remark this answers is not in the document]
samvantran: Comment is a good compromise. We may add more dashboards in the future so this helps a human to identify the dashboard without having to grep through all for the uid

stable/kommander/values.yaml
```diff
@@ -78,6 +78,15 @@ kubeaddons-catalog:
 grafana:
   enabled: true
 
+  hooks:
+    jobName: grafana-home-dashboard
```
samvantran: Can we make the job name a little clearer? Eg `set-kommander-grafana-home-dashboard` or `set-mgmt-grafana...`. Reason being there is an identical job that runs to set the konvoy home dashboard and this would help to disambiguate on the cluster.

stable/kommander/templates/grafana/hooks-home-dashboard.yaml
```diff
@@ -0,0 +1,80 @@
...
+# Unable to get post-install job hook working, which is why
+# this is a regular Job. The retries in the configmap script
+# should ensure that this successfully runs once the Grafana
+# server is up.
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Values.grafana.hooks.jobName | quote }}
+  namespace: {{ .Release.Namespace }}
+  labels:
```
samvantran: Should we add a hook delete policy for `hook-succeeded`? No need to keep the job and the pods around if everything worked

gracedo: Unfortunately I was not able to get the hook working, so it's just a regular job

stable/kommander/values.yaml
gracedo: The jobs are running in separate namespaces, but I'll change the name here just to make it clearer. Maybe next time I go into prometheus-operator I'll make some improvements there too, that job is currently ambiguously named `post-install-job`

stable/kommander/templates/grafana/hooks-home-dashboard.yaml
samvantran: oh i see, i thought it was just the post-install hook but it was all hooks. Okay, all good.

stable/kommander/templates/grafana/opsportal-credentials-secret.yaml
```diff
@@ -0,0 +1,55 @@
...
+    spec:
+      containers:
+        - name: kubectl
+          # --export flag is deprecated so we need to stick with this kubectl version
+          image: bitnami/kubectl:1.16.2
+          imagePullPolicy: IfNotPresent
+          command:
+            - /bin/sh
+            - -c
+            - kubectl get secret ops-portal-credentials --namespace=kubeaddons --export -o yaml | kubectl apply --namespace={{ .Release.Namespace }} -f -
```
gracedo: @jr0d @jongiddy I also just want to make sure it's okay to be copying the `ops-portal-credentials` secret from `kubeaddons` ns to `kommander` ns. The secret is needed to get the user to pass into the `X-Forwarded-User` header for Grafana requests

jr0d: @gracedo it does allow anyone with secret view privileges in the kommander namespace access to the ops portal password; which they wouldn't have had before. While more it's more difficult, perhaps consider only copying the username. 

What would happen access for the ops-portal user was disabled?

gracedo: Made the change to only copy over the username https://github.com/mesosphere/charts/pull/386/commits/0a2615e24b4cb259f3df350992544b656073ad17

stable/kommander/templates/grafana/opsportal-username-secret.yaml
```diff
@@ -0,0 +1,55 @@
...
+    spec:
+      containers:
+        - name: kubectl
+          # --export flag is deprecated so we need to stick with this kubectl version
+          image: bitnami/kubectl:1.16.2
+          imagePullPolicy: IfNotPresent
+          command:
+            - /bin/sh
+            - -c
+            - kubectl create secret generic {{ .Values.grafana.hooks.secretKeyRef }} -n {{ .Release.Namespace }} --from-literal=username=$(kubectl get secret ops-portal-credentials --namespace=kubeaddons --export -o jsonpath="{.data.username}" | base64 --decode)
```
jr0d: :+1:

samvantran:
Got a merge conflict but beyond that LGTM

gracedo:
@samvantran @jr0d rebased and fixed conflicts, can you guys +1 again?

stable/kommander/templates/grafana/hooks-home-dashboard.yaml
alejandroEsc: Please put a jira ticket to review this. we should be cleaning up after jobs.

stable/kommander/values.yaml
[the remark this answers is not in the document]
alejandroEsc: at least we plumbed it to the values file, good call. I am also hesitant about this but we can figure this out later.

stable/kommander/templates/grafana/hooks-home-dashboard.yaml
gracedo: https://jira.d2iq.com/browse/D2IQ-63670

gracedo:
Going to go ahead and merge this to unblock https://github.com/mesosphere/kubernetes-base-addons/pull/100

PR #387, opened 2020-01-29
chore: add enabled property to deploy tests

hectorj2f:
We were deploying the test resources, that should not happen on production.

### DIFF
--- a/staging/traefik/templates/tests/test-configmap.yaml
+++ b/staging/traefik/templates/tests/test-configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.testFramework.enabled  }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -12,4 +13,4 @@ data:
     @test "Test Access" {
       curl -D - http://{{ template "traefik.fullname" . }}/
     }
-
+{{end}}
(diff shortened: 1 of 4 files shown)


staging/traefik/templates/tests/test-configmap.yaml
```diff
@@ -12,4 +13,4 @@ data:
     @test "Test Access" {
       curl -D - http://{{ template "traefik.fullname" . }}/
     }
-
+{{end}}
```
samvantran: This doesn't need to be `{{- end}}`? You started the `{{- if` that way

staging/traefik/templates/tests/test.yaml
```diff
@@ -40,3 +41,4 @@ spec:
   - name: tools
     emptyDir: {}
   restartPolicy: Never
+{{end}}
```
samvantran: same here

hectorj2f: thx good catch ;)

staging/traefik/templates/tests/test-configmap.yaml
hectorj2f: thx good catch ;)

PR #391, opened 2020-02-01
nfs-server-provisioner: add base chart

jieyu:
This chart is derived from the upstream chart:
https://github.com/helm/charts/tree/master/stable/nfs-server-provisioner

The reason we put it here is because the upstream chart seems
unmaintained. The stable head is broken. We need to apply the following
patch to make it work:
https://github.com/helm/charts/pull/19260

Tested locally with a Konvoy cluster:

```bash
helm install --name nfs-server-provisioner .
```

And apply the following PVCs:

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: nfs-test
spec:
  storageClassName: nfs
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 100Mi
```

```bash
k get pvc
NAME         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
nfs-test     Bound    pvc-c587797b-be9a-47c7-95f3-0bb74aa66332   100Mi      RWX            nfs            11m
```

### DIFF
--- a/staging/nfs-server-provisioner/Chart.yaml
+++ b/staging/nfs-server-provisioner/Chart.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+appVersion: 2.3.0
+description: nfs-server-provisioner is an out-of-tree dynamic provisioner for Kubernetes. You can use it to quickly & easily deploy shared storage that works almost anywhere.
+name: nfs-server-provisioner
+version: 0.6.0
+maintainers:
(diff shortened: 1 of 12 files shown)


joejulian:
I'll just point out, again, that this is not an enterprise-capable storage option. It should not be used in production.

PR #392, opened 2020-02-01
nfs-server-provisioner: move pvc to the correct ns

jieyu:
The PVC object of the statefulset should be in the same namespace as the
statefulset itself. This patch fixed the issue.


joejulian:
The StatefulSet controller always creates PVCs in the same namespace: https://github.com/kubernetes/kubernetes/blob/60044a8accfe8fddeb452c4adca4fefbbff64967/pkg/controller/statefulset/stateful_set_utils.go#L145

joejulian:
The StatefulSet controller already creates PVCs in the same namespace as the StatefulSet.

jieyu:
Oops. I think when I was testing, there is a left over pvc with the same name. I incorrectly think that’s the one created by the current statefulset. Thanks Joe. Close this PR

PR #396, opened 2020-02-03
[kommander, opsportal] Add clusterroles for well known endpoints

jr0d:
# Description
This patch adds roles for protected endpoints under /ops/portal. This is a temporary solution for konvoy 1.3.1 RBAC. In 1.4 we will align these roles with the chart which defines the ingress.

## Why the temporary solution
For kibana, we intended to wrap the upstream chart, however, this strategy would introduce a breaking change which is not acceptable for a bugfix release.

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.3.23
+version: 0.3.24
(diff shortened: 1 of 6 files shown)


GoelDeepak:
@jr0d aren't the `ops-portal` endpoints already moved out of `ops-portal` even in 1.3.1?

alejandroEsc:
@jr0d please try to fill in the necessary labels. Could we split out the kommander parts and opsporrtal parts into two different prs?

jr0d:
@alejandroEsc Thanks for taking a look. I wrote a program to generate the roles; so I'll add the annotations to the generator and update the PR. Also, I'll create separate PRs as you suggested.

PR #397, opened 2020-02-04
kommander: Fix upgrade bug

gracedo:
There's a [bug](https://mesosphere.slack.com/archives/CN5S0UKAA/p1580762260057800) in upgrading kommander chart to >=0.3.21
```
$ kubectl -n kommander describe pod set-kommander-grafana-home-dashboard-td4s2
...  
Warning  Failed     57s (x8 over 2m20s)  kubelet, ip-10-0-128-187.us-west-2.compute.internal  Error: secret "ops-portal-username" not found
```
I added a `pre-upgrade` hook to the `copy-ops-portal-username` job to ensure that the username secret is copied over in upgrades as well as installs. Also modified the kubectl command to ensure that it does not fail if the resource already exists.

## Testing [WIP]
I tested this by hosting the chart on my GH. I `konvoy up` with base addons version `stable-1.15.6-4`, then redeployed with the hosted updated kommander chart 0.3.25 (`kubernetes-base-addons` branch is `gracedo/test_kommander_grafana_upgrade`). I also tested upgrade path from `0.3.24`->`0.3.25` to ensure that the copy job doesn't fail if the secret already exists.

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.3.24
+version: 0.3.25
--- a/stable/kommander/templates/grafana/opsportal-username-secret.yaml
+++ b/stable/kommander/templates/grafana/opsportal-username-secret.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
 {{ include "kommander.labels" . | indent 4 }}
   annotations:
-    "helm.sh/hook": pre-install
+    "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-weight": "5"
     "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
 spec:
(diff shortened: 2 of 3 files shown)


gracedo:
I haven't actually finished testing this (hence the `wip` 😅 ) -- I was able to get these cases working:
1) upgrade from kommander `0.3.20` (version prior to setting the home dashboard--where the `ops-portal-username` secret was introduced) to `0.3.25` (this version hosted on my HG)
2) upgrade from kommander `0.3.24` to `0.3.25` (making sure the kubectl command works even if the secret already exists)

However, upgrading from `kubernetes-base-addons` of version `stable-1.15.6-4` to `0.3.25` doesn't work - deployment times out. Was debugging this and found our favorite error:

```
[tiller] 2020/02/04 04:52:34 warning: Upgrade "kommander-kubeaddons" failed: kind ConfigMap with the name "set-kommander-grafana-home-dashboard" already exists in the cluster and wasn't defined in the previous release. Before upgrading, please either delete the resource from the cluster or remove it from the chart
```
Rollback likewise fails with
```
[tiller] 2020/02/04 04:52:35 warning: Rollback "kommander-kubeaddons" failed: kind ClusterRoleBinding with the name "kommander-kubeaddons" already exists in the cluster and wasn't defined in the previous release. Before upgrading, please either delete the resource from the cluster or remove it from the chart
```

Not sure why these errors come up on `stable-1.15.6-4` (I think that's on kommander chart `0.3.14`) but not `0.3.20` which is still before the home dashboard job was added. Looking into how to avoid these errors, will put up another PR with a fix

gracedo:
**Update to the above comment**
Put up a new cluster and tested this upgrade path again, I'm not getting the same above errors, possibly those were due to the chart getting into a bad state from all my testing. The upgrade is still timing out - I see Grafana restarting with the updated version, so that part of the kommander upgrade (grafana upgrade that I'm testing for) seems to work, but the overall upgrade of kommander times out so the post-upgrade hook to set the home dashboard doesn't trigger. Perusing the tiller logs shows these two errors continuously, which may be causing the kommander upgrade timeout? 
```
[kube] 2020/02/04 06:00:23 Deployment is not ready: kommander/kommander-kubeaddons-kommander-ui
[kube] 2020/02/04 06:00:24 Deployment is not ready: kubeaddons/opsportal-kubeaddons-kommander-ui
```

PR #402, opened 2020-02-05
chore: add traefik 2 chart

joejulian:
Copy in the upstream chart: git@github.com:containous/traefik-helm-chart.git

The chart isn't served by them so this is for us to start with.


GoelDeepak:
I personally have resistance to moving to Traefik 2.0 just yet. Maybe we end up with Traefik 2.0 but I want to run a comparison between Nginx, Traefik, and Envoy before making this move. The reason being ingress is an important component and any change would probably be a big undertaking both in terms of development and operation (upgrades and all). I also didn't like the direction that Traefik 2.0 took with custom CRDs which is a big departure from Kubernetes ingress APIs. I would like to know how Nginx and Envoy have taken

joejulian:
This is just adding the chart for hosting. It's not modified from upstream in any way and I have no plans on moving forward with creating an addon at this time. I am going to do some testing of it on my home cluster though.

staging/traefik2/README.md
```diff
@@ -0,0 +1,15 @@
+# Traefik
+
+[Traefik](https://traefik.io/) is a modern HTTP reverse proxy and load balancer made to deploy
+microservices with ease.
+
+**NOTE: THIS CHART IS IN INCUBATOR STATUS, DO NOT USE IN A PRODUCTION ENVIRONMENT**
```
samvantran: scary note

alejandroEsc: No addons will be created from this today. @joejulian wants to test this out. I am also using this chart as a test bed for a few things.

jimmidyson:
Why can't we just use the chart from upstream repo if we're not changing it?

joejulian:
It's not hosted anywhere and the way I wanted to work with it required it be hosted. If it's really a big deal, I can just host it myself.

PR #409, opened 2020-02-05
Use consistent naming for opsportal roles

jr0d:
* Change naming convention to `opsportal-<component>-view|edit|admin`


alejandroEsc:
> * Change naming convention to `opsportal-<component>-view|edit|admin`

can we get an idea about the motivation for this change?

stable/opsportal/templates/ingress-opsportal-roles.yaml
```diff
@@ -32,7 +32,6 @@ rules:
 - nonResourceURLs:
   - {{ .Values.opsportalRBAC.path | trimSuffix "/"}}
   - {{ .Values.opsportalRBAC.path | trimSuffix "/" }}/*
-  - {{ .Values.opsportalRBAC.graphqlPath | trimSuffix "/"}}
```
dkoshkin: Is this intentional?

dkoshkin: Don't want the original message to get lost.

jr0d: Yes, it was redundant

jieyu:
@joejulian the role name is a user facing thing (i.e., if user want to grant some user to opsportal, they'll have to use that name).

By default, the release name will include `-kubeaddons` suffix (e.g., `opsportal-kubeaddons-admin`). The `kubeaddons` part is really unfortunate and should not be known by the user.

I think for all the opsportal related roles, we should use consistent naming scheme like the following so that it's intuitive to the user.
`opsportal-<COMPONENT>-admin|edit|view`

cc @chrisgaun @samba @GoelDeepak

joejulian:
Is there any chance that someone would ever want this installed twice? Different departments, or projects, or something that might want to manage their own permission sets?

jieyu:
@joejulian maybe. We could in the future add a configurable option to add a prefix to those role names to solve that use case. So far, we don't have that use case.

joejulian:
https://v2.helm.sh/docs/chart_best_practices/#resource-naming-in-templates

Please update the _PR description_ to explain why we're not following best practices then I'll change my review.

joejulian:
How did this get merged without the change request being addressed?

GoelDeepak:
wouldn't this naming scheme fail during the upgrade? It will conflict with the existing object

jieyu:
@GoelDeepak fail during upgrade from 1.3.0 to 1.3.1?

joejulian:
Fail during upgrade from 0.2.2 to 0.2.3.

GoelDeepak:
I was referring to future failures. Say we release it and then in future if we are trying to upgrade this chart. Would there be an issue?

jieyu:
@joejulian 0.22 is not in any release.

@GoelDeepak no, it won't be an issue.

GoelDeepak:
> @GoelDeepak no, it won't be an issue.

Then I am good 👍

joejulian:
0.2.2 isn't released as an addon, sure, but the question remains because the 0.1.34 chart was and it had different names for the objects. The same upgrade question still applies.

jieyu:
the cluster roles are newly introduced in 0.2.x. So upgrading from 0.1.34 shouldn't be a problem

GoelDeepak:
@joejulian might be confused with the diff. @joejulian none of these changes are released yet. The changes that you see in this PR is from one unreleased version to another unreleased version

PR #410, opened 2020-02-07
Add Dispatchfile

cprovencher:


Dockerfile.dispatch
```diff
@@ -0,0 +1,5 @@
+FROM mesosphere/dispatch-dind:v0.4.4
+
+MAINTAINER Mesosphere <devprod@mesosphere.com>
+
+RUN apt-get update -y && apt-get install -y git build-essential
```
shaneutt: technically you don't need a `-y` on `apt-get update` and if you want to make the builds less noisy by turning off progress bars, you can use `-qq` on install:

```suggestion
RUN apt-get update && apt-get install -qq git build-essential
```

Just a suggestion, not a blocker. :+1:

cprovencher: thanks!

PR #411, opened 2020-02-07
[kubefed] enable the use of cert-manager to produce web-hook certs

alejandroEsc:
First pass at enabling cert-manager to produce certificates for webhooks

- [X] Tested upgrade from v0.0.2
- [X] Tested regular deployment

The above work as expected and cert-manager was able to create certs if enabled. 

Both the above were tested using the values field:
```
  webhook:
    annotations: {}
    certManager:
      enabled: true
```
in otherwords the upgrade was with this new value set and using cert-manager


staging/kubefed/values.yaml
```diff
@@ -33,10 +33,19 @@ controllermanager:
     adoptResources:
   ## Value of feature gates item should be either `Enabled` or `Disabled`
   featureGates:
-    PushReconciler:
-    SchedulerPreferences:
-    CrossClusterServiceDiscovery:
-    FederatedIngress:
+    featureGates:
```
jimmidyson: This is repetition of featureGates property above. I don't think we need this change at all and shouldn't set the featureGates values, leave blank so this is easier to upstream.

staging/kubefed/values.yaml
```diff
@@ -33,10 +33,19 @@ controllermanager:
...
+      CrossClusterServiceDiscovery: "Disabled"
+      FederatedIngress: "Enabled"
+      PushReconciler: "Enabled"
+      SchedulerPreferences: "Disabled"
+
+  controllermanager:
```
jimmidyson: Weird nesting - `controllermanager.controllermanager`... If we're going to have separate annotations for webhook and controllermanager, then either we need to add a webhook subchart or we need to pull it all up to the top level chart otherwise this is just weird looking config.

staging/kubefed/values.yaml
```diff
@@ -33,10 +33,19 @@ controllermanager:
...
+    annotations: {}
+
+  webhook:
```
jimmidyson: `controllermanager.webhook` is also a bit weird - would be better to have webhook toplevel.

staging/kubefed/values.yaml
alejandroEsc: Sure, I left these for testing in our CI. Locally i could not test because these values needed to be filled in. I will attempt to see if we can test.

staging/kubefed/values.yaml
alejandroEsc: > Weird nesting - `controllermanager.controllermanager`... If we're going to have separate annotations for webhook and controllermanager, then either we need to add a webhook subchart or we need to pull it all up to the top level chart otherwise this is just weird looking config.

```
controllermanager <- subchart
  controllermanager <- the deployment
```
this isnt weird, its just unfortunate naming and probably abuse of subcharting tbh.

staging/kubefed/values.yaml
alejandroEsc: > `controllermanager.webhook` is also a bit weird - would be better to have webhook toplevel.

As i answered above, this is just unfortunate, not weird. If i were to change this chart i would not have a subchart in this way, among other changes.

alejandroEsc:
> Some comments - I've submitted a PR (#412) to your branch with the changes I'd like. Let's discuss later.

I do not want to accept that PR in this PR for two reasons. 1. I do not want to continue to use subcharting in this fashion. I do not think that is the right approach. 2. That is a huge deviation from upstream, and one that may be accepted, but at the same time i do not want to have this change here as the change we are doing now (allow for cert-manager)  is large enough as is.

staging/kubefed/values.yaml
jimmidyson: Testing in CI can be done by specifying values files for various CI configurations - no need for this to be in values to test in CI.

staging/kubefed/values.yaml
jimmidyson: It is odd... if it's the deployment then it should be `controllermanager.deployment`.

staging/kubefed/values.yaml
jimmidyson: The use of subcharts like this is to be able to enable/disable controllermanager separately - I don't see that as an abuse, but assume they are doing so for development purposes for some reason.

Webhook isn't part of controllermanager - it's a separate deployment/component in this environment so it is weird to nest it under controllermanager :)

jimmidyson:
I do agree that this is a deviation from upstream and guess it would be much better to do this upstream first, but as this is a blocker then I guess let's go with this approach (with my noted misgivings above) and get this fixed upstream ASAP.

staging/kubefed/charts/controllermanager/templates/webhook.yaml
```diff
@@ -137,3 +150,49 @@ type: kubernetes.io/tls
...
+  selfSigned: {}
+---
+# Generate a CA Certificate used to sign certificates for the webhook
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: {{ .Release.Name}}-root-certificate
+spec:
+  secretName: {{ .Release.Name }}-root-ca
+  duration: 43800h # 5y
```
shaneutt: would we potentially want to lower this? Keeping this long could result in unintended consequences for the end-user, theoretically the following scenario:

1. PK is compromised by a bad actor
2. bad actor generates a bunch of bad certs
3. any client that trusts the CA (and maybe doesn't get a memo about the breach) continues to do so for up to 5 years

By declaring shorter CA's we can help to reduce some risks in our end-users environments and help make breach mitigations a bit more manageable by having very conservative defaults.

Let me know your thoughts on this :thinking:

staging/kubefed/values.yaml
alejandroEsc: it is under the controllermanager subchart, as such it belongs nested and not top level.

jimmidyson: I know, what I'm saying is that is odd - controllemanager and webhook are two different components and if a subchart is used for controllermanager then it makes sense to make a separate subchart for webhook.

alejandroEsc:
> It appears that none of the [standard labels](https://helm.sh/docs/topics/chart_best_practices/labels/#standard-labels) are included in this chart. Would this be a good time to add them?

I can if you would permit me another PR to do so.

staging/kubefed/values.yaml
alejandroEsc: > I know, what I'm saying is that is odd - controllemanager and webhook are two different components and if a subchart is used for controllermanager then it makes sense to make a separate subchart for webhook.

with the internal logic of moving the controller to a separate subchart, yes i agree with you, it is odd. 

I have little idea why this was done this way unless the intent was that the subchart can stand alone? 

Regardless I understand your point and the reason for doing the PR as you have added. My only concern was the timing and what upstream chart work would be accepted... I guess i need a point of contact and get some bearing there.

PR #414, opened 2020-02-07
(Kommander GA) update kommander ui with workspaces

juliangieseke:
holding back until clarified how to proceed. KCL 0.4.0 got merged which is intended for Kommander GA, this Bump would add related UI bits.

Given that Konvoy UI shipped with `opsportal` addon, it _should be fine_ to merge. we'll do some more testing tho.


stable/kommander/requirements.lock
```diff
@@ -7,7 +7,7 @@ dependencies:
   version: 0.4.0
 - name: kommander-ui
   repository: https://mesosphere.github.io/kommander/charts
-  version: 1.192.3
```
shaneutt: On bytes this looks like a big version change, I imagine it's not so serious? Have we tested the upgrade path from the previous release? Kommander is not GA so I don't necessarily consider that a blocker but it would be good if we documented whether or not it's functional, especially since we place this chart under `stable/`.

juliangieseke: it is a serious change, but it also is stable with kcl 0.4.0. 
see this threads for clarification if we can safely merge or not:
https://mesosphere.slack.com/archives/CQL8757S7/p1581093615261300
https://mesosphere.slack.com/archives/CLETZF5U6/p1581090929413100
(tl'dr: I'm not sure if 0.4.0 should have been merged)

shaneutt: Roger that, let me know if there's any help you need from @mesosphere/sig-ksphere-catalog to move things along!

juliangieseke:
Closing, changes are in #419

PR #416, opened 2020-02-07
Bump Dex to v2.22.0

jieyu:
Updated Dex chart to be based on v2.22.0 (latest) release. I also bump the Chart to the latest stable/dex chart (with patches in `patches/` applied).

Re-applied all the D2iQ themes.

Tested with latest Konvoy release by upgrading the dex chart from the earlier version to this version (including with connectors configured).


stable/dex/patches/00-extra-role-and-rolebindings.patch
```diff
@@ -0,0 +1,61 @@
+commit a5158051fc8980586639b500abd1e1bd3352686a
```
joejulian: good idea! I like including the patch files so we can more easily track upstream.

stable/dex/values.yaml
```diff
@@ -159,12 +208,17 @@ config:
...
-#    username: "admin"
-#    userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
+#  staticPasswords:
+#   - email: "admin@example.com"
+#     # bcrypt hash of the string "password"
+#     hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+#     username: "admin"
+#     userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
+
+  frontend:
```
GoelDeepak: should we move this setting to `kubernetes-base-addons`?  I like the pattern where we keep the chart close to upstream and have d2iq specific settings in `kubernetes-base-addons`

stable/dex/values.yaml
```diff
@@ -3,14 +3,16 @@
 # Declare name/value pairs to be passed into your templates.
 # name: value
 
-image: quay.io/dexidp/dex
-imageTag: "v2.17.0"
+image: mesosphere/dex
```
GoelDeepak: We are setting these values in `Kuberenetes-base-addons`. Should we continue to do that? It helps keep the chart close to the upstream and also it is easier to control these settings from `kubernetes-base-addons`

stable/dex/values.yaml
jieyu: @GoelDeepak i was debating that initially myself. I decided to bake in the chart because this is a fork anyway (i.e., D2iQ flavor of dex) with some other changes (e.g., certificate, etc.).

stable/dex/values.yaml
jieyu: See my above comment

GoelDeepak: Maybe for consistency and to avoid any confusion we just follow the style of keeping the chart close to upstream version and modifying it in `kubernetes-base-addons`. The flip side of it is that this diff should always be carry over all subsequent bumps. I also want to avoid the case where somebody mistakenly updates the image here but not in `kubernetes-base-addons`.

jieyu: @GoelDeepak I would remove the image repo/tag in the kubernetes-base-addons for dex so that there's only one place we set the image repo/tag so there will be no confusion where to bump the image.

GoelDeepak: Sorry @jieyu for pushing on this. I ideally prefer there than here just because every time you do a dex bump you would have to remember to port over this change. It is much better to not have to do that and be taken care automatically in `kubernetes-base-addons`. One less thing to worry about. Is there any benefit of keeping it here?

jieyu: hum, ok, that's a fair point. You're saying that if do in addon config, in many case, we don't need to bump the chart.

sebbrandt87:
closes #416

PR #420, opened 2020-02-10
[opsportal] All authenticated

jr0d:
This PR introduces an option to bind all authenticated users to the opsportal-admin role. This is needed to maintain parity with existing functionality where authenticated users have unrestricted access to the opsportal.


stable/opsportal/values.yaml
```diff
@@ -50,6 +50,13 @@ opsportalRBAC:
   enabled: true
```
GoelDeepak: nit: should we make rbac `false` by default,  for now? It is anyway enabled in `Kubernetes-base-addons`

joejulian: Let's leave the chart values at what is recommended. We can override that in Addons if we want to maintain compatibility with previous deployments.

stable/opsportal/values.yaml
```diff
@@ -50,6 +50,13 @@ opsportalRBAC:
...
   path: /ops/portal
   graphqlPath: /ops/portal/graphql
+  ## traefik-forward-auth 2.0 introduces RBAC support and now requires that users, regardless of whitelist,
```
shaneutt: docs++

PR #430, opened 2020-02-14
Add Dispatchfile

cprovencher:
Move CI to Dispatch
https://jira.d2iq.com/browse/D2IQ-63990


Dispatchfile
```diff
@@ -0,0 +1,72 @@
...
+    return k8s.corev1.EnvVarSource(secretKeyRef=k8s.corev1.SecretKeySelector(localObjectReference=k8s.corev1.LocalObjectReference(name=name), key=key))
+
+gitResource(git,
+    url="$(context.git.url)",
+    revision="$(context.git.commit)")
+
+dindTask("publish", inputs=[git], steps=[
+  v1.Container(
+    name = "publish-charts",
+    image = "mesosphere/dispatch-dind:mesosphere-charts",
```
makkes: Where does this image come from? Would you mind adding a link and explanation here so that we know where to look? I personally would prefer to have the Dockerfile and build process for it in this repo so everything is self-contained.

test/kind-config.yaml
```diff
@@ -3,3 +3,20 @@ apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
   - role: worker
+kubeadmConfigPatches:
```
shaneutt: could we get a comment for these patches please to make it easier for future code readers to quickly know the context?

Dispatchfile
cprovencher: @makkes good catch, I actually forgot to remove that line!

cprovencher:
@shaneutt @makkes 
I've addressed your requested changes!
I've also added changes in test/e2e-kind.sh so please take a look at that took.

PR #437, opened 2020-02-17
fix: add trailing slashes

juliangieseke:
seems like theres a config setting which solves our trailing slash issues: https://docs.traefik.io/v1.7/configuration/commons/#main-section

⚠️ I'm not sure if this breaks some of our addons since it seems like having `path: some/path/w/o/railing/slash` will break now.


juliangieseke:
might be that my ingress just was messed up…

PR #444, opened 2020-02-19
chore: bump kubeaddons-catalog version

hectorj2f:
We bumped the kubeaddons-catalog with a bug fix for the next MWT


stable/kommander/requirements.lock
```diff
@@ -1,10 +1,10 @@
 dependencies:
 - name: kubeaddons-catalog
   repository: https://mesosphere.github.io/charts/staging
-  version: 0.1.4
+  version: 0.1.5
 - name: kommander-cluster-lifecycle
   repository: https://mesosphere.github.io/kommander-cluster-lifecycle/charts
-  version: 0.4.6
+  version: v0.4.6
```
juliangieseke: out of curiosity, why are you adding this v? is it a version thing in helm or something? 🤔 only observed that v being added in your PRs 🤷‍♂

hectorj2f: 🤔 unless it is the helm version, i cannot tell

PR #445, opened 2020-02-19
[chore] update kommander catalog api

shaneutt:
This PR updates the kommander catalog API to the latest version, and updates the chart.yaml with new sources and metadata.

This helps to resolve [D2IQ-64450](https://jira.d2iq.com/browse/D2IQ-64450).


shaneutt:
> General thing: Id love to have charts for our own repos hosted in these repos, not here. Something for post ga tho…

If a JIRA to do so doesn't exist, we should make one perhaps?

PR #454, opened 2020-02-24
feat(cert-manager-setup): make certificates more customizable

sebbrandt87:
- added missing specs for kind:Certificate

[D2IQ-64608]

[D2IQ-64608]:https://jira.d2iq.com/browse/D2IQ-64608


shaneutt:
@sebbrandt87 can we move this from `wip` to `ready` yet? I just wanted to make sure I checked on the full context.

shaneutt:
@sebbrandt87 it seems there are some lingering questions here, is this PR still needed or is it now safe to close?

sebbrandt87:
@shaneutt needed.
Haven't found the time to answer the question at all, as I wanted to add more details instead of just saying. "Yes I did tests." ;)
Please leave as is I want to come back to it shortly.

shaneutt:
> @shaneutt needed.
> Haven't found the time to answer the question at all, as I wanted to add more details instead of just saying. "Yes I did tests." ;)
> Please leave as is I want to come back to it shortly.

@sebbrandt87 if you've tested this locally and are confident in it, I would be OK with merging this as is to get the improvements in, and we can reference a JIRA for the follow up tests. If that would be preferable for you let me know otherwise we'll stand by to help.

shaneutt:
SGTM just let us know how we can help

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -29,5 +32,16 @@ spec:
...
 {{- end }}
+{{- if .ipAddresses }}
+  ipAddresses:
+  {{- range .ipAddresses }}
+    - {{ . | quote -}}
+  {{- end }}
+{{- end }}
+  keyAlgorithm: {{ .keyAlgorithm | default "rsa" }}
+  keyEncoding: {{ .keyAlgorithm | default "pkcs1" }}
+  keySize: {{ .keyAlgorithm | default "2048" }}
```
mhrabovcin: Should you use `.keySize` and `.keyEnconfig` ?

sebbrandt87: `. keyEncoding` is for encoding a private key
`. keySize` is the size of the private key.

but I see the copy paste mistake 🤦‍♂ 🙇 `.keyAlgorithm`

sebbrandt87:
As you can see now I have added values to the `ci/` directory to the yaml file already present for that purpose and so it is verified that the deployment works.
I just implemented what is available in the CRD already, but which we where not able to define via this addon. We are now able to do so.

staging/cert-manager-setup/ci/general-test-values.yaml
```diff
@@ -20,6 +20,15 @@ certificates:
     dnsNames:
       - example.com
       - www.example.com
+    ipAddresses:
+      - 127.0.0.1
+    keyAlgorithm: rsa
+    keyEncoding: pkcs8
+    keySize: 4096
+    organization:
```
mhrabovcin: Lets name this as "organizations" so that its clear that multiple can be provided.

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -29,5 +32,21 @@ spec:
...
+{{- if .ipAddresses }}
+  ipAddresses:
+  {{- range .ipAddresses }}
+    - {{ . | quote -}}
+  {{- end }}
+{{- end }}
+  keyAlgorithm: {{ .keyAlgorithm | default "rsa" }}
+  keyEncoding: {{ .keyEncoding | default "pkcs1" }}
+  keySize: {{ .keySize | default "2048" }}
+{{- if .organization }}
```
mhrabovcin: Please add commented out example values to `values.yaml` file so they are discoverable and self-documenting:

This is example of existing `clusterissuers` property
```
clusterissuers: []
#  - name: kubernetes-ca
#    spec:
#      ca:
#        secretName: kubernetes-intermediate-ca
#  - name: my-ca
#    spec:
#      ca:
#        secretName: my-certificate-secret
```

staging/cert-manager-setup/ci/general-test-values.yaml
sebbrandt87: I would like to stick to the CRDs key name 😬 
But we can add a comment ;)

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -10,9 +10,12 @@ metadata:
     "helm.sh/hook-weight": "-3"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  isCA: true
-  commonName: cert-manager
-  duration: {{ .duration | default "87600h" | quote }}
+{{- if .acme }}
+{{ . | toYaml | indent 2 }}
+{{- end }}
```
joejulian: consider indenting for readability and using nindent
```suggestion
{{- if .acme }}
  {{ . | toYaml | nindent 2 }}
{{- end }}
```

sebbrandt87:
> My only concern is that we verify we've manually tested this prior to merging, otherwise LGTM 👍

a rendered certificate with all new values available looks like 

```
spec:
  acme:
    config:
    - domains:
      - foo.example.com
      - bar.example.com
      http01:
        ingressClass: nginx
  commonName: cert-manager
  dnsNames:
  - example.com
  - www.example.com
  duration: 87600h
  ipAddresses:
  - 127.0.0.1
  issuerRef:
    kind: Issuer
    name: my-root-issuer
    usages:
    - digital signature
    - key encipherment
  keyAlgorithm: rsa
  keyEncoding: pkcs8
  keySize: 4096
  name: my-certificate
  organization:
  - D2iQ, Inc.
  - Mesosphere, Inc.
  renewBefore: 360h
  secretName: my-certificate-secret
```

so exactly how it should be @shaneutt

staging/cert-manager-setup/templates/certificates.yaml
sebbrandt87: this would add a new line, which is not what we want.
`n` for `newline`

shaneutt:
@mhrabovcin you're review is still blocking, are we all set or is there more you're looking for here?

sebbrandt87:
D2IQ-64608 #resolved

PR #455, opened 2020-02-24
fix: probe values under high load

hectorj2f:
This is done to fix constant restart when dex-k8s-auth is under load of >= 50 konvoyclusters.

related to: https://jira.d2iq.com/browse/D2IQ-56464


nLight:
https://jira.d2iq.com/browse/D2IQ-56464

shaneutt:
@hectorj2f let @mesosphere/sig-ksphere-catalog know if there's any assistance you need to move this one forward!

hectorj2f:
I found some lint issues, I need a new approval here :)

PR #457, opened 2020-02-24
Allow nvidia driver to install with ami kernel

joejulian:
The nvidia driver installer checks for the kernel-headers rpm in order to install the driver. Since it's running the kernel of the node on which this container is running, in order to build and install the kernel module it needs access to the rpm for the default kernel for the centos 7 ami.

This PR makes available the upstream yum repos for all the legacy centos 7 distro kernels so that the driver installer can find the matching rpm without the user needing to upgrade the node's kernel.

### DIFF
--- a/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
+++ b/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
@@ -264,7 +264,7 @@ data:
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: {{ template "nvidia-driver.fullname" . }}-centos7-vault-repo
+  name: {{ template "nvidia-driver.fullname" . }}-wrapper-sh
   namespace: kube-system
 data:
   wrapper.sh: |
--- a/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
+++ b/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
@@ -270,5 +270,5 @@ data:
   wrapper.sh: |
     #!/bin/bash
     mkdir -p /lib/modules/$(uname -r)/proc
-    cp /proc/version /lib/modules/proc/version
+    cp /proc/version /lib/modules/$(uname -r)/proc/version
     /usr/local/bin/nvidia-driver init -a
--- a/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
+++ b/staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
@@ -1,0 +1,274 @@
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ template "nvidia-driver.fullname" . }}-centos7-vault-repo
+  namespace: kube-system
+data:
+  CentOS-Vault.repo: |
+    # CentOS Vault contains rpms from older releases in the CentOS-7
+    # tree.
+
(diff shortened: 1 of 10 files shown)


staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
```diff
@@ -0,0 +1,274 @@
...
+apiVersion: v1
+metadata:
+  name: {{ template "nvidia-driver.fullname" . }}-centos7-vault-repo
+  namespace: kube-system
+data:
+  wrapper.sh: |
+    #!/bin/bash
+    mkdir -p /lib/modules/$(uname -r)/proc
+    cp /proc/version /lib/modules/proc/version
+    /usr/local/bin/nvidia-driver init -a
```
GoelDeepak: Just curious, is this command idempotent? Thinking of an upgrade scenario where the driver might already be in use

alejandroEsc: Upgrade in use should be, what is making you doubt that?

joejulian: The nvidia-driver script handles that. The other commands just succeed, regardless.

jieyu:
Could you please add some descriptions about this PR?

joejulian:
Tested on a vanilla konvoy cluster using the default aws ami on a p2 instance.

staging/nvidia/charts/nvidia-driver/templates/daemonset.yaml
```diff
@@ -43,10 +44,28 @@ spec:
...
       volumes:
       - name: update-driver
         hostPath:
           path: /etc/kernel/postinst.d
       - name: driver-path
         hostPath:
           path: /run/nvidia
+      - name: modules
+        hostPath:
+          path: /lib/modules
```
jieyu: is there a way to mount this as RO? I do see `rm -rf /lib/modules/${KERNEL_VERSION}` in the driver init script, but I didn't check deeper. Allowing the container to delete files on the host under `/lib/modules` sounds very dangerous to me.

joejulian: More dangerous than installing a kernel module?

staging/nvidia/charts/nvidia-driver/templates/configmap.yaml
```diff
@@ -0,0 +1,274 @@
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ template "nvidia-driver.fullname" . }}-centos7-vault-repo
+  namespace: kube-system
+data:
```
dkoshkin: Will this work when air-gapped?

joejulian: No, but neither will the driver install as it is now. I have an idea, but I'd like to do it as a next step.

staging/nvidia/charts/nvidia-driver/templates/daemonset.yaml
jieyu: I will check the init script more carefully today. I just don’t want the files in modules dir to be accidentally deleted as I do see some rm in the script.

Kernel module is trusted in this case. But the script might assume modules dir in the container is decoupled from the host

PR #460, opened 2020-02-25
kommander: Fixes to the clusters summary dashboard

gracedo:
Fixed the cluster summary dashboard which wasn't getting automatically imported correctly (reworked the query to use the same datasource instead of trying to use two on the same dashboard which seems to be causing issues) and also added `[Global]` to the dashboard name and changed the uid so it's different from the konvoy-equivalent dashboard's name/uid (just to avoid potential name and uid collisions if someone were to ever deploy a grafana that pulls in both sets of dashboards).

I'll be opening another PR later today adding more central dashboards (porting over the base k8s dashboards from konvoy), wanted to keep them separate since this one's a bug fix.

### DIFF
--- a/stable/kommander/templates/grafana/dashboards/k8s-resources-clusters.yaml
+++ b/stable/kommander/templates/grafana/dashboards/k8s-resources-clusters.yaml
@@ -2669,7 +2669,7 @@ data:
       },
       "timezone": "",
       "title": "Kubernetes / Compute Resources / Clusters",
-      "uid": "efa86fd1d0c121a26444b636a3f509a8",
+      "uid": "f0U1EswWk",
       "version": 1
     }
 {{- end }}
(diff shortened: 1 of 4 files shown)


stable/kommander/templates/grafana/dashboards/k8s-resources-clusters.yaml
```diff
@@ -2677,8 +2668,8 @@ data:
         ]
       },
       "timezone": "",
-      "title": "Kubernetes / Compute Resources / Clusters",
-      "uid": "efa86fd1d0c121a26444b636a3f509a8",
+      "title": "Kubernetes / Compute Resources / Cluster [Global]",
+      "uid": "f0U1EswWk",
```
joejulian: Minor nit, but I'll not block on it. This doesn't seem like it's a `uid`. Doesn't even seem to be a `uuid` any more. Should this key be more descriptive?

gracedo: It's a grafana key so we can't change it, but this was the value automatically generated by grafana when I created the dashboard in the UI so I wouldn't worry too much about it!

PR #470, opened 2020-02-27
Readd Dispatchfile

cprovencher:
This reverts commit 44fef52.
and adds a simple fix at this line: https://github.com/mesosphere/charts/compare/fix_dispatchfile?expand=1#diff-37eda1430799e62f92596aa54fae0596R49

This initially was reverted because the Dispatch build was failing and causing TeamCity CI to fail as well. The problem was a certain entrypoint.sh that was being downloaded. That file was in a private repo and required a token in its URL. That token was not a security hazard; it only exposed that single revision of that single file. However, little did I know, that token expires after a few days, which was causing the failure. We are also trying to upstream that file in the kubernetes-sigs/kind repo, so there is no harm in making it public. I added it as a gist for the d2iq-dispatch user so we can download it from there. I also added the -f switch to the "curl" commands. It would have made debugging this issue much easier in the first place.

### DIFF
--- a/Dispatchfile
+++ b/Dispatchfile
@@ -1,0 +1,71 @@
+#!starlark
+# vi:syntax=python
+#
+
+git = "src-git"
+
+def secretVar(name, key):
+    return k8s.corev1.EnvVarSource(secretKeyRef=k8s.corev1.SecretKeySelector(localObjectReference=k8s.corev1.LocalObjectReference(name=name), key=key))
+
+gitResource(git,
+    url="$(context.git.url)",
+    revision="$(context.git.commit)")
+
+dindTask("publish", inputs=[git], steps=[
+  v1.Container(
+    name = "publish-charts",
+    workingDir =  "/workspace/src-git",
+    args = [
+      "/bin/bash", "-c",
+      """
+      apt-get update
+      apt-get install -qq git build-essential
+      git fetch origin master --unshallow
+      export GIT_REMOTE_URL=https://mesosphere:${GITHUB_USER_TOKEN}@github.com/mesosphere/charts.git
+      make publish
+      """
+    ],
+    env=[k8s.corev1.EnvVar(name="GITHUB_USER_TOKEN", valueFrom=secretVar("scmtoken", "password"))]
+  )
+])
+
+dindTask("lint", inputs=[git], steps=[
+  v1.Container(
+    name = "lint-charts",
+    workingDir =  "/workspace/src-git",
+    args = [
+      "/bin/bash", "-c",
+      """
+      apt-get update
+      apt-get install -qq git build-essential
+      git fetch origin master --unshallow
+      make lint
+      """
+    ]
+  )
+])
+
+dindTask("test", inputs=[git], deps=["lint"], steps=[
+  v1.Container(
+    name = "test-charts",
+    workingDir =  "/workspace/src-git",
+    args = [
+      "/bin/bash", "-c",
+      """
+      apt-get update
+      apt-get install -qq git build-essential
+      git fetch origin master --unshallow
+      make test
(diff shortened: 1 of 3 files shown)


test/e2e-kind.sh
```diff
@@ -41,12 +42,22 @@ docker_exec() {
...
+    cat << EOF > tmp_dockerfile
+FROM kindest/node:$K8S_VERSION
+ADD ./entrypoint.sh /usr/local/bin/entrypoint
+EOF
+
+    docker build -t tmp-dispatch-kind:latest -f tmp_dockerfile "${tmp}"
+
     "${tmp}/kind" create cluster --name "$CLUSTER_NAME" \
-        --config test/kind-config.yaml --image "kindest/node:$K8S_VERSION" \
+        --config test/kind-config.yaml --image "tmp-dispatch-kind:latest" \
```
orsenthil: What is this special image `tmp-dispatch-kind` ?

cprovencher: @orsenthil it got built two lines above: https://github.com/mesosphere/charts/pull/470/files#diff-37eda1430799e62f92596aa54fae0596R57

cprovencher: As soon as our version of entrypoint.sh is contributed to github.com/kubernetes-sigs/kind, we can replace the "gist" url by a link to entrypoint.sh in that repo.

orsenthil: Hi @cprovencher : could that be added as a comment with the link to the PR?

cprovencher: Done

cprovencher:
> I agree with @orsenthil. Once you comment the gist and link to the PR and/or issue it lgtm.

Done

faiq:
It all looks good to me! Great work Charles :)

Dispatchfile
```diff
@@ -0,0 +1,71 @@
...
+    args = [
+      "/bin/bash", "-c",
+      """
+      apt-get update
+      apt-get install -qq git build-essential
+      git fetch origin master --unshallow
+      make test
+      """
+    ],
+    resources = k8s.corev1.ResourceRequirements(
```
makkes: Why aren't limits (and requests) set on all the tasks?

Dispatchfile
```diff
@@ -0,0 +1,71 @@
...
+    revision="$(context.git.commit)")
+
+dindTask("publish", inputs=[git], steps=[
+  v1.Container(
+    name = "publish-charts",
+    workingDir =  "/workspace/src-git",
+    args = [
+      "/bin/bash", "-c",
+      """
+      apt-get update
```
makkes: Is there a way to produce or rather use an image that already has the required packages installed? This would dramatically speed up running these tasks.

Dispatchfile
cprovencher: @makkes Limits in Dispatch are pretty low by default. Since the "test" task creates a kind cluster, it requires more resources to ensure it succeeds.

Dispatchfile
cprovencher: @makkes Reasoning behind why I chose to do it this way:
Every time the K8S_VERSION variable gets updated, the contributor would need to also make sure to build a new image. I don't find this reliable as people are bound to forget!
That Dockerfile would also only be 2 lines. It's only taking the kindest/node image and adding a file to it. So between just pulling a pre-built image vs pulling the kindest/node image and adding a file to it, I don't think the difference in speed is significant.

makkes: @cprovencher The difference in speed will be caused by not having to `apt-get update` and `apt get install [...]` every time we publish, test or lint the code (i.e. on every PR). Please consider having a Docker image for that.

PR #483, opened 2020-02-28
[PROPOSAL] maintain multiple chart versions for kommander chart

juliangieseke:
This PR is a proposal to keep the possibility open to release a fix to the chart that is used in "Kommander 1.0" GA build (e.g. Kommander 1.0.1) while we're already working on "Kommander 1.1". 

It is a proposal which I'm happy to discuss. so please dont take this PR as set in stone.

(its probably best to review checked out on your machine, since its basilcy only moving & copying files)

addons PR that would consume this PR: https://github.com/mesosphere/kubeaddons-kommander/pull/34
Lets keep conversation in this PR to keep things focused.

---

From the README:

It might be a bit overly cautious, but following "better safe than sorry" approach, we're maintaining individual chart versions per supported Kommander addon version.

In case we need to release "Kommander 1.0.1" while already working on "Kommander 1.1", we need to be able to release a new chart version for "Kommander 1.0.1".

Tying chart version to appVersion and using revisions should help us being flexible while reducing overall version confusion.

We might find a better solution in future…


juliangieseke:
yeah, its probably a better idea anyway. will do as soon as time allows

PR #484, opened 2020-02-28
chore: fix generic-ingress chart

jieyu:
The YAML is not valid due to `---` being aligned to the wrong place.
This patch fixed it.

Tested with the following:
```
helm template -f new-values.yaml . | yamllint -
```

`new-values.yaml` the same we use in KCL
https://github.com/mesosphere/kommander-cluster-lifecycle/blob/mh/kubefed/ops-portal-ingress/chart/kommander-cluster-lifecycle/templates/federatedaddon_ops_portal_ingress.yaml#L27-L75

### DIFF
--- a/staging/generic-ingress/templates/ingresses.yaml
+++ b/staging/generic-ingress/templates/ingresses.yaml
@@ -18,4 +18,4 @@ spec:
         {{- with $ingress.paths -}}
           {{ toYaml . | nindent 10 }}
         {{- end }}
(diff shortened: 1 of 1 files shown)


jieyu:
@mesosphere/sig-ksphere-catalog we hope to skip the chart version bump for the release today (sorry about that). This would save us change to KCL and re-run everything (could save us 1 hour).

PR #488, opened 2020-03-05
fix: don't install packages as they're already installed

makkes:
The base image already has those packages installed, anyway.


makkes:
Please don't merge, yet, as long as @cprovencher hasn't approved. Talked to him out-of-band.

PR #494, opened 2020-03-07
velero: cleanup velero minio

jieyu:
Instead of installing minio operator and install the minio instance in
the init container, this refactor uses the minio upstream chart as a
subchart of this chart to install minio instance.

The bundled minio chart has been bumpted to stable/minio 5.0.12.

Tested manually by first delete velero addon in the konvoy cluster, and do the following (also requires another change to traefik addon to add the `velero-minio` entrypoint, will follow up soon).

Verified using `velero backup get` after the upgrade that the backup is still there and can be downloaded. Also verified the idempotency by upgrade this chart multiple times.
```
helm upgrade velero-kubeaddons --install . -f myvalues.yaml --namespace velero
```

myvalues.yaml
```yaml
configuration:
  provider: "aws"
  backupStorageLocation:
    name: "aws"
    bucket: "velero"
    config:
      region: "fallback"     # enables non-production fallback minio backend
      s3ForcePathStyle: true # allows usage of fallback backend
      s3Url: http://minio.velero.svc:9000
  volumeSnapshotLocation:
    name: "aws"
    config:
      region: "fallback"
credentials:
  secretContents:
    cloud: "placeholder"
schedules:
  default:
    schedule: "0 0 * * *"
metrics:
  enabled: true
  service:
    labels:
      servicemonitor.kubeaddons.mesosphere.io/path: "metrics"

initContainers:
- name: initialize-velero
  image: jieyu/kubeaddons-addon-initializer:latest
  args: ["velero"]
  env:
  - name: "VELERO_MINIO_FALLBACK_SECRET_NAME"
    value: "velero-kubeaddons"

minioBackend: true
minio:
  mode: distributed
  defaultBucket:
    enabled: true
    name: velero
  bucketRoot: "/data"
  existingSecret: minio-creds-secret
  livenessProbe:
    initialDelaySeconds: 120
    periodSeconds: 20
  resources:
    requests:
      memory: 256Mi
      cpu: 250m
    limits:
      memory: 512Mi
      cpu: 750m
  persistence:
    volumeTemplatePrefix: data
  statefulSetNameOverride: minio
  ingress:
    enabled: true
    hosts:
    - ""
    annotations:
      traefik.ingress.kubernetes.io/frontend-entry-points: velero-minio
```


jieyu:
hum, what does this mean?
`Error: Dependency minio has an invalid version/constraint format: improper constraint: `

staging/velero/values.yaml
```diff
@@ -192,11 +192,6 @@ configMaps: {}
 # Deploy a default Minio cluster to use as the backend for Velero.
 minioBackend: false
```
hectorj2f: Should not it be true?

jieyu:
@hectorj2f because we carried two extra patches. I am in the process of upstream those. In fact, one is already merged, the other one might be controversial. The patches are in the patches/ dir of minio

hectorj2f:
Okay we should document or add a Jira ticket to the files if possible

jieyu:
I did document in PATCHES.md in the nested minio chart

hectorj2f:
Oh, good

staging/velero/values.yaml
shaneutt: When we first put this chart together we didn't want to provide minio as enabled by default because it's intra-cluster and has worse characteristics for failure cases than say, using AWS S3. The addons that wrap this chart for Konvoy deployments however _do_ enable this by default.

shaneutt:
We definitely need to put this through some tests since it's a big change, we should probably take several passes over it in soak, but :heavy_check_mark: for now.

PR #496, opened 2020-03-09
cluster-autoscaler: initial commit

mhrabovcin:
JIRA: https://jira.d2iq.com/browse/D2IQ-65258

### DIFF
--- a/stable/cluster-autoscaler/Chart.yaml
+++ b/stable/cluster-autoscaler/Chart.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+description: Scales worker nodes within autoscaling groups.
+icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
+name: cluster-autoscaler
+version: 7.1.0
+appVersion: 1.17.1
+home: https://github.com/kubernetes/autoscaler
+sources:
+- https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
+- https://github.com/spotinst/kubernetes-autoscaler/tree/master/cluster-autoscaler
+maintainers:
+- name: mhrabovcin
+  email: mhrabovcin.c@d2iq.com
+engine: gotpl
--- a/stable/cluster-autoscaler/templates/clusterrole.yaml
+++ b/stable/cluster-autoscaler/templates/clusterrole.yaml
@@ -1,0 +1,146 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  labels:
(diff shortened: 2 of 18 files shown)


stable/cluster-autoscaler/templates/clusterrole.yaml
```diff
@@ -0,0 +1,146 @@
...
+    resources:
+      - events
+      - endpoints
+    verbs:
+      - create
+      - patch
+  - apiGroups:
+    - ""
+    resources:
+    - pods/eviction
```
hectorj2f: we are gonna need to perform actions on the konvoycluster objects

stable/cluster-autoscaler/templates/deployment.yaml
```diff
@@ -0,0 +1,213 @@
...
+    {{- if .Values.affinity }}
+      affinity:
+{{ toYaml .Values.affinity | indent 8 }}
+    {{- end }}
+    {{- if .Values.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 8 }}
+    {{- end }}
+      serviceAccountName: {{ template "serviceaccount.name" . }}
+      tolerations:
```
hectorj2f: we want it to be deployed on the control-plane by default

stable/cluster-autoscaler/Chart.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: Scales worker nodes within autoscaling groups.
+icon: https://github.com/kubernetes/kubernetes/blob/master/logo/logo.png
+name: cluster-autoscaler
+version: 7.1.0
+appVersion: 1.17.1
+home: https://github.com/kubernetes/autoscaler
+sources:
+- https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
+- https://github.com/spotinst/kubernetes-autoscaler/tree/master/cluster-autoscaler
+maintainers:
```
shaneutt: Are these upstream maintainers?

If so, this is a `stable/` chart so we need to add our own maintainers so as to not confuse our end-users.

mhrabovcin: Yes, these are upstream maintainers @shaneutt . Should I replace completely maintainers section or just extend it with our names?

hectorj2f: @shaneutt we are gonna apply changes to this vanilla chart, so I'd say we could add them as part of a different PR, imho.

shaneutt: > Yes, these are upstream maintainers @shaneutt . Should I replace completely maintainers section or just extend it with our names?

For now, replace it with our names. We should not claim those people are responsible for our fork.

PR #499, opened 2020-03-11
Update CODEOWNERS

shaneutt:
This PR adds @juliangieseke as a codeowner over the Kommander chart as per the [catalog codeowners guide](https://github.com/mesosphere/ksphere-platform/blob/master/docs/sig-ksphere-catalog/codeowners-guide.md).


shaneutt:
we're using github teams here instead now, should be all set :+1:

PR #505, opened 2020-03-18
fix: Dispatch publish tasks

cprovencher:
Publish task requires name and email to be set. This was working on teamcity because that environment has username and hostname set, which git uses by default.


Dispatchfile
```diff
@@ -20,6 +20,8 @@ dindTask("publish", inputs=[git], steps=[
       """
       git fetch origin master --unshallow
       export GIT_REMOTE_URL=https://mesosphere:${GITHUB_USER_TOKEN}@github.com/mesosphere/charts.git
+      git config --global user.name 'Dispatch CI'
+      git config --global user.email '<>'
```
makkes: What about setting at least a valid generic email address here so people reading the git history have a way to contact someone who could help with certain commits. Does sth. like `devprod@d2iq.com` exist?

cprovencher: @makkes The current publish job for TeamCity has a placeholder/invalid email, so I replicated that for Dispatch. I don't know who looks at the commit history for the gh-pages branch, but if they had any questions I would not direct them to devprod. The commits are done by Dispatch, but the content of the commits is unrelated to Dispatch.

PR #506, opened 2020-03-19
chore: Delete kommander-thanos and kommander-karma charts

samvantran:
Commit 721d64e turned `kommander-thanos` and `kommander-karma` charts into
subcharts under stable/kommander so these charts are no longer needed as
they are now duplicated.


juliangieseke:
You would have been able to keep history when doing this in the same commit, now its new files that are added (and old ones being deleted) instead.

PR #507, opened 2020-03-19
feat(kube-oidc-proxy): added caSystemDefault and caCertPemHostPath

sebbrandt87:
Allow to use system default CA in the image or a bundle from the hosting OS.

Tested with the following combinations by checking `helm template` output:
```yaml
oidc:
  caSystemDefault: true
```
```yaml
oidc:
  caCertPemHostPath: /etc/ssl/certs/ca-bundle.crt
```
```yaml
oidc:
  caSystemDefault: true # This will take precedence.
  caCertPemHostPath: /etc/ssl/certs/ca-bundle.crt
```

[D2IQ-65234]: https://jira.d2iq.com/browse/D2IQ-65234

### DIFF
--- a/staging/kube-oidc-proxy/templates/deployment.yaml
+++ b/staging/kube-oidc-proxy/templates/deployment.yaml
@@ -47,10 +47,10 @@ spec:
           - "--oidc-client-id=$(OIDC_CLIENT_ID)"
           - "--oidc-issuer-url=$(OIDC_ISSUER_URL)"
           - "--oidc-username-claim=$(OIDC_USERNAME_CLAIM)"
-          {{- if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
+          {{- if .Values.oidc.caSystemDefault }}
+          # Not setting `--oidc-ca-file` to use CA bundle from the image
+          {{- else if or .Values.oidc.caPEM .Values.oidc.caSecretName .Values.oidc.caCertPemHostPath }}
           - "--oidc-ca-file=/etc/oidc/oidc-ca.pem"
-          {{- else if and .Values.oidc.caKubernetesDefault }}
-          # disabled --oidc-ca-file to use container image available root CAs
           {{- else }}
           - "--oidc-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
           {{ end }}
@@ -143,20 +143,24 @@ spec:
               key: api-audiences
         {{ end }}
         volumeMounts:
-          {{ if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
+          {{- if .Values.oidc.caSystemDefault }}
+          # No need to mount volumes. Use the CA bundle in the image.
+          {{- else if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
           - name: kube-oidc-proxy-config
             mountPath: /etc/oidc
             readOnly: true
-          {{ else if .Values.oidc.caCertPemHostPath }}
+          {{- else if .Values.oidc.caCertPemHostPath }}
           - name: ca-cert-pem
-            mountPath: {{ .Values.oidc.caCertPemMountPath }}
+            mountPath: /etc/oidc/oidc-ca.pem
             readOnly: true
           {{ end }}
           - name: kube-oidc-proxy-tls
             mountPath: /etc/oidc/tls
             readOnly: true
       volumes:
-        {{ if .Values.oidc.caPEM }}
+        {{- if .Values.oidc.caSystemDefault }}
+        # No need to mount volumes. Use the CA bundle in the image.
+        {{- else if .Values.oidc.caPEM }}
         - name: kube-oidc-proxy-config
           secret:
             secretName: {{ include "kube-oidc-proxy.fullname" . }}-config
--- a/staging/kube-oidc-proxy/templates/deployment.yaml
+++ b/staging/kube-oidc-proxy/templates/deployment.yaml
@@ -49,6 +49,8 @@ spec:
           - "--oidc-username-claim=$(OIDC_USERNAME_CLAIM)"
           {{- if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
           - "--oidc-ca-file=/etc/oidc/oidc-ca.pem"
+          {{- else if and .Values.oidc.caKubernetesDefault }}
+          # disabled --oidc-ca-file to use container image available root CAs
           {{- else }}
           - "--oidc-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
           {{ end }}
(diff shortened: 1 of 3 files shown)


staging/kube-oidc-proxy/templates/deployment.yaml
```diff
@@ -166,6 +172,11 @@ spec:
             items:
             - key: ca.crt
               path: oidc-ca.pem
+        {{- else if .Values.oidc.caCertPemHostPath }}
+        - name: ca-cert-pem
+          hostPath:
```
jieyu: I don't think we should rely on host path for this change. The docker image should already have openssl root ca certs installed (i checked `quay.io/jetstack/kube-oidc-proxy:v0.2.0`).

staging/kube-oidc-proxy/templates/deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
           - "--oidc-username-claim=$(OIDC_USERNAME_CLAIM)"
           {{- if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
           - "--oidc-ca-file=/etc/oidc/oidc-ca.pem"
+          {{- else if and .Values.oidc.caCertPemHostPath (not .Values.oidc.caPEM) (not .Values.oidc.caSecretName) }}
+          # disabled --oidc-ca-file to use hostVolume mount
```
jieyu: See my comments below about hostpath. This would be my recommendation:

1. We introduce another knob in `values.yaml` like the following
```yaml
oidc:
  caPEM:
  caSecretName:
  caKubernetesDefault: true
```

At most one of them should be set (or set to true). If all of them are not set (and set to false), `--oidc-ca-file` flag won't be set in the deployment.

staging/kube-oidc-proxy/templates/deployment.yaml
sebbrandt87: yes it has, but you need to keep the local ca certs updated in the image then also, which means you need to update the image somewhen if you see that your pod starts to fail and the CA is the reason.
It is a beneficial thing to mount especially something from the host path as read-only, like the CA certs. As the system already should take care of regular updating the CA certs bundle.

jieyu: Sorry @sebbrandt87 but I still insist we should not use hostpath at all. The host OS might be different, and the location for those root CA certs bundle might be different. See https://golang.org/src/crypto/x509/root_unix.go#16.

The is the benefit of containers that OS dependencies are decoupled from the underlying host OS. The certs update can be done by running `update-ca-certificates` or equivalent command in the container before running the actual command. Or the container just need to be updated regularly to pick up the changes to the root CA certs.

sebbrandt87: the root CAs PEM location can be different yes, I thought about that earlier already and I know that ;) 
But as we use here the upstream container image from kube-oidc-proxy and we have a fixed version right now, which is using alpine as distribution we know where the CAs cert is placed and we can expect that it will not change within that tag and maybe even not in the next tags. I will change the PR slightly just in case that you would like to change the mountpoint in the running container ;)

> The certs update can be done by running update-ca-certificates or equivalent command in the container before running the actual command. 
This is an anti-pattern to having consistent container images that you want to run and also adds "some" computing / operation time.

> Or the container just need to be updated regularly to pick up the changes to the root CA certs.
Yes, that is true but do we provide this for the used container image and would we tag the current container image then with the same tag? I would say no.

jieyu: @sebbrandt87 the problem is not the container image (alpine), the problem is the underlying OS (currently, we support debian, ubunto and centos/rhel), and they have different locations for saving root CA bundles. if we go with hostpath route, we'd probably have to have different configurations for the addon for different underlying OS distro, which is unacceptable to me as this is very hard to maintain.

jieyu: Also, the upstream build pipeline already update CA bundle on every release https://github.com/jetstack/kube-oidc-proxy/blob/master/Dockerfile#L5

So we just need to constantly picking up their new releases during base tech upgrade.

victhar: @sebbrandt87 I think we should find a trade-off and to me the "update-ca-certificates" seems to be the most harmless option, although I understand its disadvantages, but there is no good solution it seems. Let's not get stuck on this and move forward as quickly as possible. 
also Jie just pointed out in slack - https://github.com/jetstack/kube-oidc-proxy/blob/master/Dockerfile#L5

jieyu: OK, if SRE really prefers this way, we could add this option.

staging/kube-oidc-proxy/templates/deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
           - "--oidc-username-claim=$(OIDC_USERNAME_CLAIM)"
           {{- if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
           - "--oidc-ca-file=/etc/oidc/oidc-ca.pem"
+          {{- else if and .Values.oidc.caCertPemHostPath (not .Values.oidc.caPEM) (not .Values.oidc.caSecretName) }}
```
jieyu: The `not` sections are not needed as this is an `else if`

jieyu: Also, I think we should also have a way to *not* specify this flag if the user simply want to use the ca bundle in the image. i.e., even if `caCertPemHostPath` is not set, we still need a way allowing users to not setting this flag.

Thus, we still need `caKubernetesDefault` flag i suggested below.

staging/kube-oidc-proxy/templates/deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
           - "--oidc-username-claim=$(OIDC_USERNAME_CLAIM)"
           {{- if or .Values.oidc.caPEM .Values.oidc.caSecretName }}
           - "--oidc-ca-file=/etc/oidc/oidc-ca.pem"
+          {{- else if and .Values.oidc.caKubernetesDefault }}
```
jieyu: Do you also need to check for `oidc.caCertPemHostPath` here and set `--oidc-ca-file` accordingly? I think we should define the precedence order:
1. caPEM or caSecretName
2. caCertPemHostPath
3. caKubernetesDefault
4. system default in the image

lower the number means higher the priority. We should document this in the values file.

Also, no need for `and` here.

staging/kube-oidc-proxy/values.yaml
```diff
@@ -31,14 +31,32 @@ oidc:
...
+  # to be able to use container image available root CA's
+  caKubernetesDefault: false
+  # ---
+  # caCertPemHostPath, which points to a cert.pem including CA's on your hostMachine
+  # should be used if you issue an certificate from an ACME provider like letsencrypt
+  # it will be mounted read-only and only one file e.g. /etc/ssl/cert.pem.
+  # you can define the target within the container through caCertPemMountPath.
+  # This should be used in combination with caKubernetesDefault: true
+  caCertPemHostPath:
+  caCertPemMountPath: /etc/ssl/cert.pem
```
jieyu: No need for this. I'd suggest we remove this configuration option and just hard code it in the manifest.

PR #509, opened 2020-03-20
kube-oidc-proxy: cleanup some certs logic

jieyu:
I happened to notice the issue when I was working on another
improvement. The chart is not really conditional on `tls.enabled` as it
always expect the server certs to be present in the deployment. Thus, I
remove all the conditional logics.


staging/kube-oidc-proxy/templates/certificate.yaml
```diff
@@ -1,11 +1,10 @@
-{{- if .Values.tls.enabled }}
 apiVersion: certmanager.k8s.io/v1alpha1
 kind: Certificate
 metadata:
   name: kube-oidc-proxy
   namespace: {{ .Release.Namespace }}
 spec:
-  secretName: kube-oidc-proxy-kubeaddons-server-tls
+  secretName: {{ include "kube-oidc-proxy.fullname" . }}-server-tls
```
mhrabovcin: is this backward compatible for existing installations?

staging/kube-oidc-proxy/templates/deployment.yaml
```diff
@@ -1,6 +1,4 @@
 {{ $fullname := include "kube-oidc-proxy.fullname" . }}
-{{ $defaultTlsSecretName := printf "%s-tls" $fullname }}
```
mhrabovcin: 👍

staging/kube-oidc-proxy/templates/certificate.yaml
jieyu: @mhrabovcin yes, i tested with an old verison, and perform an `helm upgrade kube-oidc-proxy .` and observed no change.

jieyu:
@mhrabovcin i rebased and need re-approval.

staging/kube-oidc-proxy/templates/certificate.yaml
makkes: We need tests to automate backwards compatibility. *sigh*

PR #512, opened 2020-03-23
chore: move values from addon to chart

juliangieseke:
first step to address kommander chart not being able to install w/o addon. I was _not_ able to install the chart on its own, but that is no change to before. we need to fix that going forward

This PR adds everything from addon to chart, companioning PR removing these things from addon:


hectorj2f:
what do you gain with this change ?

juliangieseke:
goal is to be able to install the chart w/o its addon.

PR #513, opened 2020-03-23
[D2IQ-65734] add konvoy-async-auth chart

shaneutt:
As per [D2IQ-65734](https://jira.d2iq.com/browse/D2IQ-65734)


shaneutt:
Blocked on https://github.com/mesosphere/konvoy-async-auth/pull/2 and [D2IQ-65269](https://jira.d2iq.com/browse/D2IQ-65269)

jr0d:
@shaneutt should we move this to staging?

shaneutt:
> @shaneutt should we move this to staging?

Originally we had intended staging for charts that we wanted to eventually send upstream :thinking: but I suppose it's more become a place where we put new charts. We don't have any real process in place right now to graduate charts one way or the other so I'm tempted to just keep it here but I'm open to suggestions. cc @joejulian

joejulian:
https://github.com/mesosphere/charts/blob/master/README.md#repository-structure

shaneutt:
we've changed course and decided we're not going to make a separate helm chart for this.

PR #515, opened 2020-03-24
[kommander] Revert "chore: move values from addon to chart"

juliangieseke:
Reverts mesosphere/charts#512
---
after talking about it in sync, seems like most of the stuff was actually intentionally in the addon and being able to use the chart is not a priority right now. I failed to get it working in a couple hours, so I'm stopping this effort for now. we should revisit it later tho.


alejandroEsc:
@juliangieseke could you please add the appropriate label, is this a WIP or is it ready for review?

juliangieseke:
@alejandroEsc its ready, but had to update again due to chart version conflict. 
Is there any documentation about label usage (and also assignee usage) for these repos? seems like there is some common understanding about what to do, but wasnt able to find it written down anywhere

PR #517, opened 2020-03-24
[prometheus-operator] fix: hook should not call static service name

alejandroEsc:
hook should not point to static service name, it's now dynamic, and should fail silently to allow prometheus to deploy. This latter point is quite specific as for the hook to actually succeed it requires a specific configuration of grafana allowing for automatic creation of a user.

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.7.5
+version: 8.8.0
 appVersion: 0.35.0
 tillerVersion: ">=2.12.0"
 home: https://github.com/coreos/prometheus-operator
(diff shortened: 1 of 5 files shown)


staging/prometheus-operator/Chart.yaml
```diff
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.7.4
+version: 8.8.4
```
samvantran: shouldn't this be `8.7..5` since it's a fix or would you consider this a new feature? In that case should it be `8.8.0`?

alejandroEsc: Oh, I am making a breaking change since one field, the serviceURL is no longer valid. It is being deprecated and not consumed anywhere atm.

alejandroEsc:
> 👍 nice work - would you mind also fixing up the code in https://github.com/mesosphere/charts/blob/master/stable/kommander/templates/grafana/hooks-home-dashboard.yaml ?

No problem, didnt check that code at all, so yeah worth a look.

alejandroEsc:
@samvantran @gracedo I will make the changes to the ingress record in another PR since its really kommander specific and i want to run a few tests before i make any additional changes. Note, that this will make ` serviceURL: http://kommander-kubeaddons-grafana.kommander` deprecated since this will be done at deploy time and not be a function of user input.

PR #522, opened 2020-03-25
fix: grafana dashboard job only if defaultDashboardsEnabled

armandgrillet:
This affects the prometheus-operator by not deploying the job grafana-default-dashboard if we set defaultDashboardsEnabled to false.

Related ticket: https://jira.d2iq.com/browse/D2IQ-66014

This was tested by deploying the updated chart combined with a `values.yaml` specifying:
```
grafana:
            defaultDashboardsEnabled: false
```


staging/prometheus-operator/templates/mesosphere-hooks/grafana-default-dashboard-hook.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.grafana.defaultDashboardsEnabled }}
```
gracedo: can we also add a `.Values.grafana.enabled` check? thanks for catching this! :)

PR #523, opened 2020-03-25
kommander: Federate kubecost addon to managed clusters

gracedo:
Adds kubecost federatedaddon that is federated out to managed clusters. This is defaulted to off until all of the other parts are in place to centralize costs up to kommander. There are also some TODOs noted throughout for future phases of work.

Along with the kubecost components, a separate Prometheus instance scraping only kubecost metrics is also installed.

## Testing
I tested this by hosting the updated chart on my personal GH. You can test using the kubeaddons-kommander branch `gracedo/kommander_add_kubecost_subchart_D2IQ-65173` which points to this chart. This is also blocked until a new release for `kubeaddons-extrasteps` goes out, so until then I overrode the tag by applying the following kommander config into my config.yaml:
```
...
  - configRepository: https://github.com/mesosphere/kubeaddons-kommander
    #configVersion: stable-1.16-1.0.0
    configVersion: gracedo/kommander_add_kubecost_subchart_D2IQ-65173
    addonsList:
    - name: kommander
      enabled: true
      values: |
        kubecost:
          enabled: true
          federate:
            addons: true
            addonsInitializer:
              tag: "test"
        federate:
          kubeaddonsInitializer:
            tag: "test"
...
```
I `konvoy up` using this configuration and then provisioned a konvoy cluster following [these instructions](https://github.com/mesosphere/kommander-cluster-lifecycle#providing-a-cluster-on-aws). Once the cluster was provisioned, I checked that all the kubecost components were up and running successfully in the `kubecost` namespace. The kubecost UI is also accessible at: `<CLUSTER_URL>/ops/portal/kubecost/overview.html` and the Allocation page shows data indicating that prometheus and the kubecost helm chart were all deployed correctly.
image


stable/kommander/values.yaml
```diff
@@ -18,6 +18,10 @@ kommander-karma:
   kommanderServiceAccount: kommander-kubeaddons
   kommanderKarmaConfigMap: kommander-kubeaddons-config
 
+kubecost:
+  enabled: false
+  thanosAddress: "kommander-kubeaddons-kubecost-kubecost-kubeaddons-prometheus-se.kubecost.svc.cluster.local:10901"
```
samvantran: that super long address is maxed at 63 chars... could there be a potential nameclash problem on a restart?

stable/kommander/charts/kubecost/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+appVersion: 1.0.0
```
samvantran: should this be `1.55.0`? based on [this](https://github.com/mesosphere/charts/pull/523/files#diff-5de584009d684fa30ea6c164a931f218R73)

stable/kommander/charts/kubecost/templates/ingress-roles.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if .Values.portalRBAC.enabled }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: opsportal-kommander-kubecost-admin
```
samvantran: :

stable/kommander/values.yaml
gracedo: hopefully there are no other prometheus components that start with `se` 🙃 but yeah the name length is unfortunate, this is even after I renamed this subchart from `kommander-kubecost` to just `kubecost`. I think the other option is to override this dynamically-created name entirely and set our own name which I had tried out in my testing as well.

stable/kommander/charts/kubecost/templates/ingress-roles.yaml
gracedo: can you repeat the question 😄

stable/kommander/values.yaml
alejandroEsc: We most likely can pick this up dynamically, can we check if that is possible before going this route of using a static value>

stable/kommander/requirements.yaml
```diff
@@ -15,3 +15,8 @@ dependencies:
     version: 4.5.*
     repository: https://kubernetes-charts.storage.googleapis.com
     condition: grafana.enabled
+  # Bug with `helm dep update` on local subchart: https://github.com/helm/helm/issues/3742
```
alejandroEsc: You can still add it, i think there is a working example of this in kubefed and nvidia. Please check that.

stable/kommander/charts/kubecost/templates/ingress-roles.yaml
alejandroEsc: Please let's add release name to the name of a resource when possible. This will handle the multiple release cases and avoid collisions.

gracedo:
> Why are we adding this manually to the charts directory and not pulling the chart in from source? If we need to make changes/modifications to original chart, why not put the chart in the /staging subdirectory?

Sorry it is confusing because I named the kommander subchart `kubecost` (originally wanted `kommander-kubecost` to avoid this confusion but it made the generated name too long) but we aren't forking the actual kubecost helm chart (their chart is actually called `cost-analyzer`). the upstream chart at the moment is only being referenced in the FederatedAddon (https://github.com/mesosphere/charts/pull/523/files#diff-5de584009d684fa30ea6c164a931f218R70-R73) so we're not actually pulling it in yet. We will be pulling it in in the near future, as a dep of our `kubecost`, to get it running on kommander but not in this stage.

alejandroEsc:
> https://github.com/mesosphere/charts/pull/523/files#diff-5de584009d684fa30ea6c164a931f218R70-R73

So ill have to check again and maybe i missed the point, but you are saying then that the additional changes here are to federate the addon out? Why then not have these in kcl? Where most of the federation of addons happens.

gracedo:
> > https://github.com/mesosphere/charts/pull/523/files#diff-5de584009d684fa30ea6c164a931f218R70-R73
> 
> So ill have to check again and maybe i missed the point, but you are saying then that the additional changes here are to federate the addon out? Why then not have these in kcl? Where most of the federation of addons happens.

@alejandroEsc 🤦‍♀ tbh it never occurred to me to do it in kcl, I think because we have been federating out the mtls-proxies in this way so I just copied that pattern. Is it the "better" way to do it via kcl?

gracedo:
I'm going to close this and move the federatedaddon for kubecost over to KCL

stable/kommander/charts/kubecost/templates/ingress-roles.yaml
samvantran: lol i thought i wrote 👍 ... but it was a `:` sorry too hasty

PR #538, opened 2020-04-07
chore: update kommander appversion

juliangieseke:
lets reflect appversion in chart too


juliangieseke:
Agree - kommander is a public product tho 🤔

hectorj2f:
@juliangieseke which version is specifying the user the appVersion or the chart version ?

juliangieseke:
appversion

chartversion only is the chart itself, not the app thats shipped with the chart

PR #544, opened 2020-04-10
velero: bump minio to fix flake during init

jieyu:
We mainly need the following fix to fix some flakyness during
initialization.
https://github.com/minio/minio/issues/9277
https://github.com/minio/minio/pull/9278

We also introduced an option in the chart to force delete the
minio statefulset. This is required because minio does not support
rolling update between certain versions. For example:
https://github.com/minio/minio/releases/tag/RELEASE.2020-03-25T07-03-04Z

Tested the upgrade with this values file:
```yaml
...
minioBackend: true
minio:
  mode: distributed
  defaultBucket:
    enabled: true
    name: velero
  bucketRoot: "/data"
  mountPath: "/data"
  existingSecret: minio-creds-secret
  livenessProbe:
    initialDelaySeconds: 120
    periodSeconds: 20
  resources:
    requests:
      memory: 256Mi
      cpu: 250m
    limits:
      memory: 512Mi
      cpu: 750m
  persistence:
    volumeTemplatePrefix: data
  statefulSetNameOverride: minio
  statefulSetForceCleanup: true
  ingress:
    enabled: true
    hosts:
    - ""
    annotations:
      traefik.ingress.kubernetes.io/frontend-entry-points: velero-minio
```


staging/velero/charts/minio/templates/cleanup-job.yaml
```diff
@@ -44,6 +44,12 @@ spec:
                 kubectl delete minioinstances -n {{ .Release.Namespace }} minio
                 kubectl delete statefulset -n {{ .Release.Namespace }} --ignore-not-found minio
               fi
+              # In case rolling update is not possible, which does
+              # happen in many minio releases, we'll need to delete
+              # the statefulset.
+              {{- if .Values.statefulSetForceCleanup }}
+              kubectl delete statefulset -n {{ .Release.Namespace }} --ignore-not-found {{ template "minio.statefulset.name" . }}
```
dkoshkin, line 51: What happens to data when this option is used?
Is it safe to always pass in `true` here?

jieyu: @dkoshkin data won't be lost because pvc won't be deleted, and the new statefulset will pick up the left pvc. And yes, it's always safe to pass `true` here.

PR #547, opened 2020-04-14
velero: backport of extraArgs fix: #20091

vespian:
This is a backport of https://github.com/helm/charts/pull/20091/ In the currently used charts the `extraArgs` chart option does not work, hence we cannot pass the `--json` option to the minio server.

Having enabled the json logging, we will be able to properly correlate log eventsas described in [1]. 

[1] https://jira.d2iq.com/browse/D2IQ-66612

### DIFF
--- a/staging/velero/Chart.yaml
+++ b/staging/velero/Chart.yaml
@@ -16,4 +16,4 @@ name: velero
 sources:
 - https://github.com/heptio/velero
 tillerVersion: '>=2.10.0'
-version: 3.0.3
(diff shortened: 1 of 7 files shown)


vespian:
@shaneutt Yes, I tested it manually by launching konvoy cluster using the fork of the charts repo with the change in question. Minio backend started properly. Let me know and I can DM you the login creds for the cluster if you would like to have a closer look.

Is this sufficient or is there something more I can doublecheck?

PR #549, opened 2020-04-14
chore: bump kind image version of CI exclude test directory from grep result

alejandroEsc:
Additional small items in attempt to improve CI a little.
Bumped quay image for Helm since we are now using 2.16.5 in most places. 

Currently attempting to figure out why certain charts are being tested.


Makefile
```diff
@@ -1,4 +1,4 @@
-HELM_VERSION := v2.13.0
+HELM_VERSION := v2.16.5
```
joejulian, line 1:
```suggestion
HELM_VERSION := v2.16.6
```

PR #553, opened 2020-04-17
Additional nonResourceURLS to opsportal-kibana-view ClusterRole

geekbass:
Per D2IQ-66863, Adding additional nonResourceURLs to the opsportal-kibana that will allow a user "view" type of permissions within Kibana addon.

Currently a user is not able:
 - To use the Discover tab which is a requirement in our Product for viewing logs for Kubernetes, Konvoy and Kommander by default.
 - Use the Timelion app which comes preinstalled in our Kibana. 

The following nonResourceURLs have been inspected to ensure that users with the ClusterRole are not able create/delete Indexes or use the Dev Tools tool to do anything that would be considered destructive in Elasticsearch.

### DIFF
--- a/stable/opsportal/templates/kibana-roles.yaml
+++ b/stable/opsportal/templates/kibana-roles.yaml
@@ -57,7 +57,11 @@ rules:
 - nonResourceURLs:
   - {{ .Values.kibanaRBAC.path | trimSuffix "/"}}
   - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/api/saved_objects/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/elasticsearch/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/api/timelion/run
   verbs:
   - get
   - head
+  - post
 {{- end }}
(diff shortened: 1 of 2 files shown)


stable/opsportal/templates/kibana-roles.yaml
```diff
@@ -54,7 +54,11 @@ rules:
 - nonResourceURLs:
   - {{ .Values.kibanaRBAC.path | trimSuffix "/"}}
   - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/api/saved_objects/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/elasticsearch/*
+  - {{ .Values.kibanaRBAC.path | trimSuffix "/" }}/api/timelion/run
   verbs:
   - get
   - head
+  - post
```
samvantran, line 63: `view` role with `POST`? Is this like opsportal needing to post to graphql?

cc @jr0d

juliangieseke: KSphere UI does POST requests to fetch information ("viewing things") too. seems like Kibana UI does a similar thing? not all APIs follow REST standards 🤷‍♂️

juliangieseke: we also should extract kibana chart from opsportal… AFAIK there is a JIRA somewhere

PR #556, opened 2020-04-20
kubecost: Fix cm name templating, shorten job names

gracedo:
Fix configmap name templating, shorten the job names to avoid hitting k8s limits

Updates base kubecost as well which is necessary for future work integrating kubecost into kommander.


shaneutt:
@gracedo please don't hesitate to reach out to @mesosphere/sig-ksphere-catalog if you need any assistance with the CI here :vulcan_salute:

makkes:
Something is not right with Dispatch. It fails to create a kind cluster:
```
ERROR: failed to create cluster: docker run error: command "docker run --hostname chart-testing-control-plane --name chart-testing-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=chart-testing --publish=127.0.0.1:0:6443/TCP tmp-dispatch-kind:latest" failed with error: exit status 125
```
Funny enough, I stumbled upon the exact same error today locally and after I restarted Docker (for Mac), it worked. :-)

PR #558, opened 2020-04-21
chore: bump kcl version

hectorj2f:
Bump KCL to use the latest kcl version


stable/kommander/requirements.lock
```diff
@@ -4,12 +4,12 @@ dependencies:
   version: 0.1.6
 - name: kommander-cluster-lifecycle
   repository: https://mesosphere.github.io/kommander-cluster-lifecycle/charts
-  version: 0.5.3
+  version: v0.5.4
```
makkes, line 7: I can't tell whether that's valid or not, just pointing out that you added a `v` in front here.

hectorj2f: I didn't, helm did it `v2.16.5`

hectorj2f: this is the result of `helm dep update`

PR #561, opened 2020-04-22
chore: added impersonation headers

sebbrandt87:
- kommander added impersonation headers
- opsportal added impersonation headers

[D2IQ-67264]

[D2IQ-67264]: https://jira.d2iq.com/browse/D2IQ-67264


jieyu:
@sebbrandt87 is bumping the init container version required for this work? If not, I'd like not to do that in this PR given extra-steps tend to break from version to version.

sebbrandt87:
> @sebbrandt87 is bumping the init container version required for this work? If not, I'd like not to do that in this PR given extra-steps tend to break from version to version (i observed several in the past)

that is sad to hear, but I can remove it @jieyu for sure! :) thanks for letting me know

PR #562, opened 2020-04-22
chore: add default kubeaddons prometheus label for service monitors

hectorj2f:


stable/cluster-autoscaler/values.yaml
```diff
@@ -191,6 +191,7 @@ serviceMonitor:
    ## [Kube Prometheus Selector Label](https://github.com/helm/charts/tree/master/stable/prometheus-operator#exporters)
   selector:
     prometheus: kube-prometheus
+    release: prometheus-kubeaddons
```
joejulian, line 194: This should be in the addon, not the chart default values.

alejandroEsc: i think that is arguable. There are enough konvoy specific items i dont see that as an issue, but yeah maybe those should be revisited.

hectorj2f: @joejulian are u blocking a PR for having a label with kubeaddons? This is a big contradiction if you consider the rest of charts and their default values.

joejulian: It's not a label, it's a selector. By changing this to something d2iq specific when it can just as easily be overridden in the addon, we're diverging even further from upstream and making this chart harder to maintain. These kind of changes should be done in the addon record.

joejulian: Wait... My bad. I read "prometheus" and missed that this wasn't prometheus and is rather our non-upstream chart. I'm sorry for that misunderstanding. I'm hitting approve.

hectorj2f:
I messed it up with the commits, I am gonna open a new PR 🤦

PR #571, opened 2020-04-30
Add chart for konvoysoaktests addon

rgo3:
This PR is the start for an addon that we will use to run tests on the konvoy-soak cluster. The first test being a CronJob that will run twice a week, testing the new autoscaler feature.


stable/konvoysoaktests/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: konvoysoaktests
+home: https://github.com/mesosphere/konvoy
```
makkes, line 3:
```suggestion
home: https://github.com/mesosphere/konvoy-soak
```
Maybe?

hectorj2f: no, the code will be living under konvoy to avoid having to manage go-k8s tests deps in konvoy-soak.

hectorj2f: Besides, it is the best place to validate any coming soak test

PR #572, opened 2020-05-01
kommander-thanos: Upgrade thanos

gracedo:
https://jira.d2iq.com/browse/D2IQ-67855
Upgrade thanos subchart in kommander-thanos to latest version.


stable/kommander/charts/kommander-thanos/requirements.yaml
```diff
@@ -1,4 +1,4 @@
 dependencies:
   - name: thanos
-    version: 0.3.9
+    version: 0.3.20
```
shaneutt, line 3: Just double checking, we are jumping 11 versions yes? Have we done a spot check test to make sure that things look solid when doing this upgrade?

stable/kommander/charts/kommander-thanos/values.yaml
```diff
@@ -20,6 +20,8 @@ thanos:
     enabled: false
   bucket:
     enabled: false
+  rule:
```
shaneutt, line 23: Just for my own edification what are we doing here?

stable/kommander/charts/kommander-thanos/values.yaml
```diff
@@ -45,10 +47,18 @@ thanos:
     - "--grpc-client-tls-ca=/etc/certs/ca.crt"
     - "--grpc-client-server-name=server.thanos.localhost.localdomain"
     certSecretName: kommander-thanos-client-tls
+    labels:
```
alejandroEsc, line 50: Since this is our chart I am ok with these defaults, BUT these really belong in the addon right?

stable/kommander/charts/kommander-thanos/values.yaml
gracedo: just being extra explicit about the rule component being disabled along with the others here

stable/kommander/charts/kommander-thanos/values.yaml
gracedo: I did this here because it's only meant to be a subchart of kommander (hence `kommander-thanos` and also defined within that chart), but also I wanted to tie the thanos deployed in this chart specifically to this chart (and not to kommander, which the kubecost thanos will also be running under) which is why I defined the custom `app.kubernetes.io/parent-component: kommander-thanos` labels here. does that make sense?

stable/kommander/charts/kommander-thanos/values.yaml
```diff
@@ -45,10 +47,18 @@ thanos:
...
+      app.kubernetes.io/parent-component: kommander-thanos
+    grpc:
+      service:
+        matchLabels:
+          app.kubernetes.io/parent-component: kommander-thanos
```
branden, line 55: I haven't seen the `app.kubernetes.io/parent-component` label before, and googling it doesn't turn up anything. Is this ours? It won't break anything but we probably shouldn't use the `app.kubernetes.io` namespace for labels we come up with.

Is there a reason the existing selectors aren't sufficient? Maybe I'm missing something. I'd expect selecting by chart release name would be enough to prevent collisions with other thanos deployments: https://github.com/banzaicloud/banzai-charts/blob/3a18d063d72fd21d7f190f2b1be1e462d31f513a/thanos/templates/query-service.yaml#L25

stable/kommander/charts/kommander-thanos/values.yaml
alejandroEsc: Yeah, like i said it makes sense in the context of our chart, so Im ok with it.

stable/kommander/charts/kommander-thanos/requirements.yaml
gracedo: Yep upgrades all look good! Not a major version jump and I checked the diff, no really huge changes

stable/kommander/charts/kommander-thanos/values.yaml
gracedo: It is one that I came up with, couldn't find a standard one to reflect what I needed, I can see see about changing it to a mesosphere one (wasn't sure what the convention was).

The existing selectors are sufficient, I guess this was me making it extra clear where these components were coming from since the `name` label atm is set to `kommander-kubeaddons` which is pretty general wrt the kommander chart vs kommander-thanos subchart. But yeah, the matchLabels aren't required in this case, this was just me being a nitpick. If it's excessive (and also currently not very consistent since we can't define the same matchLabels on deployment) I'm happy to remove them too, don't feel too strongly about it either way. I can just leave this as a general thanos upgrade in that case.

branden: If the new match labels don't actually help the service match with its intended pods them IMO we should drop them. That also saves us the need to come up with label name.

gracedo: sgtm I shall remove! simplifies this pr a bunch :)

gracedo:
hmm looks like we need an approval from someone in @mesosphere/sig-ksphere-multi-cluster or @mesosphere/sig-ksphere-ui

PR #577, opened 2020-05-06
konvoysoaktests: fix args and clusterrole

rgo3:
Signed-off-by: Robin Gögge <r.goegge@outlook.com>


rgo3:
@hectorj2f should we wait to merge this until we have a new image on the mesosphere docker hub and can update it with this PR?

PR #578, opened 2020-05-06
[stable/awsebscsiprovisioner] Add CRD create permission to ebs-csi-snapshot-controller-role

branden:
https://jira.d2iq.com/browse/D2IQ-68046

The snapshot controller requires this permission on Kubernetes 1.17 to create the `volumesnapshot` CRD. Without this permission on Kubernetes 1.17, the `ebs-csi-snapshot-controller-0` pod fails with this error: `customresourcedefinitions.apiextensions.k8s.io is forbidden: User \"system:serviceaccount:kube-system:ebs-csi-snapshot-controller\" cannot create resource \"customresourcedefinitions\" in API group \"apiextensions.k8s.io\" at the cluster scope`.

## Testing

I tested this PR by deploying the latest Konvoy release (k8s 1.16) with the `awsebscsciprovisioner` addon disabled. Several addons that depend on it failed to deploy, as expected. I then deployed my local copy of the chart from this branch into the `kube-system` namespace with the chart values from https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml. I observed that the chart deployed successfully and all its pods were running. I then ran `konvoy deploy addons` and all addons deployed successfully.

I then repeated the test with a Konvoy release built from https://github.com/mesosphere/konvoy/pull/1504 and using k8s version 1.17.5.


stable/awsebscsiprovisioner/templates/roles.yaml
```diff
@@ -187,6 +187,9 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
     verbs: ["update"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["create"]
```
dkoshkin, line 192: Do we not need the other verbs like in https://github.com/mesosphere/charts/blob/2336e7191f0a62d189fd88450b3ea3d9ef465101/stable/awsebscsiprovisioner/templates/roles.yaml#L141?

hectorj2f: +1. yes it is weird we need create but not other verbs

branden: I only saw an error with creating a CRD, not listing etc. And this change was sufficient to fix it. I'll add the rest though, shouldn't hurt.

stable/awsebscsiprovisioner/templates/roles.yaml
```diff
@@ -187,6 +187,9 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
     verbs: ["update"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["create", "list", "watch", "delete"]
```
joejulian, line 192:
```suggestion
    verbs: ["create", "list", "get"]
```

I disagree, somewhat, with the previous conversation. It's considered best practice to apply the [principal of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). If create is the minimum this controller needs, that's what it should get.

I suspect, however, that the controller would actually need list, get, and update in order to check to see if the crd is installed, check to see if it's current, and to update it if the installed crd is out of date. I doubt it uses watch and we don't want CRDs being deleted by automation.

stable/awsebscsiprovisioner/templates/roles.yaml
dkoshkin: Maybe delete on cleanup?

stable/awsebscsiprovisioner/templates/roles.yaml
```diff
@@ -187,6 +187,9 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
     verbs: ["update"]
+  - apiGroups: ["apiextensions.k8s.io"]
```
hectorj2f, line 190: It is still weird why this permission is not set in the aws-csi-driver repository. Should we raise it there and propose this change in the upstream repo?

sebbrandt87: yes, we should raise it upstream

sebbrandt87:
wait ... it looks like I did a mistake with the snapshot-controller image tag ... so this could be the issue @hectorj2f @branden 
looking into it already because I started https://jira.d2iq.com/browse/D2IQ-68177

sebbrandt87:
I can confirm with the corrected snapshot-controller which is valid for 1.17 this issue does not exists anymore.

branden:
Closing in favor of #580.

PR #579, opened 2020-05-07
kommander: Add centralized kubecost

gracedo:
Adds centralized kubecost on kommander - runs a separate prometheus instance for kubecost and a separate thanos query which is set up to pull kubecost metric data remotely from managed cluster kubecost prometheis. Reuses existing centralized monitoring grafana.

## Outstanding work
- Licensing - but we can add these changes in later
- Alertmanager to monitor and alert on kubecost health - add later, non-blocking

## Testing
I have tested running this version of kommander with centralized kubecost and pointing the KCL controller manager to `dev-thanos` image which I built and pushed using [this branch](https://github.com/mesosphere/kommander-cluster-lifecycle/pull/515). I provisioned a cluster from kommander and was able to verify that the managed cluster's kubecost thanos ELB was getting pulled into the kommander kubecost thanos query stores. I was able to see data getting pulled from the managed cluster using the kommander kubecost thanos query UI. The kommander kubecost UI likewise is installed successfully along with all the other kubecost components.

1. Point kubeaddons-kommander branch to `configVersion: gracedo/test_kubecost`
1. Use `dev-thanos` image for KCL controller manager
     ```
    addonsList:
    - name: kommander
      enabled: true
      values: |
        kommander-cluster-lifecyle:
          controller:
            containers:
              manager:
                image:
                  tag: dev-thanos
    ```
    ^ This actually didn't work for me, so when the cluster came up, I edited the `kcl-cm` deployment to use `dev-thanos` image
1. `konvoy up`
1. [Provision a managed cluster](https://github.com/mesosphere/kommander-cluster-lifecycle#creating-konvoy-clusters)
1. Check thanos query UI for added store from managed cluster: `<CLUSTER_URL>/ops/portal/kommander/kubecost/query/stores`
1. Check Kubecost UI `<CLUSTER_URL>/ops/portal/kommander/kubecost/detail.html`. NOTE: Data won't actually be presented in the UI until approx 2hr after the cluster is joined. The Kubecost team has informed us that as long as we can see the data flowing in to Thanos Query, that should be good enough to confirm. I was able to keep managed clusters up long enough to verify that data shows up in kubecost.
1. Add license key for enterprise version on the Kubecost UI Home page `<CLUSTER_URL>/ops/portal/kommander/kubecost/overview.html`, via `UPGRADE` link on top right (ping me for the key). **licensing integration is TBD at this point, and blocking this release**
image

1. Kubecost grafana dashboards are installed into the centralized monitoring grafana instance: `<CLUSTER_URL>/ops/portal/kommander/monitoring/grafana` and are also linked to from the Kubecost UI. It is using Thanos query as a datasource so data should already be present in those dashboards.

image
image


gracedo:
@hector on a managed cluster, there would be 2 prometheus (1 monitoring, 1 kubecost). on kommander, there can be 3 if the cluster is attached to itself - I will look into this later, but it is possible we can just use the kommander kubecost prometheus instead of federating another kubecost once attached to itself (is it possible to use cluster placement constraints on federated resources to NOT federate to specifically kommander clusters, similar to how you can do this with konvoy/non-konvoy constraints?).

The reasoning behind a separate stack is to get kubecost MVP out and get some customer feedback, then work to fully integrate the product to use the existing monitoring stack. If it helps, the prometheus that is being run with kubecost is a significantly trimmed down instance that is only scraping kubecost-specific metrics, a fraction of the usual metrics ingested.

Re federated addon to deploy mtls-proxy, I was mainly following the existing pattern where karma and thanos define that here, but I believe it is because some of the variables are dependent on the kommander chart values which may require the federate resources to be defined here. I'll look and see what can be moved into kcl if possible!

hectorj2f:
@gracedo could we have this set to false by default https://github.com/mesosphere/charts/pull/579/files#diff-/ a6876f47273ad1ee05ea80fa7c22e139R198 and set to `true` on federated clusters ?

gracedo:
@hectorj2f I can default prometheus and grafana to `false` in our kubecost parent chart and enable in kommander and federated charts, is that what you're asking?

gracedo:
@hectorj2f sorry just looked at it some more, we have to default prometheus and grafana to on so that the kubecost chart is able to be deployed on its own. Otherwise an existing prom/graf instance would have to be supplied.

hectorj2f:
@gracedo gotcha!

stable/kommander/values.yaml
```diff
@@ -181,3 +181,119 @@ portalRBAC:
...
+        # Use kommander monitoring Grafana instance
+        domainName: kommander-kubeaddons-grafana.kommander.svc.cluster.local
+
+    ingress:
+      enabled: true
+      annotations:
+        kubernetes.io/ingress.class: traefik
+        ingress.kubernetes.io/auth-response-headers: X-Forwarded-User
+        traefik.frontend.rule.type: PathPrefixStrip
+        traefik.ingress.kubernetes.io/auth-response-headers: X-Forwarded-User,Authorization,Impersonate-User,Impersonate-Group
```
jr0d, line 216: Does kubecost require impersonation headers?

gracedo: woops good question, no they don't, must have been a bad copy paste 🙃 I'll remove!

PR #580, opened 2020-05-07
feat(awsebscsiprovisioner): update images

sebbrandt87:
- fixing a wrong image usage for snapshot-controller
- logic for same image tag being used for snapshotter as
snapshot-controller
- updated images to compatible latest
- added replicas for SnapshotController
- only leader-election enabled if replicas > 1
- added CRD install for v1beta1
- added nodeSelector, priorityClassName and tolerations for
SnapshotController

Notes:

Closes #578

[D2IQ-68177] #comment See GitHub

### DIFF
--- a/stable/awsebscsiprovisioner/Chart.yaml
+++ b/stable/awsebscsiprovisioner/Chart.yaml
@@ -6,7 +6,8 @@ maintainers:
   - name: alejandroEsc
   - name: gpaul
   - name: hectorj2f
-version: 0.3.6
+  - name: sebbrandt87
+version: 0.3.7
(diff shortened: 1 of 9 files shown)


sebbrandt87:
> > atm with this an "upgrade" would fail, and there would be the need to
> > manual remove the v1alpha1 snapshot CRDs. Which would also purge
> > snapshots.
> 
> Is there any way to have this not happen, or is there no upstream fix for this?

not from what is documented. By kubernetes definition this is what the alpha to beta change always can contain - breaking changes.
see https://github.com/kubernetes-csi/external-snapshotter#upgrade-from-v1alpha1-to-v1beta1

PR #584, opened 2020-05-08
feat(gcpdisk-csi-driver): update images

sebbrandt87:
- logic for same image tag being used for snapshotter as
snapshot-controller
- updated images to compatible latest
- added replicas for SnapshotController
- only leader-election enabled if replicas > 1
- added CRD install for v1beta1
- added tolerations

Notes:

changes as for #580 #581

[D2IQ-68177] #comment See GitHub

### DIFF
--- a/stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
+++ b/stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
@@ -6,7 +6,7 @@ metadata:
 {{ include "gcpdisk.labels" . | indent 2 }}
 spec:
   serviceName: "csi-gce-pd"
-  replicas: 1
+  replicas: {{ .Values.replicas }}
   selector:
     matchLabels:
       app: gcp-compute-persistent-disk-csi-driver
@@ -15,20 +15,21 @@ spec:
 {{ include "gcpdisk.labels" . | indent 6 }}
         app: gcp-compute-persistent-disk-csi-driver
     spec:
-      # Host network must be used for interaction with Workload Identity in GKE
-      # since it replaces GCE Metadata Server with GKE Metadata Server. Remove
-      # this requirement when issue is resolved and before any exposure of
-      # metrics ports
-      hostNetwork: true
       serviceAccountName: csi-gce-pd-controller-sa
       nodeSelector:
         beta.kubernetes.io/os: linux
       priorityClassName: system-cluster-critical
+{{- if .Values.tolerations }}
+      tolerations:
+{{ toYaml .Values.tolerations | indent 8 }}
+{{- end }}
       containers:
         - name: csi-provisioner
           image: {{ .Values.provisioner.image.repository }}:{{ .Values.provisioner.image.tag }}
           args:
+            - --v=5
             - --csi-address=$(ADDRESS)
+            - --metrics-address=:22011
             - --timeout={{ .Values.provisioner.timeout }}
             - --retry-interval-start={{ .Values.provisioner.retryIntervalStart }}
             - --retry-interval-max={{ .Values.provisioner.retryIntervalMax }}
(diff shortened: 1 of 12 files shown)


stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -15,52 +15,48 @@ spec:
...
       priorityClassName: system-cluster-critical
+{{- if .Values.tolerations }}
+      tolerations:
+{{ toYaml .Values.tolerations | indent 8 }}
+{{- end }}
       containers:
         - name: csi-provisioner
           image: {{ .Values.provisioner.image.repository }}:{{ .Values.provisioner.image.tag }}
           args:
+            - --v=5
```
sreis, line 30: Configurable via `values.yaml`?

sreis: Just noticed it seems to be the same setting in the other csi drivers. Maybe it should be a setting available for all of them? wdyt?

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -15,52 +15,48 @@ spec:
...
-            - --v=5
-          env:
-            - name: ADDRESS
-              value: /csi/csi.sock
-          volumeMounts:
-            - name: socket-dir
-              mountPath: /csi
         - name: csi-snapshotter
-          image: {{ .Values.snapshotter.image.repository }}:{{ .Values.snapshotter.image.tag }}
+          image: "{{ .Values.snapshotter.image.repository }}:{{ include "gcpdisk.snapshotter.image.tag" . }}"
```
hectorj2f, line 52: why did you change from include to `{{ .Values.snapshotter.image.tag }}` ?

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -15,52 +15,48 @@ spec:
...
           args:
+            - --v=5
             - --csi-address=$(ADDRESS)
+            - --metrics-address=:22014
```
hectorj2f, line 56: are these ports the default ones ? why to set them then ?

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
```diff
@@ -88,6 +88,10 @@ spec:
           args:
             - --v=5
             - --csi-address=$(ADDRESS)
+            - --metrics-address=:22013
```
hectorj2f, line 91: if these ports are the default ones, I don't see much interest on adding them

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: could be done for all sure, but it is not right now.
We can tackle it in another ticket I will create.

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: `--metrics-address: The TCP network address where the prometheus metrics endpoint will run (example: :8080 which corresponds to port 8080 on local host). The default is empty string, which means metrics endpoint is disabled.`

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: see https://github.com/mesosphere/charts/pull/584/files#r430020207

stable/gcpdisk-csi-driver/templates/csi-gcpdisk-controller.yaml
sebbrandt87: https://jira.d2iq.com/browse/D2IQ-68803

PR #585, opened 2020-05-08
feat(gcpdiskprovisoner): updated snapshotclass

sebbrandt87:
- added detection for Kubernetes >= Minor 17

plays with to #584

[D2IQ-68177] #comment gcpdiskprovisioner chart changes done.

### DIFF
--- a/stable/gcpdiskprovisioner/values.yaml
+++ b/stable/gcpdiskprovisioner/values.yaml
@@ -11,3 +11,4 @@ storageclass:
   topologyZone: []
 snapshotclass:
   isDefault: true
+  deletionPolicy: Delete
--- a/stable/gcpdiskprovisioner/Chart.yaml
+++ b/stable/gcpdiskprovisioner/Chart.yaml
@@ -4,5 +4,5 @@ description: Google Compute Engine Persistent Disk (GCE PD) storage provisioner 
 name: gcpdiskprovisioner
 maintainers:
   - name: sebbrandt87
-version: 0.1.0
+version: 0.2.0
 home: https://kubernetes.io/docs/concepts/storage/storage-classes/
--- a/stable/gcpdiskprovisioner/templates/_helpers.tpl
+++ b/stable/gcpdiskprovisioner/templates/_helpers.tpl
@@ -1,0 +1,13 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+Switching snapshot classes api version depending on our detected Kubernetes version >= Minor 17
+It works together with our csi-driver
+*/}}
+{{- define "gcpdiskprovisioner.snapshotclass.apiversion" -}}
+{{- if or (gt (.Capabilities.KubeVersion.Minor | int) 17) (eq (.Capabilities.KubeVersion.Minor | int) 17) -}}
+v1beta1
+{{- else -}}
+v1alpha1
+{{- end -}}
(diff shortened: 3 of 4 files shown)


stable/gcpdiskprovisioner/values.yaml
```diff
@@ -11,3 +11,4 @@ storageclass:
   topologyZone: []
 snapshotclass:
   isDefault: true
+  deletionPolicy: Delete
```
sreis, line 14: Are you sure this is safe? https://kubernetes.io/docs/concepts/storage/volume-snapshots/#delete

hectorj2f: Yeah, this should not be set by us.

sebbrandt87: we set the same for default `StorageClass`.
Example: If I delete my kubernetes managed resources, I want that these are cleaned up.
And if I would have PVs that I want to retain, it is a common use case to `patch` these. 
I mean what is the reason than behind not setting "Retain" for `StorageClass` as well?

sreis: @sebbrandt87 You're not wrong but I think the defaults should be safe since customers will ship them most likely without considering all the options.

Pinging @samba and @GoelDeepak for final call on this one since ultimately it's a product decision.

sebbrandt87: Snapshot specific

`beta` details: https://kubernetes.io/docs/concepts/storage/volume-snapshots/#lifecycle-of-a-volume-snapshot-and-volume-snapshot-content
`alpha` blog article https://kubernetes.io/blog/2019/01/17/update-on-volume-snapshot-alpha-for-kubernetes/

GoelDeepak: I would expect if somebody is configuring snapshot they would want to keep them. Otherwise, why somebody would take a snapshot. This is also what makes snapshot different from general storage. Is it possible to disable the snapshot by default? If a customer wants they could enable it and also decide their own retention policy.

sebbrandt87: It is just one default snapshot class. If it is used or not customer can decide.

samba: @sebbrandt87 the problem is, we have long-standing relationships with customers whose support staff (including overseas contractors) don't pay attention to details like this, and simply follow instruction manuals. They won't look for these details, and they probably wouldn't understand the implications if they stumbled upon it. They will shoot themselves in the feet with a gun they didn't know they were carrying. 

Though orthogonal, existing defaults for `reclaimPolicy` were reasoned: if a user `kubectl delete pvc <name>`, we expect that this was done intentionally, and it was better to let the API's own rules (including RBAC) govern the lifecycle of resources it created (e.g. EBS volumes). 

With the current definition, `deletionPolicy: Delete`, it dictates basically the same behavior for snapshots. 

In this case I think consistency is better here, at least for the present time, for the Konvoy 1.5 release timeframe. 

@GoelDeepak @sreis @joejulian we should discuss whether changing these defaults in 1.6 is appropriate.

GoelDeepak: I am fine with that. It will also let us collect customer feedback

PR #588, opened 2020-05-14
fix: federate karma-proxy to any kind of clusters

armandgrillet:
We want to federate Prometheus Alert Manager and Karma proxy when non-Konvoy
clusters get attached to Kommander (D2IQ-66286) thus we do not need to
federate karma-proxy only to Konvoy clusters anymore.


tillt:
@armandgrillet  note that we do actually not want to federate Karma. We need to federate the Karma proxy and the Prometheus Alertmanager.

armandgrillet:
@tillt Thanks for the note, I have updated the commit message while doing the squash and merge to mention "Karma proxy" instead of just "Karma".

PR #598, opened 2020-05-25
chore: transition to yakcl

makkes:
This replaces the old kcl dep with the three new charts of the yakcl
repo. The chart configuration values are kept intact for an easier
step-by-step transition process.


makkes:
dup of #596

PR #599, opened 2020-05-26
chore: enable yakcl-provisioning until it is deployed by konvoy

hectorj2f:
yakcl-federation requires the konvoycluster crd, therefore we need to deploy it for now (temporary) until konvoy deploys it by default.


stable/kommander/requirements.lock
```diff
@@ -4,10 +4,13 @@ dependencies:
   version: 0.1.9
 - name: yakcl-federation
   repository: https://mesosphere.github.io/yakcl/charts
-  version: 0.0.1
+  version: v0.0.1
```
juliangieseke, line 7: this is so weird… 🤔

hectorj2f: helm dep update fellas :D v2.16.7

juliangieseke:
commites arent signed :(

hectorj2f:
I will squash and sign them...

PR #608, opened 2020-05-29
kommander: Update karma dep to 1.4.1

gracedo:
Karma dep patch version bump


samvantran:
@gracedo you might need to bump the kommander chart version for all your PRs

PR #610, opened 2020-05-30
awsebscsiprovisioner: make upgrade automatic

jieyu:
We observed that the upgrade e2e tests on aws always fail with aws csi
driver failing to install v1beta1 volumesnapshot CRDs. This starts to
happen after we bumped the master konvoy to use 1.17 by default.

The issue is caused by this PR
https://github.com/mesosphere/charts/pull/580. On 1.17 clusters, the
chart will try to install v1beta1 CRD. On a 1.16.x cluster, v1alpha1 CRD
exists. The upgrade won't work because only v1beta1 version is listed in
the CRD definition while the storage version is v1alpha1 still.

The above PR states that:
```
Previously, upgrade is manual
```

This patch makes this process automatic.

https://jira.d2iq.com/browse/D2IQ-68987

Tested on a 1.17 cluster by installing 0.3.6 first and then upgrade to this version.


jieyu:
This is causing e2e failures. I don't think asking customer to manually fix those during upgrade is reasonable from product's perspective. We need to automate it. We'll need to adjust the release notes once this change is in.

> The delete permission, if you really want it to be done by the pre hook script, should only be present in that case not in all others.

The service account used (as well as the role and rolebindings) will be deleted after upgrade is finished.

> Also there is still a possible case that the snapshotter itself gets restarted and recreates the Alpha CRDs.

OK, i'll scale down the controller deployment during the upgrade.

jieyu:
Note that I cannot use `scale` to stop the controller (instead, I used delete) because of this helm issue:
https://github.com/helm/helm/issues/1873

The 3-way merge is only available in helmv3 unfortunately.

sebbrandt87:
This looks better 👍 
> The service account used (as well as the role and rolebindings) will be deleted after upgrade is finished.

Anyway, I always try to avoid to set permissions used only in another "case".

Approve and merge.

PR #611, opened 2020-05-31
chore: bump yakcl to v0.0.4

hectorj2f:
I keep yakcl-provisioning enabled by default in the chart, but I'll disable it in kubeaddons so we can get an ease transition for everyone.


stable/kommander/requirements.lock
```diff
@@ -4,13 +4,13 @@ dependencies:
...
-  version: 0.0.3
+  version: v0.0.4
 - name: yakcl-licensing
   repository: https://mesosphere.github.io/yakcl/charts
-  version: 0.0.3
+  version: v0.0.4
 - name: yakcl-provisioning
   repository: https://mesosphere.github.io/yakcl/charts
-  version: 0.0.3
+  version: v0.0.4
```
dkoshkin, line 13: Is it `v0.0.4` or `0.0.4`?

hectorj2f: This is what helm does for me since the early days

PR #612, opened 2020-06-02
cluster-autoscaler: Add grafana dashboard

gracedo:
Used https://github.com/skyscrapers/charts/blob/master/cluster-monitoring/dashboards-default/grafana-dashboard-k8s-autoscaler.json, made some slight modifications.

Tested this on our dry-run cluster today. We didn't get autoscaling to work so the metrics are empty, however I was able to verify that the autoscaler metrics pipeline was working as expected and getting scraped by prometheus so once there's data, it should show up in grafana.

image
image
image

I also tested a basic helm install of the chart on an existing cluster and saw the grafana dashboard get installed.


hectorj2f:
Thanks @gracedo 👍

PR #614, opened 2020-06-03
feat: update traefik chart

sebbrandt87:
- version 1.7.24
- grabbed upstream changes
- caSecretName not needed anymore
as it is now a combination of
`secrets` and `extraCerts`
- accessLogs filter is included
- mTLS is included

[D2IQ-65688]

[D2IQ-65688]: https://jira.d2iq.com/browse/D2IQ-65688


sebbrandt87:
we may still need `caSecretName` but I would try to avoid it if we can.

jr0d:
@GoelDeepak Customers are currently using caCertname for custom certs, right? We would need an upgrade path for this change.

GoelDeepak:
> @GoelDeepak Customers are currently using caCertname for custom certs, right? We would need an upgrade path for this change.

yeah, we need to do it in backward compatible way

sebbrandt87:
caSecretName is still in, I just reverted that 4days ago to be in again.

sebbrandt87:
oh sorry I see I missed one part.
will readd

GoelDeepak:
I have verified the changes against `stable/traefik`. We need these changes for https://jira.d2iq.com/browse/COPS-6284

staging/traefik/values.yaml
```diff
@@ -17,7 +17,6 @@ testFramework:
 
 ## can switch the service type to NodePort if required
 serviceType: LoadBalancer
-extraServicePorts: []
```
jieyu, line 20: hum? KBA currently relies on this to specify the velero port 9000. Removing this without adjusting kba would break the user.

PR #615, opened 2020-06-03
Bump UI chart to 0.8.8

DanielMSchmidt:
chart enables kubecost and impersonate user just for kommander


stable/kommander/values.yaml
```diff
@@ -164,6 +164,8 @@ grafana:
           mountPath: /var/lib/grafana/shared-plugins/
 
 kommander-ui:
+  showCost: true
```
juliangieseke, line 167: 🎉

PR #626, opened 2020-06-10
[kommander] Rewrite Kubecost URL to include trailing slash

branden:
https://jira.d2iq.com/browse/D2IQ-68100

Kubecost assumes its URL has a trailing slash. If served without a trailing slash, the web app doesn't work. This PR causes Traefik to redirect `/ops/portal/kommander/kubecost` to `/ops/portal/kommander/kubecost/` so that Kubecost's URL always has a trailing slash.

## Testing

I tested this PR by deploying a Konvoy cluster with the kommander addon disabled, then installing a local copy of this branch's kommander chart with the chart values defined in the kommander addon. Then I browsed to `/ops/portal/kommander/kubecost`, saw that I was redirected to `/ops/portal/kommander/kubecost/`, and saw that the resulting Kubecost page listed this cluster and its projected cost.

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.8.15
+version: 0.8.16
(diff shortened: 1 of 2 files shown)


juliangieseke:
there are no changes left in this PR, is that correct?

branden:
Thanks for pointing it out @juliangieseke, I must've pushed a bad rebase or something.

PR #632, opened 2020-06-11
kommander: Update kubecost chart version

gracedo:
Use latest kubecost parent chart version


gracedo:
@faiq fyi for this PR I had to `/test` in order to trigger the tests

PR #635, opened 2020-06-15
kommander: Upgrade grafana subchart

gracedo:
Upgrade grafana subchart of kommander to `4.6.3` (latest version before a major version jump) from `4.5.*`


gracedo:
@juliangieseke I didn't want to risk breaking something so close to release :) I don't think there would actually be a breaking change but I'd rather do that bump after we release! #justincase

juliangieseke:
fair enough :D

PR #636, opened 2020-06-16
kubecost: Fix prom server cm name

gracedo:
We recently saw e2e tests failures due to kommander failing to come up properly. There was a bug after the work to override kubecost component names in kommander that was causing a post-install job to fail due to referencing the wrong cm name. I made the fix to use the `fullnameOverride` value for prometheus server if configured and verified that kommander deployed successfully with the fix.

## Testing
I hosted the kubecost chart on my personal GH then updated kommander to pull in that updated version, and hosted the kommander chart as well. I used a `kubeaddons-kommander` branch pointing to this kommander chart and tested it out. I verified that the job ran successfully and modified the prometheus server cm.


hectorj2f:
👍

PR #637, opened 2020-06-16
kommander: update ingress priority for Grafana

armandgrillet:
This is now set at the same value as for karma, `4`.
This resolves the bug https://jira.d2iq.com/browse/D2IQ-68570
which was making Grafana inacessible from the Kommander UI
after self-attaching a Konvoy cluster.

This change was tested manually by updating the ingress
kommander-kubeaddons-grafana on a running Konvoy
cluster after self-attachment.

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.8.18
+version: 0.8.19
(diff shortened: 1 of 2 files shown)


juliangieseke:
as stated by armand, its related to attaching kommander to itself. we need to make sure kommander (host) ingress entries override the ones deployed from yakcl (managed)

armandgrillet:
Traefik offers a UI accessible using Konvoy that gives more details regarding the overrides:
Before attachment:
PS 2020-06-16 à 16 44 41
After attachment, without fix:
Before
After attachment, with fix:
After

armandgrillet:
@gracedo I will do the changes your recommended in a second PR as they are not strictly related to the bug https://jira.d2iq.com/browse/D2IQ-68570. I tried to check what you mentioned and couldn't see the issues but I think it was due to my Konvoy version thus I'll try again.

armandgrillet:
@gracedo From my test: before attaching, `ops/portal/kommander/kubecost/` works but not `ops/portal/kubecost/`. After attaching, both work.

PR #639, opened 2020-06-16
Fix for pull request matcher with branch names that include /

faiq:
We use this https://godoc.org/github.com/gobwas/glob to do our matching in dispatch for pull requests. The reason this one wasn't triggering was because `/` was included in our branch names and `/` is a separator character.


Dispatchfile
```diff
@@ -64,6 +65,6 @@ dind_task("test", inputs=[git], deps=["lint"], steps=[
   )
 ])
 
-action(tasks=["test"], on=pullRequest(branches=["*"]))
+action(tasks=["test"], on=pullRequest())
```
orsenthil, line 68: Thank you. Please add a note on why removal of this `branches=["*"]` and having just the default worked, and will approve this PR.

The note could be in the comment itself because I could gather that from the description.

faiq: The description in the pull request isn't enough?

orsenthil: > The reason this one wasn't triggering was because / was included in our branch names and / is a separator character.

@faiq - The missing part for me was - How did removing the `branches=["*"]` solved the problem? Was the default good enough to recognize any branch pattern? Or having the branch pattern was a mistake?

PR #643, opened 2020-06-17
fix: set some initial limits to prometheus components

hectorj2f:
The prometheus components did not set any limit to the mem and cpu resources.
[x] Set limits to alertmanager
[x] Set limits to prometheus-server


hectorj2f:
Does anyone has an idea why it fails ? The error doesn't look related to this chart :/

PR #644, opened 2020-06-17
update the catalog CRD to support the kubeaddons v1beta2 api

joejulian:


armandgrillet:
`helm dep update` should help making the CI 🟢 by updating the `requirements.lock`.

joejulian:
looks like I'll have to do this in two PRs since I can't update the lock file until the new catalog chart is published.

PR #646, opened 2020-06-18
kommander: do not recreate the kubecost-query-stores configmap on upgrade

dlipovetsky:
Fixes https://jira.d2iq.com/browse/D2IQ-69505

Note: If the configmap already exists and has the `pre-upgrade` hook annotation, the annotation will remain, because the configmap will not be deleted/re-created during the upgrade. This seems correct, but also non-intuitive, and I'm still trying to understand it.

### DIFF
--- a/stable/kommander/Chart.yaml
+++ b/stable/kommander/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.8.22
+version: 0.8.23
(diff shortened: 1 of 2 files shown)


dlipovetsky:
> You'll need to bump the chart.
Thanks, will bump the version!

> Also how did you test?
1. Applied the changes and published the chart in my own helm repo.
2. In a private branch in https://github.com/mesosphere/kubeaddons-kommander , I updated the kommander `ClusterAddon` manifest to point to my own helm repo. 
3. I modified the cluster configuration to read the kommander `ClusterAddon` manifest from my branch.
4. I ran `konvoy deploy addons`.

gracedo:
You also need a rebase, looks like master has kommander at 0.8.22 already

gracedo:
ah, looks like you need to sign all your commits
[```The base branch requires all commits to be signed```](https://help.github.com/articles/about-gpg/)

PR #649, opened 2020-06-18
chore: bumps yakcl to v0.1.2

tillt:


stable/kommander/requirements.lock
```diff
@@ -4,13 +4,13 @@ dependencies:
   version: 0.1.10
 - name: kommander-federation
   repository: https://mesosphere.github.io/yakcl/charts
-  version: v0.1.0
+  version: 0.1.2
```
tillt, line 7: This seems unintended. Need to find out why that "v" prefix got changed.

stable/kommander/Chart.yaml
```diff
@@ -1,10 +1,10 @@
 apiVersion: v1
-appVersion: 1.1.0
+appVersion: 1.1.2
```
makkes, line 2:
```suggestion
appVersion: 1.1.0
```
this doesn't change.

stable/kommander/requirements.lock
makkes: nvm, this file is auto-generated, anyway.

tillt:
> you're not bumping kommander, but some kommander components…

you are perfectly right :)

PR #651, opened 2020-06-18
[autoscaler] fix: available nodes grafana gauge color

samvantran:
**What type of PR is this?**
Fix

Noticed this in the MWT. Previously it would show 100% available nodes in red and 80% as green.
This reverses the coloring. See screenshots:

Before:

Screen Shot 2020-06-16 at 1 37 49 PM

After: 

Screen Shot 2020-06-16 at 1 06 00 PM

```release-note
NONE
```
Not really worth calling out a color change on a grafana dashboard

**Checklist**

* [] The commit message explains the changes and why are needed.
* [] The code builds and passes lint/style checks locally.


stable/cluster-autoscaler/charts/grafana/templates/k8s-autoscaler.yaml
```diff
@@ -58,7 +58,7 @@ data:
       "editable": true,
       "gnetId": 3831,
       "graphTooltip": 0,
-      "id": null,
+      "id": 36,
```
gracedo, line 61: this needs to be `null`

samvantran, line 61:
```suggestion
      "id": null,
```

samvantran: i was wondering if that was going to have any effect. thanks, fixed in b19a677

gracedo: I always forget you can suggest commits, I shoulda done that lol

gracedo: np, we ran into this a lot in dc/os when people started contributing dashboards - the trick is remembering to check the "export for sharing externally" box
image

samvantran: ty 🙇

PR #656, opened 2020-06-22
kommander: Fix missing kubecost cm on upgrade

gracedo:
**What type of PR is this?**
Bug

We saw upgrade tests beginning to fail in the konvoy nightlies after removing the `pre-upgrade` hook to the job that creates the `kubecost-thanos-query-stores` cm. I've fixed this to instead run a bash script that checks whether the cm exists or not and only creates it if it does. I applied the same changes to the centralized thanos query stores cm job as well since the same problem applies.

relates to D2IQ-69505

I tested this using my kubeaddons-kommander branch `test-kubecost` to deploy this updated version of the chart. I tested a fresh install (job pod logs):
```
checking if kubecost-thanos-query-stores configmap exists
Error from server (NotFound): configmaps "kubecost-thanos-query-stores" not found
kubecost-thanos-query-stores configmap does not exist - creating
configmap/kubecost-thanos-query-stores created
```
as well as an upgrade where the cm already exists:
```
checking if kubecost-thanos-query-stores configmap exists
kubecost-thanos-query-stores configmap already exists - no need to create
```

```release-note
NONE
```

**Checklist**


stable/kommander/templates/kubecost/kubecost-query-stores-configmap.yaml
```diff
@@ -4,19 +4,48 @@
...
+          - "-c"
+          - |
+            /bin/bash <<'EOF'
+            set -o nounset
+            set -o errexit
+            set -o pipefail
+
+            echo "checking if {{ . }} configmap exists"
+
+            RES=$(kubectl get configmap {{ . }} || true)
```
dlipovetsky, line 39: Do you want the script to fail if `kubectl get configmap` fails? If you do, I believe you will need to enable `pipefail` within the subshell. If you just want to ignore a 404 (which is normal, if the configmap wasn't created in the previous version of the Chart, and is to be created on upgrade), then I recommend using `--ignore-not-found=true`.

```suggestion
            RES=$(set -o errexit; set -o pipefail; kubectl get --ignore-not-found=true configmap {{ . }})
```

stable/kommander/charts/kommander-thanos/templates/query-stores-configmap.yaml
```diff
@@ -4,19 +4,48 @@
...
+          - "-c"
+          - |
+            /bin/bash <<'EOF'
+            set -o nounset
+            set -o errexit
+            set -o pipefail
+
+            echo "checking if {{ . }} configmap exists"
+
+            RES=$(kubectl get configmap {{ . }} || true)
```
dlipovetsky, line 39: Same as my other comment

stable/kommander/templates/kubecost/kubecost-query-stores-configmap.yaml
gracedo: good call! I'll pass in the `ignore-not-found` flag which is cleaner :) (`set -o errexit` and `set -o pipefail` are set at the top of script)

dlipovetsky: I think we were both a little wrong about the bash options :joy:

Bash subshells do not inherit options intuitively.

For years, I had thought that a command-substitution subshell does not inherit `pipefail`, but in fact it does, so it is not necessary to set it in any subshell.

On the other hand, a command-substitution subshell does not inherit `errexit`, so it is necessary to set it in every subshell. (There are two workarounds: Use bash in its non-default POSIX mode, or use bash >= 4.4 and set the `inherit_errexit` option in the containing shell.)

For additional details and some good examples, please see https://dougrichardson.us/2018/08/03/fail-fast-bash-scripting.html

dlipovetsky: I think that, if we don't set `errexit` in the subshell, then, if kubectl throws an error for some reason (e.g. network issue), then the script will drop down to "configmap already exists."

PR #660, opened 2020-06-24
fix(kubecost): define extraFlags

sebbrandt87:
**What type of PR is this?**

Bug

- so that we won't hit a thanos-sidecar error
- enable admin-api
- enable wal-compression
- log.format set to json!
- log.level set to info
- latest prometheus version
- latest alertmanager version
- removed deprecated storage.tsdb.retention

[D2IQ-69703]

[D2IQ-69703]: https://jira.d2iq.com/browse/D2IQ-69703

```release-note
NONE
```

**Checklist**

### DIFF
--- a/stable/kubecost/values.yaml
+++ b/stable/kubecost/values.yaml
@@ -87,6 +87,12 @@ cost-analyzer:
       # This overrides the cluster_id set in prometheus.server.global.external_labels.
       # NOTE: This does not affect the external_labels set in prometheus config.
       clusterIDConfigmap: kubecost-cluster-info-configmap
+      image:
+        tag: v2.19.2
+      extraFlags:
+      - web.enable-admin-api
+      - web.enable-lifecycle
+      - storage.tsdb.wal-compression
       resources:
         limits:
           cpu: 1000m
@@ -104,9 +110,10 @@ cost-analyzer:
         size: 32Gi
         enabled: true
       extraArgs:
+        log.level: info
+        log.format: json
         storage.tsdb.min-block-duration: 2h
         storage.tsdb.max-block-duration: 2h
-        storage.tsdb.retention: 2w
       # extraVolumes: # TODO
       # - name: object-store-volume
       #   secret:
@@ -152,6 +159,8 @@ cost-analyzer:
         #   mountPath: /etc/config
     alertmanager:
       enabled: true
+      image:
+        tag: v0.21.0
       resources:
         limits:
           cpu: 50m
--- a/stable/kubecost/Chart.yaml
+++ b/stable/kubecost/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.60.1
+appVersion: 1.61.2
 description: Kubecost
 name: kubecost
 home: https://github.com/mesosphere/charts
-version: 0.1.10
+version: 0.1.11
 maintainers:
   - name: gracedo


stable/kubecost/values.yaml
```diff
@@ -87,13 +87,19 @@ cost-analyzer:
       # This overrides the cluster_id set in prometheus.server.global.external_labels.
       # NOTE: This does not affect the external_labels set in prometheus config.
       clusterIDConfigmap: kubecost-cluster-info-configmap
+      image:
+        tag: v2.19.1
```
gracedo, line 91: I am just concerned with the version bump here since the kubecost team hasn't vetted their product with this version of prometheus so it's possible something may break if we upgrade the version. All the other changes lgtm

stable/kubecost/values.yaml
```diff
@@ -152,6 +159,8 @@ cost-analyzer:
         #   mountPath: /etc/config
     alertmanager:
       enabled: true
```
dwbrown2, line 161: You may consider disabling this + pushGateway by default if you aren't using in core integration. @gracedo as FYI

stable/kubecost/values.yaml
dwbrown2: Yeah, we have not tested against this version yet. Should be fine but can't say with a high degree of certainty until we run all of our tests against this. Is there a specific feature you need?

sebbrandt87: nothing should break as it is only a minor version change and nothing was mentioned in the release notes.
https://github.com/prometheus/prometheus/releases
enhancements, fixes
some related to memory consumption

stable/kubecost/values.yaml
```diff
@@ -87,13 +87,19 @@ cost-analyzer:
...
+      image:
+        tag: v2.19.1
+      extraFlags:
+      - web.enable-admin-api
+      - web.enable-lifecycle
+      - storage.tsdb.wal-compression
       resources:
         limits:
+          cpu: 2000m
+          memory: 5000Mi
```
hectorj2f, line 99: why do you set 5G ? can you justify this change ?

hectorj2f: Only Prometheus sets this limits 🤔 ... is this consuming that many resources ? I have been looking at the usage and it cannot be compared with the native prometheus used by the whole cluster

This shows how during the last 30d the memory did go beyond  1.4G and cpu beyond 0.1. I'd suggest yo lower a bit this values accordingly to a common usage. I think the initial values are good enough.
Screen Shot 2020-06-25 at 11 28 52 AM

```
cpu: 1000m
          memory: 2500Mi
```

sebbrandt87:
image

hectorj2f:
@sebbrandt87 can you share more details ? It is a bit hard to discuss anything when you don't explain the source of an image or how you got these numbers.

hectorj2f:
@sebbrandt87 I have been checked the values of all the kubecost-prometheus-server pods which restarted due to the oom killer. 

Let's identify two things:
1) If you set these values they would be applied to any  **default** cluster while you are defining these based on the SOAK cluster. So let's distinguish default cluster and our soak cluster because we cannot adopt the soak settings (as defined for prometheus elasticsearc...) for default clusters. If you look over time the numbers of the cpu consumption, the average is quite low so setting 2 is too much in my opinion. I'll keep cpu: 1000m. Regarding the memory, it is a bit confusing because the value is only going up to 4/5GiB at specific times in our SOAK cluster. I'd prefer if we set a half way value because the soak settings are the same than a default cluster (sry for repeating myself again). So I'd say let's set a limit to memory of 3GiB as defined by default for Prometheus.

2) I saw our SOAK settings are using the defaults, let's set the value 5GiB and 1.5cpu for the konvoy-soak settings. We also customized other addons for the soak cluster such as prometheus, elasticsearch,... Let's set specific values for kubecost prometheus server there.

sebbrandt87:
- with wal-compression enabled, it will take more cpu and it is just a throttling limit.
- with a lot of data in a cluster and a oom killed pod it basically will happen that the pod never restarts fully again, because it is not able to load all data into memory.
- I defined them based on our long running konvoy-staging cluster. which basically got kube-cost enabled a few days ago.

hectorj2f:
@sebbrandt87 We don't use the soak resource consumption to define the defaults resource allocation for all our addons on any cluster.

stable/kubecost/values.yaml
geekbass: We are running version 2.19.0 in soak as of yesterday to clear up a query error.

hectorj2f:
@sebbrandt87 As another example, prometheus limits on soak point to set 14GiB but we don't set those for any cluster due to obvious reasons.

stable/kubecost/values.yaml
gracedo: We decided to enable alertmanager by default so we can define some rules and alerts which we felt were important for monitoring kubecost. Pushgateway is off by default

shaneutt:
@sebbrandt87 please let us know in @mesosphere/sig-ksphere-catalog if we can provide any assistance to keep you unblocked here :vulcan_salute:

sebbrandt87:
> @sebbrandt87 please let us know in @mesosphere/sig-ksphere-catalog if we can provide any assistance to keep you unblocked here 🖖

will come back to this soon 😬

sebbrandt87:
So this got a little bit out of my focus. Sorry!

I am fine with setting the defaults for resources back to where it was before, but want to say, this issue of to many resources being utilized at pod restart as of eviction or draining a node will also hit customers.

Updating this chart again now.

samba:
Historically we've set defaults based on fairly small clusters (i.e. default demo size) and instructed customers that they can customize the scale & resource allocation according to their cluster configuration. It has been a customers' responsibility to scale up & allocate accordingly -- not ours.

For Prometheus (and others, like Elasticsearch) we will need to perform a more exhaustive study of the utilization patterns we see across all the clusters we have been maintaining and operating, to make some educated guesses for default configurations, based on a *default target cluster topology*. 

Further, there have been some discussions in the past to _let addons scale themselves_, however this requires more sophisticated automation, in particular the use of _operators_ in some cases where none exist today.

For now, we should retain existing defaults (or upstream defaults), and all changes should be raised for discussion with product owners accordingly (me and @GoelDeepak) before merging into charts or addon repos.

PR #675, opened 2020-07-01
kommander: Update kubecost to fix grafana datasource bug

gracedo:
**What type of PR is this?**

Bug

Updates kubecost to latest version, which pulls in a fix for bug where the kubecost grafana dashboards were using `default` datasource which was pointing to kommander thanos query ds - it should use the `default-kubecost` ds (kubecost thanos query).

https://jira.d2iq.com/browse/D2IQ-69513

Tested by hosting chart and deploying - verified that the correct datasources are loaded to grafana, and the kubecost dashboards are correctly using the `default-kubecost` ds.
image
image

```release-note
NONE
```

**Checklist**


stable/kommander/values.yaml
```diff
@@ -238,7 +238,6 @@ kubecost:
         datasources:
           enabled: true
           defaultDatasourceEnabled: false
-          # dataSourceName: KubecostPrometheus
```
shaneutt, line 241: just a leftover eh?

gracedo: hmm.. actually I'm gonna add this back (uncomment it). originally this field wasn't supported/respected but it is now and `KubecostPrometheus` is more descriptive than the default `Prometheus`

shaneutt: :+1:

PR #676, opened 2020-07-01
Enable wal compression as a default and update Prometheus version

geekbass:
- enabling wal compression by default per D2IQ-69798
- updating Prometheus to latest version per D2IQ-69855 
- bumping the Chart version 

**What type of PR is this?** Chore

**What this PR does/ why we need it**: Enable wal compression based on findings in soak and update Prometheus version to latest for some bug fixes.

**Which issue(s) this PR fixes**:[ D2IQ-69798](https://jira.d2iq.com/browse/D2IQ-69798) and[ D2IQ-69855](https://jira.d2iq.com/browse/D2IQ-69855). 

This is based on current AWS soak findings and what is currently running there.

```release-note
NONE
```

**Checklist**

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.13.16
+version: 8.13.17
 appVersion: 0.38.1
 tillerVersion: ">=2.12.0"
 home: https://github.com/coreos/prometheus-operator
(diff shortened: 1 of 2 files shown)


geekbass:
> I'd prefer to set these one level up in kba, can we do that there so we're not modifying the base chart? (it'll make upgrading this chart against upstream easier as well since this is a fork)

branden:
If the new chart values are all that we need and those are going into KBA then I don't think we need this PR any longer. Looks like it's now a chart bump with no associated changes.

geekbass:
> If the new chart values are all that we need and those are going into KBA then I don't think we need this PR any longer. Looks like it's now a chart bump with no associated changes.

Yes true sorry. Lets close.

PR #677, opened 2020-07-02
fix: removes a left behind certificate upon istio addon deletion

GoelDeepak:
**What type of PR is this?**
Bug

It adds a certificate deletion job so that future installation doesn't error out

https://jira.d2iq.com/browse/D2IQ-69875

Verified by running manual test
```
1. helm install ./git/staging/istio --namespace istio-system -n istio-kubeaddons
2. helm delete --purge istio-kubeaddons
3. Verified that certificate is not left behind
```

```release-note

```

**Checklist**

### DIFF
--- a/staging/istio/charts/security/templates/cert-delete-job.yaml
+++ b/staging/istio/charts/security/templates/cert-delete-job.yaml
@@ -2,38 +2,38 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: istio-delete-job
+  name: cert-delete-job
   namespace: {{ .Release.Namespace }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: istio-delete-job
+  name: cert-delete-job
   namespace: {{ .Release.Namespace }}
 rules:
-- apiGroups: ["install.istio.io"]
-  resources: ["istiooperators"]
+- apiGroups: ["certmanager.k8s.io"]
+  resources: ["certificates"]
   verbs: ["get", "list", "watch", "delete"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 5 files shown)


staging/istio/charts/security/templates/cert-delete-job.yaml
```diff
@@ -0,0 +1,51 @@
...
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false" 
+    spec:
+      serviceAccountName: cert-delete-job
+      containers:
+      - name: istio-ca-delete
+        image: docker.io/bitnami/kubectl:1.16.2
+        command: ["kubectl",  "delete", "certificate", "istio-ca", "-n", "{{ .Release.Namespace }}"]
```
shaneutt, line 50: Are there situations where this certificate would NOT be present and would block deletion due to this command failing? Should we add a check here to succeed when the cert is not present, regardless of whether this job had to delete it?

GoelDeepak: Doesn't hurt to add. So just added. PTAL

staging/istio/charts/security/templates/cert-delete-job.yaml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      serviceAccountName: cert-delete-job
+      containers:
+      - name: istio-ca-delete
+        image: docker.io/bitnami/kubectl:1.16.2
```
sreis, line 49: Consider making this version and or image configurable.

GoelDeepak: Done

PR #682, opened 2020-07-03
[prometheus-operator] Fix Grafana flaky home dashboard

branden:
**What type of PR is this?**

Bug

This fixes an issue that can cause Grafana to come up without the home dashboard set. A post-install hook job may end up in a race with the Grafana sidecar that loads dashboards from configmaps. This PR replaces the hook with a cronjob that periodically sets the home dashboard. This should be replaced with setting the home dashboard in grafana.ini once that feature is available.

https://jira.d2iq.com/browse/D2IQ-69497

I tested this PR by deploying a cluster with the KBA `configRepository` set to `https://github.com/branden/kubernetes-base-addons` and the `configVersion` set to `grafana-dashboard-debug`. This points to a prometheus addon that uses the chart from this PR. I repeatedly disabled and enabled prometheus and observed that Grafana's home dashboard was always set.

```release-note
[prometheus-operator] Fixes an issue that may prevent Grafana from having a default home dashboard.
```

**Checklist**


staging/prometheus-operator/Chart.yaml
```diff
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 8.13.16
```
alejandroEsc, line 15: not sure i would say this is a patch version anymore based on the changes to the values file. However since we are probably keeping in step with upstream I wont block.

PR #683, opened 2020-07-03
[kommander] Fix Grafana flaky home dashboard

branden:
**What type of PR is this?**

Bug

This fixes an issue that can cause Grafana to come up without the home dashboard set. A post-install hook job may end up in a race with the Grafana sidecar that loads dashboards from configmaps. This PR replaces the hook with a cronjob that periodically sets the home dashboard. This should be replaced with setting the home dashboard in `grafana.ini` once that feature is available.

https://jira.d2iq.com/browse/D2IQ-69497

I tested this PR by deploying a cluster with the kommander addon's `configRepository` set to `https://github.com/branden/kubeaddons-kommander` and the `configVersion` set to `grafana-dashboard-debug`. This points to a kommander addon that uses the chart from this PR. I repeatedly enabled and disabled kommander and observed that Grafana's home dashboard was always set.

```release-note
\[kommander\] Fixes an issue that may prevent Grafana from having a default home dashboard.
```

**Checklist**


branden:
@gracedo Yeah let's have @alejandroEsc weigh in then. If that's the intended behavior then I can use this PR to add comments that make that clear. However if that's the case, then we probably shouldn't be treating this bug as a release blocker.

Either way, we need to be able to investigate this job when it fails. Exiting with an error to prevent cleanup is one approach, maybe instead we could update the hook delete policy to remove `on-success`?

alejandroEsc:
So i would first suggest that if things error out we print to stdout a comment that a failure occurred and maybe details, that would help with the debugging part of this. Under what circumstances are we seeing this and is it ephemeral? Could we mitigate this with retry logic? or maybe have the job retry when it fails.

branden:
One of the problems I'm trying to address is that it's difficult to get the job's logs, since it's deleted when it succeeds. (This is an issue due to the separate problem that the logs aren't making it to elasticsearch.) I've been debugging with `on-success` removed from the delete policy, and that's allowed me to make progress.

I changed the exit codes with the impression that an error was getting suppressed, but that turns out to not be the case. See the ticket in the description for more context on the bug that I'm chasing down. However I'm not totally convinced that the script should succeed when it fails to set the dashboard. Failing the entire kommander deployment is drastic but I think I prefer it over silently failing to set the home dashboard. Retrying the job when it fails sounds reasonable to me.

alejandroEsc:
> One of the problems I'm trying to address is that it's difficult to get the job's logs, since it's deleted when it succeeds. (This is an issue due to the separate problem that the logs aren't making it to elasticsearch.) I've been debugging with `on-success` removed from the delete policy, and that's allowed me to make progress.
> 
> I changed the exit codes with the impression that an error was getting suppressed, but that turns out to not be the case. See the ticket in the description for more context on the bug that I'm chasing down. However I'm not totally convinced that the script should succeed when it fails to set the dashboard. Failing the entire kommander deployment is drastic but I think I prefer it over silently failing to set the home dashboard. Retrying the job when it fails sounds reasonable to me.

I thought about this some time ago, do we not see something in tiller?

alejandroEsc:
@branden @gracedo then let's switch to retry 4 times with some sleep time, assuming that the issue is just that a timing one, of 30s (so maybe delay things by 2 minutes), exit and let's not delete the job due to failure (only on success) and then exit 1 if error occurs. Keep in mind we may need to "rescue" or redeploy the addon if failure does occur.

samvantran:
We definitely should not fail an entire deployment of Konvoy (+ Kommander) because the home dashboard isn't set. That is a very bad user experience. With no home dashboard, a user can still find dashboards and use other apps/services. With no deployment, they can't do anything and will think our product is faulty. We should handle failures gracefully.

I agree with multiple retries and printing error messages that are easy to find. If the last retry fails, we should still continue the rest of deployment.

branden:
@alejandroEsc 
> I thought about this some time ago, do we not see something in tiller?
Tiller logs that the hook gets created, cleaned up, etc. It's something but it's not enough to know what the job is actually doing.

alejandroEsc:
> We definitely should not fail an entire deployment of Konvoy (+ Kommander) because the home dashboard isn't set. That is a very bad user experience. With no home dashboard, a user can still find dashboards and use other apps/services. With no deployment, they can't do anything and will think our product is faulty. We should handle failures gracefully.
> 
> I agree with multiple retries and printing error messages that are easy to find. If the last retry fails, we should still continue the rest of deployment.

Yeah i agree, but running up again should be fine if need be, I guess we should test more. In the case were we do decide to fail silently, which again is ok and doesnt mean we shouldnt retry either way, we should update the runbook to reflect how to do this manually. Ill go ahead and do that soon.

PR #707, opened 2020-07-20
fix[cert-manager-setup]: deploy Issuer CRs in the release namespace

dlipovetsky:
**What type of PR is this?**
Bug

Prior to this fix, the namespace is set to an empty string. When the namespace
is an empty string, helm v2 deploys the Issuer CR in the release namespace, but
helm v3 deploys it in namespace where the the kubeaddons-controller runs.
That's a problem, because cert-manager looks for the Issuers in its own
namespace.

https://jira.d2iq.com/browse/D2IQ-70238

```release-note
NONE
```

**Checklist**


alejandroEsc:
> lgtm - can we add a test to verify its deployed in the correct ns or do we not have the framework for v3 testing yet?

I tested this locally, it works, adding tests are a bit more difficult and probably belongs to another PR.

dlipovetsky:
> lgtm - can we add a test to verify its deployed in the correct ns or do we not have the framework for v3 testing yet?

+1. Looking into this.

dlipovetsky:
Update on tests:

The root cause of this problem is a field that is required (i.e. assumed to be defined), but is in fact undefined.  To that end, I'm adding an assertion: the `required` function, supported in both [helm v3](https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions) and [helm v2](https://v2.helm.sh/docs/charts_tips_and_tricks/#using-the-required-function).

I think that an assertion, not a test, is the best way to ensure that required fields are not undefined. However, it might be good to have a test that verifies that resources are created in the namespaces where the application (in this case cert-manager) expects them.

samvantran:
Thanks @dlipovetsky that sounds good to me

dlipovetsky:
The namespace field now relies directly on the [helm built-in](https://helm.sh/docs/chart_template_guide/builtin_objects/) `.Release.Namespace`, which helm (apparently) always defines. For that reason, I think the assertion is not needed.

PR #708, opened 2020-07-21
fix[cert-manager-setup]: deploy Certificate CRs in the release namespace

dlipovetsky:
**What type of PR is this?**
Bug

The certificates are not created with the right namespace, because the certificate template does not define a namespace. Where helm v2 used the release namespace (correct), helm v3 uses the kubeaddons-controller namespace (incorrect).

https://jira.d2iq.com/browse/D2IQ-70273

```release-note
NONE
```

**Checklist**

### DIFF
--- a/staging/cert-manager-setup/Chart.yaml
+++ b/staging/cert-manager-setup/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.1.11
+version: 0.1.12
 appVersion: 0.10.1
 description: Install cert-manager and optionally add a ClusterIssuer
 keywords:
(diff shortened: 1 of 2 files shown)


sebbrandt87:
it may would be an option to also include right away the possibility to also define Certificates for other namespaces.
E.g. We may want to create a certificate right away by cert-manager-setup like in the kube-system namespace.

alejandroEsc:
> it may would be an option to also include right away the possibility to also define Certificates for other namespaces.
> E.g. We may want to create a certificate right away by cert-manager-setup like in the kube-system namespace.

yep this was discussed, for now though we want to unblock ourselves for helm v3 work.

sebbrandt87:
just for the record https://jira.d2iq.com/browse/D2IQ-65997
I may just create PR 😬

PR #717, opened 2020-07-28
feat(cert-manager): being able to define full issuer spec

sebbrandt87:
**What type of PR is this?**

Feature

We have to make it possible to define spec for issuers through values, instead of precreated
secretName only.

no issue

```release-note
- `Issuer` spec definable instead of only `ca.secretName`
```

**Checklist**


sebbrandt87:
Out of my experience 😬 
As a customer who wants to go all in for Kubernetes, I want to manage everything there and do not want to run extra script or so external and import my results later.

alejandroEsc:
@jr0d could you add more to the discussion and though i see your are ok with the changes are we are locking ourselves to something we may not want? 

@sebbrandt87 I reran the test, could you check if it fails again, if so could you investigate.

sebbrandt87:
We need it to be able to define extra certificates with kubeaddons. Instead of waiting for kubeaddons to be done and manage certs and issuers with extra manifests applied

sebbrandt87:
Found the issue. Was a stupid test value I had set. `Duration` should be bigger then `RenewBefore` ;)

jr0d:
@alejandroEsc The example shows how to create a self signed, end entity certificate using a null issuer. That's what I was referencing in my comment. However, this PR will also allow the user to configure arbitrary issuers through the chart, rather than creating issuers manually, as is our current process: https://docs.d2iq.com/ksphere/konvoy/1.5/security/letsencrypt/

shaneutt:
This still says `wip`, but is this `ready`? :thinking:

alejandroEsc:
Is this PR ready?

shaneutt:
@sebbrandt87 just a heads up your PR has become a guinea pig for me testing helmv3 CI fixes :smiling_imp:

shaneutt:
@sebbrandt87 I'm investigating the CI issues seen here and seeing if I can help.

shaneutt:
@sebbrandt87 have you seen [D2IQ-71381](https://jira.d2iq.com/browse/D2IQ-71381) which would appear to be relevant to your CI issues seen here? cc @dkoshkin

sebbrandt87:
> @sebbrandt87 have you seen [D2IQ-71381](https://jira.d2iq.com/browse/D2IQ-71381) which would appear to be relevant to your CI issues seen here? cc @dkoshkin

nope not seen, as I am focused on other topics at the moment.

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: APIService "v1beta1.webhook.certmanager.k8s.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "cert-manager-setup-oynub9rkez": current value is "cert-manager"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "cert-manager-setup-oynub9rkez": current value is "cert-manager"
Upgrade testing for release 'cert-manager-setup-oynub9rkez' skipped because of previous revision installation error: Error waiting for process: exit status 1
```

I don't see that it is related but instead the the namespace seems to be wrong and ownership?

sebbrandt87:
> Since the upgrade to cert-manager is imminent, if prefer we not do this at this time. Let's see if we can use the upstream chart in the upgrade, first, and if we can't, we should track patches to the upstream chart in the same way as our prometheus-operator fork.

This isn't a upstream chart change.
We just change how we are able to define in our chart, which has a dependency to the cert-manager chart, how to define issuers.

makkes:
@sebbrandt87 can this PR be closed? It looks pretty old.

makkes:
closing now for lack of activity. Please feel free to re-open and refresh it.

PR #721, opened 2020-08-05
[velero] fix: allow velero to deploy on helm2&3 without needing install hooks

alejandroEsc:
**What type of PR is this?**

BUG

https://jira.d2iq.com/browse/D2IQ-70289

```release-note
NONE
```

**Checklist**

### DIFF
--- a/staging/velero/Chart.yaml
+++ b/staging/velero/Chart.yaml
@@ -20,4 +20,4 @@ name: velero
 sources:
 - https://github.com/heptio/velero
 tillerVersion: '>=2.10.0'
-version: 3.0.5
+version: 3.0.6
--- a/staging/velero/crds/crds.yaml
+++ b/staging/velero/crds/crds.yaml
@@ -1,0 +1,132 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: backups.velero.io
+spec:
+  group: velero.io
+  version: v1
+  scope: Namespaced
+  names:
+    plural: backups
+    kind: Backup
+---
+apiVersion: apiextensions.k8s.io/v1beta1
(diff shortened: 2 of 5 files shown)


samvantran:
Should we run an AWS e2e test against Konvoy master so that we know this will work on Konvoy? Want to make sure we don't get caught again. (if the test fails with ubi8, someone is already looking into that one).

alejandroEsc:
> Should we run an AWS e2e test against Konvoy master so that we know this will work on Konvoy? Want to make sure we don't get caught again. (if the test fails with ubi8, someone is already looking into that one).

Suggestions about which test and where? Do you mean a manual test?

dlipovetsky:
Thanks for fixing this @alejandroEsc!

Could you please amend the README (or document in a separate file) how CRD management works? There are CRDs in the `crds/` folder, there are CRDs in `files/crds.yaml`, and of course, there's the `templates/install-crds.yaml` What gets used when, I still don't have straight, unfortunately, and I would appreciate having a reference.

alejandroEsc:
> Thanks for fixing this @alejandroEsc!
> 
> Could you please amend the README (or document in a separate file) how CRD management works? There are CRDs in the `crds/` folder, there are CRDs in `files/crds.yaml`, and of course, there's the `templates/install-crds.yaml` What gets used when, I still don't have straight, unfortunately, and I would appreciate having a reference.

Sounds like a good idea. I will do so in another PR.

alejandroEsc:
@samvantran the aws test passed https://github.com/mesosphere/kubernetes-base-addons/pull/423, however since this is a chart that needs pv, i suspect this is why the ci test is failing

samvantran:
> however since this is a chart that needs pv, i suspect this is why the ci test is failing

another area to improve in testing. can we fix that in the testing framework?

PR #733, opened 2020-08-12
[testing] additional changes to run tests separately, fix helm args

alejandroEsc:
**What type of PR is this?**

BUG

Splits testing of charts to helm2 and helm3, also fixes bug where tiimeout for helm3 requires units. 

Changes how we test a bit, and uses the right args for helm3

```release-note
NONE
```

**Checklist**


test/helm3/ct.yaml
```diff
@@ -0,0 +1,20 @@
+debug: true
+target-branch: master
+chart-dirs:
+  - stable
+  - staging
+excluded-charts:
+  - common
+  - dex-controller # Moved to a different helm repo
```
samvantran, line 8: All [of these charts](https://github.com/mesosphere/charts/pull/733/files#diff-a139d5b87837ab46aab5151c40082121R9-R28) failed to install with ct for helm2. Is this no longer the case with ct helm3?

alejandroEsc:
> 👍 on the separation but helm3 test failing?

dummylb deployment failing, my suspicion is collision of ips, looking into it.

test/helm3/ct.yaml
alejandroEsc: I would have to test these 1 by 1. The reasons for the lack of testing an be complicated, and there are jiras on a few of these.

test/helm3/ct.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - mesosphere-stable=https://mesosphere.github.io/charts/stable
+  - kubefed-charts=https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
+  - jetstack-charts=https://charts.jetstack.io
+  - banzaicloud=https://kubernetes-charts.banzaicloud.com
+  - helm-stable=https://kubernetes-charts.storage.googleapis.com/
+  - dex-controller=https://mesosphere.github.io/dex-controller/charts
+  - kommander=https://mesosphere.github.io/kommander/charts
+  - kubecost=https://kubecost.github.io/cost-analyzer
+  - yakcl=https://mesosphere.github.io/yakcl/charts
+helm-extra-args: --timeout 600s
```
shaneutt, line 20: Is 600s a bit high? Do we know what the longest average run is?

alejandroEsc: default is 5ms, but yeah i have seen a few things run for around 10m in helm2 land.

shaneutt: ok :+1:

PR #742, opened 2020-08-17
[prometheus-operator] feat: add scripts to make upgrading easier

samvantran:
**What type of PR is this?**
Feature

Upgrading prometheus has been problematic because we have a bunch of patches we apply on top of the upstream chart. Each time has been manual and sometimes we've missed something during an upgrade.

This PR adds all the mesosphere-specific files & changes (templates, hooks, etc) to the `/patches` folder. There are also scripts that apply those patches as a set. This should make it easy to follow which files are added/amended after copying the upstream chart. This is somewhat akin to how DB migrations work in Rails (this was years ago though, things may have changed).

Now we can run `./update_operator.sh` and it should take care of all the rest.

Here are the files that are mesophere-specific:
```
 > tree patch/mesosphere
patch/mesosphere
└── templates
    ├── grafana
    │   ├── cronjob-home-dashboard.yaml
    │   └── dashboards
    │       ├── autoscalerdashboard.yaml
    │       ├── calicodashboard.yaml
    │       ├── elasticsearchdashboard.yaml
    │       ├── fluentbitdashboard.yaml
    │       ├── grafanadashboard.yaml
    │       ├── kibanadashboard.yaml
    │       ├── localvolumeprovisioner.yaml
    │       ├── opsportaldashboard.yaml
    │       ├── traefikdashboard.yaml
    │       └── velerodashboard.yaml
    ├── hooks
    │   └── prometheus-cluster-id-hooks.yaml
    ├── ingress-rbac
    │   ├── alertmanager.yaml
    │   ├── grafana.yaml
    │   └── prometheus.yaml
    └── rules
        ├── etcd-rules.yaml
        └── velero-rules.yaml

6 directories, 17 files
```

https://jira.d2iq.com/browse/D2IQ-67095

I mostly did this from the diff of copying the latest chart and trying to follow PR #619. It's possible I missed some files so definitely would appreciate anyone who's done a prometheus upgrade to make doubly sure I got everything.

```release-note
NONE
```

**Checklist**


alejandroEsc:
wow nice. Could we get a quick readme about how these scripts are run?

samvantran:
> wow nice. Could we get a quick readme about how these scripts are run?

oh yeah for sure. I'll add in comments to the scripts as well so it's easier to understand

> Amazing 🙌 the only other thing i can think of is this change here that we often forget to apply:

@gracedo i knew there was something missing! I thought it was [this fix](https://github.com/mesosphere/charts/pull/742/files#diff-2acfca100d72548077f993b6ac757998R26) but yes, there's more. Okay I'll add a patch for that. Thanks for catching 🙏

PR #744, opened 2020-08-18
test: Add retries to e2e test curls

gracedo:
**What type of PR is this?**

Bug

Adds retries to the curl requests in the e2e test to avoid failing the test suite in case of connectivity or other transient issues.
```
--max-time 10 # each retry waits up to 10s
--retry 10 # retry 10 times
--retry-max-time 60 # retry for total 60s before failing
--retry-connrefused # retry on connection refused errors
```

https://jira.d2iq.com/browse/D2IQ-70518

```release-note
NONE
```

**Checklist**

### DIFF
--- a/test/e2e-kind.sh
+++ b/test/e2e-kind.sh
@@ -13,6 +13,7 @@ readonly CLUSTER_NAME=chart-testing
 CT_VERSION=$1
 HELM_VERSION=$2
 GIT_REMOTE_NAME=${GIT_REMOTE_NAME:=origin}
+CURL_RETRY_OPTS="--max-time 10 --retry 10 --retry-max-time 60 --retry-connrefused"
 
 tmp=$(mktemp -d)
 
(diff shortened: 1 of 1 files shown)


test/e2e-kind.sh
```diff
@@ -13,6 +13,7 @@ readonly CLUSTER_NAME=chart-testing
 CT_VERSION=$1
 HELM_VERSION=$2
 GIT_REMOTE_NAME=${GIT_REMOTE_NAME:=origin}
+CURL_RETRY_OPTS="--max-time 30 --retry 20 --retry-connrefused"
```
armandgrillet, line 16: If there is no reason behind the numbers used here, I believe the values are too high. I suggest using `--retry-max-time` to set the total time before the entire request with retries is considered failed (maybe set to `60` so that we know a request can take 1 minute max?) with lower `--max-time` and `--retry` values.

gracedo: thanks for the tips! still playing around with the numbers, I'll lower `--max-time` and `--retry` and set `--retry-max-time`

armandgrillet (approved):

PR #746, opened 2020-08-19
TESTING: do not merge test

alejandroEsc:
**What type of PR is this?**

**What this PR does/ why we need it**:

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #749, opened 2020-08-19
fix: lint requires ct.yaml to be in test dir

alejandroEsc:
**What type of PR is this?**

Lint is broken because we need to have the ct file in a particular location, reverting back to 3.0.0 but using the right ct file.

```release-note
NONE
```

**Checklist**


alejandroEsc:
https://github.com/mesosphere/charts/pull/746 failed as expected, this fixes the linting problem

PR #756, opened 2020-08-20
[kommander] helm3 changes

alejandroEsc:
**What type of PR is this?**

Chore

**What this PR does/ why we need it**:

This moves kommander into helm3, this is a breaking change hence the version

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
~~The helm2 test will and should fail intentionally. ~~ seems like we skip these tests. 

**Does this PR introduce a user-facing change?**:

```release-note
Kommander is now deployable by helm3 only
```

**Checklist**

* [x] *If a chart is changed, the chart version is correctly incremented.*
* [x] The commit message explains the changes and why are needed.
* [x] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #757, opened 2020-08-20
kommander: Remove hardcoded service account names

gracedo:
**What type of PR is this?**

Chore

@alejandroEsc mentioned that we should not be hardcoding the service account names for the templates defined in the same chart. The service account name is templatized in https://github.com/mesosphere/charts/blob/master/stable/kommander/templates/hooks-roles.yaml#L5 which is also how we should be pulling in the sa.

no issue

```release-note
NONE
```

**Checklist**


gracedo:
do not merge until we figure out why #756 failed to publish

PR #740, opened 2020-08-13
Publish with helm v2 by setting explicitly

faiq:
**What type of PR is this?**

**What this PR does/ why we need it**:

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #759, opened 2020-08-20
ensure helm is installed where it's executed

joejulian:
**What type of PR is this?**

bug

helm calls fail because the binary doesn't exist at $(HELM)

https://jira.d2iq.com/browse/D2IQ-71191

```release-note
NONE
```

**Checklist**

### DIFF
--- a/Dispatchfile
+++ b/Dispatchfile
@@ -26,7 +26,7 @@ dind_task("publish",
               export GIT_REMOTE_URL=https://mesosphere:${GITHUB_USER_TOKEN}@github.com/mesosphere/charts.git
               git config --global user.name 'Dispatch CI'
               git config --global user.email '56653984+d2iq-dispatch@users.noreply.github.com'
-              make publish HELM_VERSION=v2.16.9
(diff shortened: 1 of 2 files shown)


Dispatchfile
```diff
@@ -26,7 +26,7 @@ dind_task("publish",
               export GIT_REMOTE_URL=https://mesosphere:${GITHUB_USER_TOKEN}@github.com/mesosphere/charts.git
               git config --global user.name 'Dispatch CI'
               git config --global user.email '56653984+d2iq-dispatch@users.noreply.github.com'
-              make publish HELM_VERSION=v2.16.9
+              make publish
```
samvantran, line 29: @faiq you set this explicitly before https://github.com/mesosphere/charts/pull/740

Will this change break anything in dispatch since we're trying to publish v3 charts now?

joejulian: This is just running "make publish" and specifying an old helm version. We don't want this set here, we want it set in the Makefile.

PR #763, opened 2020-08-24
ensure we're using the correct helm

joejulian:
**What type of PR is this?**

chore

If helm is in the path, the Makefile would use it - regardless of whether it was a version that would work. This checks for the correct version and errors if it's wrong.

nada

```release-note
none
```

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ else
 HELM_INSTALLED=$(shell $(HELM) version --short --client | sed 's/.*\(v[0-9]\+\.[0-9]\+\).*/\1/')
 HELM_REQUESTED=$(shell echo $(HELM_VERSION) | cut -d. -f-2)
 ifneq ($(HELM_INSTALLED),$(HELM_REQUESTED))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,7 +33,15 @@ export HELM_DATA_HOME=$(TMPDIR)/.helm/data
 HELM := $(shell command -v helm)
 ifeq ($(HELM),)
 	HELM := $(TMPDIR)/helm
+else
+# compare short versions to see if this is the requested version
+HELM_INSTALLED=$(shell $(HELM) version --short --client | sed 's/.*\(v[0-9]\+\.[0-9]\+\).*/\1/')
+HELM_REQUESTED=$(shell echo $(HELM_VERSION) | cut -d. -f-2)
+ifneq ($(HELM_INSTALLED),$(HELM_REQUESTED))
+$(error $(HELM) is not the requested version: $(HELM_VERSION))
```
gracedo, line 41: instead of erroring and exiting, can we instead defer to downloading and using the correct helm version?

hectorj2f (approved):

PR #768, opened 2020-08-27
Fixed the KubeHpaReplicasMismatch fault alert.

Gilbert88:
**What type of PR is this?**
Bug

KubeHpaReplicasMismatch alert is faulty on konvoy soak cluster. The alert is due to
kube_hpa_status_current_replicas is 1, while
kube_hpa_status_desired_replicas is 0

The Prometheus alert expects the knative autoscaler hpa pod scales back to zore, which conflict with the knative HPA config `spec.minReplicas: 1`.

The alert should be less strict as "currentReplicas >= desiredReplicas" instead.

https://jira.d2iq.com/browse/D2IQ-71382

No.

**Checklist**

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 9.3.1
+version: 9.3.2
 appVersion: 0.38.1
 tillerVersion: ">=2.12.0"
 home: https://github.com/coreos/prometheus-operator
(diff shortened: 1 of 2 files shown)


Gilbert88:
@gracedo would you mind carrying this over?

After the second thought, this PR does not seem correct to me. Because we want to test scale down, so "currentReplicas >= desiredReplicas" may not be a healthy status.

Instead, we would need to add a bit more logics to the alert to compare desiredReplicas with minReplicas/maxReplicas from the Spec.

shaneutt:
Made this `wip` based on the comments/context, feel free to change that if it's not accurate.

Gilbert88:
Closed this PR because it is not the correct fix. See https://jira.d2iq.com/browse/D2IQ-71382?focusedCommentId=348611&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-348611 for root cause

PR #771, opened 2020-08-31
feat: handle airgapped deployment

hectorj2f:
**What type of PR is this?**

Feature

Allow adding certain settings when having an airgapped deployment.

```release-note

```

**Checklist**


vespian (approved):

drewvanstone (approved):

hectorj2f:
I re-run the tests... ⭕

armandgrillet (approved):

hectorj2f:
@jimmidyson The federated addons are not all using `https://mesosphere.github.io/charts/stable`. This value does not fit to all, however the value when it is airgapped. It does.

jimmidyson:
Now I look at the PR again, I realised that airgapped should be applied to all or none of the federated addons, correct? If so, then I propose to move the value to `global.federated.airgapped` and use that in the subcharts. This would allow a single setting to apply to all.

stable/kommander/Chart.yaml
```diff
@@ -7,4 +7,4 @@ maintainers:
   - name: alejandroEsc
   - name: jimmidyson
 name: kommander
-version: 0.10.3
+version: 0.10.4
```
jimmidyson, line 10: Let's bump the minor version to reflect adding the airgapped capability:

```suggestion
version: 0.11.0
```

jimmidyson (approved):
LGTM but please bump version to 0.11.0.

hectorj2f:
We need to bump yakcl with a bugfix, then we could merge this PR.

PR #773, opened 2020-09-01
replace deployment when upgrading

joejulian:

The upstream cert-manager chart uses a dynamically generated label in
the selector. The selector is immutable so this fails when upgrading
from tiller to helm v3.

By forking and changing just the deployment name, the deployment will be
recreated using the new selector.

Since the goal is to retire the 0.10 version of cert-manager from our
products in the near future, this should not create an
undue maintenance burden.

https://jira.d2iq.com/browse/D2IQ-70812

```release-note
none
```

**Checklist**


dkoshkin:
@joejulian can you split up the forking of cert-manager and the changes applied to it into 2 separate commits please.

joejulian:
> @joejulian can you split up the forking of cert-manager and the changes applied to it into 2 separate commits please.

Good point, done.

alejandroEsc:
shoot, i think we have to include crd management in this PR similar to velero.

joejulian:
This will not pass e2e tests due to https://github.com/mesosphere/charts/blob/master/test/e2e-kind.sh#L284 installing cert-manager.

I have performed these tests, locally, commenting out that line and the tests pass.

shaneutt (approved):
We should take extra special care that we've manually tested this upgrade a few times (I think we already have I'm just being extra cautious) and then :heavy_check_mark: for merge

PR #774, opened 2020-09-02
kubecost: Upgrade kubecost dep to 1.64.0

gracedo:
**What type of PR is this?**

Chore

**What this PR does/ why we need it**:

Upgrade kubecost dep to 1.64.0 to stay up to date

**Which issue(s) this PR fixes**:

no issue

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Checklist**

* [x] *If a chart is changed, the chart version is correctly incremented.*
* [x] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #775, opened 2020-09-02
Exposing 'registriesSkippingTagResolving' parameter for KNative Serving

akirillov:
**What type of PR is this?**
Feature

KNative Serving resolves image tags to their digests when service deployment is created. This doesn't work for air-gapped installations and we need an option to exclude registries for which it doesn't work.

[D2IQ-71478: pods launched by kfserving use image SHA instead of tags](https://jira.d2iq.com/browse/D2IQ-71478)

N/A

NONE

**Checklist**

### DIFF
--- a/staging/knative/charts/serving/values.yaml
+++ b/staging/knative/charts/serving/values.yaml
@@ -7,3 +7,6 @@ customMetricsApiservice:
 
 namespaceKnativeServing:
   additionalLabels: {}
+
+deployment:
+  registriesSkippingTagResolving: ""
--- a/staging/knative/charts/serving/values.yaml
+++ b/staging/knative/charts/serving/values.yaml
@@ -8,5 +8,5 @@ customMetricsApiservice:
 namespaceKnativeServing:
   additionalLabels: {}
 
-deployment:
+configDeployment:
   registriesSkippingTagResolving: ""
--- a/staging/knative/values.yaml
+++ b/staging/knative/values.yaml
@@ -14,3 +14,5 @@ serving:
     enabled: true
   namespaceKnativeServing:
     additionalLabels: {}
+  deployment:
+    registriesSkippingTagResolving: ""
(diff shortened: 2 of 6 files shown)


akirillov:
soft ping @branden, @gracedo. I've tested the installation locally with Helm 2 and verified it is working. Is CI failure related to the cert manager CRD issue?

gracedo:
@akirillov yeah looks like it. From #774 yesterday:
> @alejandroEsc:
Helm2 testing is broken atm since cert-manager is using a crd that is apiextensions/v1 and not v1beta1 which helm2 currently only supports. This will not change since this helm2 bug fixes are not accepted any longer. If we are in a rush to merge this PR in, I am all in since I saw this be green before the cert-manager change.

staging/knative/charts/serving/values.yaml
```diff
@@ -7,3 +7,6 @@ customMetricsApiservice:
 
 namespaceKnativeServing:
   additionalLabels: {}
+
+deployment:
```
Gilbert88, line 11: There multiple deployment kind in knativer serving. To be explicit, let's rename it as `configDeployment` or `deploymentConfigMap`?

staging/knative/values.yaml
```diff
@@ -14,3 +14,5 @@ serving:
     enabled: true
   namespaceKnativeServing:
     additionalLabels: {}
+  deployment:
```
Gilbert88, line 17: ditto

Gilbert88 (approved):
LGTM! Thanks @akirillov

staging/knative/charts/serving/values.yaml
akirillov: renamed to `configDeployment`

staging/knative/values.yaml
akirillov: updated

joejulian (approved):

PR #778, opened 2020-09-03
fix: usage of kubectl:latest

hectorj2f:
**What type of PR is this?**

Bug

Generally, we should NOT use `latest` tag due to multiple reason. But today, the `latest` tag is breaking the Konvoy airgapped bundle generation due to the lack of repository tags when pulling it. Latest tag is 1.18.8 but 1.18.4 doesn't have that problem and should work as well.

```release-note
none
```

**Checklist**

### DIFF
--- a/staging/cert-manager-setup/Chart.yaml
+++ b/staging/cert-manager-setup/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.1.15
+version: 0.1.16
 appVersion: 0.10.1
 description: Install cert-manager and optionally add a ClusterIssuer
 keywords:
--- a/staging/cert-manager-setup/templates/install-crds.yaml
+++ b/staging/cert-manager-setup/templates/install-crds.yaml
@@ -31,7 +31,7 @@ spec:
       serviceAccountName: cert-manager-setup-crds
       containers:
         - name: cert-manager-setup-crds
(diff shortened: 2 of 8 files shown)


drewvanstone (approved):

dkoshkin (changes requested):
Need to also change https://github.com/mesosphere/charts/blob/cfd38f1dd00ea5d07a75bc43c4608b5237f98288/staging/cert-manager-setup/templates/install-crds.yaml#L34

Also lets use `v1.18.8`.
```
➜  docker pull bitnami/kubectl:1.18.8
1.18.8: Pulling from bitnami/kubectl
522a7372a4ae: Pull complete
f4b1c0affcc3: Pull complete
6b9ccf0a32fb: Pull complete
af4e419d0d56: Pull complete
0a4ddf3da8ae: Pull complete
945c62472926: Pull complete
Digest: sha256:520ed8f8ee4f6e6615dfd6553727e0548185dfcdbdb31fe8a06bc127c653875c
Status: Downloaded newer image for bitnami/kubectl:1.18.8
docker.io/bitnami/kubectl:1.18.8
```

tillt (approved):

gracedo (approved):

joejulian:
This is the same CI failure that I encountered in #773 which is caused by installing cert-manager before testing cert-manager - causing cert-manager to be installed twice and preventing the re-installation of cluster-scoped resources (webhooks).

Again, manual testing without installing the duplicate cert-manager passes.

dkoshkin (approved):

PR #788, opened 2020-09-14
cert-manager v1.0.1

jr0d:
**What type of PR is this?**

Chore

* updates cert-manager chart
* removes  CRD install job and files
* updates resources to certmanager.io/v1

https://jira.d2iq.com/browse/D2IQ-70659

```release-note
[cert-manager-setup] cert-manager has been updated to v1.0.1
```

**Checklist**


staging/cert-manager-setup/values.yaml
```diff
@@ -82,3 +82,7 @@ clusterissuer: {}
 # This will work only in case when a cert-manager is installed in
 # `cert-manager` namespace.
 installCertManager: true
+
+cert-manager:
+  # This wrapper chart no longer installs CRDs, instead we rely on the upstream chart.
```
joejulian, line 87: Does upstream finally offer a way to reliably ensure the CRD is installed and updated to the current version? Or are we just expecting that v1 is mature enough that it's just not going to change?

joejulian (approved):

staging/cert-manager-setup/values.yaml
jr0d: @joejulian I need to check (I hope so). The reason I made the change was that the CRDs no longer fit in a configmap.

```
 Error: UPGRADE FAILED: pre-upgrade hooks failed: warning: Hook pre-upgrade cert-manager-setup/templates/install-crds.yaml failed: ConfigMap "cert-manager-setup-crds" is invalid: []: Too long: must have at most 1048576 bytes
```

```
  File: /home/jared/git/helm/d2iq/charts/staging/cert-manager-setup/files/crds.yaml
  Size: 1679688         Blocks: 3288       IO Block: 4096   regular file
Device: 823h/2083d      Inode: 30045113    Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/   jared)   Gid: ( 1000/   jared)
Access: 2020-07-20 10:41:17.459799075 -0500
Modify: 2020-09-10 08:16:06.290054123 -0500
Change: 2020-09-10 08:16:06.290054123 -0500
 Birth: 2020-07-20 10:41:17.459799075 -0500
```

staging/cert-manager-setup/templates/clusterissuer.yaml
```diff
@@ -1,6 +1,6 @@
 # DEPRECATED, this file should be deleted soon
```
sebbrandt87, line 1: Shouldn't we now finally delete this file?

staging/cert-manager-setup/values.yaml
sebbrandt87: I don't think upstream does now.
They seam to release with Bazel a version which installs CRDs, if `installCRDs` is true.
But haven't seen an upgrade strategy so far.

We may should split our CRDs files as upstream does.
I did this once here https://github.com/mesosphere/charts/commit/2c7263efe91457ffc6ac6cb6df19d16cdda609fb.
https://github.com/jetstack/cert-manager/tree/master/deploy/crds

sebbrandt87 (changes requested):

sebbrandt87:
Is this change really a `Chore` 😬 ?

joejulian:
> Is this change really a `Chore` :grimacing:  ?

Isn't everything? :wink:

staging/cert-manager-setup/templates/clusterissuer.yaml
jr0d: :+1:

sebbrandt87 (approved):
beside of the still existing deprecated file, I think this looks good now

jr0d:
helmv2 is failing with:
```
Error: unable to convert to CRD type: no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "k8s.io/client-go/kubernetes/scheme/register.go:67"
```

jr0d:
We (@jimmidyson) submitted a patch for this issue, but was rejected due to v2 EOL: https://github.com/helm/helm/issues/6783. This chart will never be deployed with v2, so I think we can bypass the v2 test and merge.

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -39,9 +39,6 @@ spec:
     - {{ . | quote -}}
   {{- end }}
 {{- end }}
-  keyAlgorithm: {{ .keyAlgorithm | default "rsa" }}
-  keyEncoding: {{ .keyEncoding | default "pkcs1" }}
-  keySize: {{ .keySize | default "2048" }}
```
sebbrandt87, line 44: Old definition was https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1alpha2.CertificateSpec

This must be kept now with new defintion of https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificatePrivateKey

sebbrandt87: I expect the tests must be changed for that as well.

sebbrandt87 (changes requested):
Please don't remove features being added but migrate.

staging/cert-manager-setup/ci/general-test-values.yaml
```diff
@@ -24,9 +24,6 @@ certificates:
       - www.example.com
     ipAddresses:
       - 127.0.0.1
-    keyAlgorithm: rsa
-    keyEncoding: pkcs8
-    keySize: 4096
```
sebbrandt87, line 29: must be kept for privateKey defintion as mentioned in change request

sebbrandt87 (changes requested):

staging/cert-manager-setup/templates/post-install-hook-job.yaml
```diff
@@ -20,4 +20,4 @@ spec:
       - name: {{ .Chart.Name }}
         image: bitnami/kubectl:1.18.8-debian-10-r15
         imagePullPolicy: IfNotPresent
-        command: ["kubectl", "wait", "--for=condition=Available", "--timeout=360s", "apiservice", "v1beta1.webhook.certmanager.k8s.io"]
+        command: ["sleep", "30"]
```
jr0d, line 23: I've tried `kubectl wait -n {{ .Release.Namespace }} --for=condition=Ready pods -l app=webhook`, which succeeds before the dynamic server certificate is populated in the service. This results in the post install operations  (issuer and  certificate creation) failing. This sleep duration seems to work consistently on my local machine and also for other cert-manager users in the wild. Still, I think it would be best to block until a test issuer creation succeeds. Here is a jira for this follow up: https://jira.d2iq.com/browse/D2IQ-71954

jimmidyson: If helm is run with `--wait` doesn't that wait for the deployment to be ready?

jr0d: It does, but the deployment becomes Available when all containers in the Pod template become `Ready`. Unfortunately, this happens before the service certificate is generated. So, using --wait, has the same issue unfortunately.

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -39,14 +39,16 @@ spec:
...
+  privateKey:
+    algorithm: {{ .keyAlgorithm | default "RSA" }}
+    encoding: {{ .keyEncoding | default "PKCS1" }}
```
sebbrandt87, line 44: out of api docs
>  allowed values are “pkcs1” and “pkcs8” standing for PKCS#1 and PKCS#8

are the docs wrong ? if so 🤦

staging/cert-manager-setup/templates/certificates.yaml
```diff
@@ -39,14 +39,16 @@ spec:
     - {{ . | quote -}}
   {{- end }}
 {{- end }}
-  keyAlgorithm: {{ .keyAlgorithm | default "rsa" }}
-  keyEncoding: {{ .keyEncoding | default "pkcs1" }}
-  keySize: {{ .keySize | default "2048" }}
+  privateKey:
+    algorithm: {{ .keyAlgorithm | default "RSA" }}
```
sebbrandt87, line 43: out of api docs
>  allowed values are either “rsa” or “ecdsa”

are the docs wrong ? if so 🤦

sebbrandt87 (changes requested):

sebbrandt87:
helm3 test seems to only succeed as the upgrade check isn't considered to run?

```
Skipping upgrade test of 'cert-manager-setup => (version: "0.2.0", path: "staging/cert-manager-setup")' because: 1 error occurred:
	* 0.2.0 does not have same major and minor version as 0.1.17


Testing upgrades of chart 'cert-manager-setup => (version: "0.2.0", path: "staging/cert-manager-setup")' relative to previous revision 'cert-manager-setup => (version: "0.2.0", path: "staging/cert-manager-setup")'...

Installing chart 'cert-manager-setup => (version: "0.2.0", path: "staging/cert-manager-setup")' with values file 'staging/cert-manager-setup/ci/general-test-values.yaml'...

Creating namespace 'cert-manager-setup-l1owzutxb6'...
```

Is this message true?

staging/cert-manager-setup/templates/certificates.yaml
jr0d: They are. Validation fails unless they are uppercase. It would have been nice if the enums contained both cases. I can convert them to uppercase to avoid validation issues with custom values.

staging/cert-manager-setup/templates/certificates.yaml
jr0d: They are. Validation fails unless they are uppercase. It would have been nice if the enums contained both cases. I can convert them to uppercase to avoid validation issues with custom values.

jr0d:
> Is this message true?

@sebbrandt87 From the chart-testing docs:

```
			in given chart directories. If upgrade (--upgrade) is true, then this
			command will validate that 'helm test' passes for the following upgrade paths:
			* previous chart revision => current chart version (if non-breaking SemVer change)
```
Upgrade tests are not performed on breaking changes, it seems.

staging/cert-manager-setup/templates/certificates.yaml
jimmidyson: Can you please send a PR upstream to update docs?

staging/cert-manager-setup/templates/certificates.yaml
jimmidyson: Again, a PR upstream would be awesome.

hectorj2f:
Please, add to this task bumping the federated_cert_manager version in yakcl.

jimmidyson:
@hectorj2f That is bumping the `FederatedAddon`, not chart, but I do agree this should be tackled as one initiative.

hectorj2f:
> @hectorj2f That is bumping the FederatedAddon, not chart, but I do agree this should be tackled as one initiative.

@jimmidyson I knew, I wanted to highlight the pain we faced when adding new addons to yakcl and never bump their versions. It needs to be maintained, and I believe the best time is whenever you bump the addon version, you should bump the federated addon too. That would help aligned the versions everywhere, otherwise we will get outdated addons how we already had/have.

hectorj2f:
@jimmidyson In addition, I added this comment to the chart repo because I know we often bump the addon (its chart version) but never the chart version or the template spec of the respective federated addon.

jr0d:
Upstream issue:

https://github.com/cert-manager/website/issues/330

These are likely auto-generated so I'll let them PR to fix.

sebbrandt87 (approved):
🤞

PR #791, opened 2020-09-17
fix: the port name in gatekeeper service

GoelDeepak:
**What type of PR is this?**
Bug

It fixes the port name from http to https. This is needed as it confuses istio which depend on port name of all the services in a cluster (don't be surprised ... its istio!! )

https://jira.d2iq.com/browse/D2IQ-71798

No

```release-note

```

**Checklist**


staging/gatekeeper/templates/service.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   - port: {{ .Values.service.port }}
     targetPort: http
     protocol: TCP
-    name: http
+    name: https
```
samba, line 17: This seems inconsistent -- `targetPort: http ... name: https`. Is this intentional? Should it be the same, both corrected to `https`?

joejulian (approved):

staging/gatekeeper/templates/service.yaml
GoelDeepak: I am not sure. Unlike `name` which is just a hint, `targetPort` is the actual port on which service is listening. I don't know enough to change the listening port.

joejulian: https://github.com/mesosphere/charts/blob/813c68dd6f8c8811e56ec243a74e3ae04d451dfc/staging/gatekeeper/templates/deployment.yaml#L48

joejulian: Isn't this actually http though?

joejulian (changes requested):
I linked in the comment to the line that needs change to keep the targetPort name consistent

staging/gatekeeper/templates/service.yaml
GoelDeepak: @joejulian yeah, it is `http`. However, the issue is here https://github.com/mesosphere/charts/blob/master/staging/gatekeeper/values.yaml#L39. Why that needs to be `443`?

joejulian: If it's http, why does it have all the certificate files?
https://github.com/mesosphere/charts/blob/813c68dd6f8c8811e56ec243a74e3ae04d451dfc/staging/gatekeeper/templates/deployment.yaml#L63-L68

I didn't think webhooks were allowed to _not_ be https.

joejulian: As an aside, the new upstream chart doesn't allow the service port to be specified in the values.

GoelDeepak: Thats nice. I guess the right course of action in that case is to just update the chart. I'll make necessary changes in this PR.

GoelDeepak:
Closing this in favor of upgrading the gatekeeper chart

PR #804, opened 2020-09-29
fix: override cert-manager release name to work with hardcoded values in CRD files

jr0d:
**What type of PR is this?**

Bug

Override cert-manager release name to work with hard coded values in CRD files.

no issue

```release-note
[cert-manager] 
```

**Checklist**


dkoshkin (approved):
These changes look good, I haven't looked if there are others though.

samvantran (commented):
Looks good but the PR title and description says `use template variables for webhook names` but the code seems to be using hardcoded values instead?

joejulian:
I think this needs to be done the other way around. Make the webhook  use the fixed name. As it is, if the name is changed in the addon, this will break again. If the webhook always uses a fixed name and is referenced by that same name everywhere, then it won't.

This is ok because we can only ever have cert-manager installed just once.

joejulian (changes requested):
I'm afraid that this needs to be done the other way around. Make the webhook use the fixed name. As it is, if the name is changed in the addon, this will break again. If the webhook always uses a fixed name and is referenced by that same name everywhere, then it won't.

This is ok because we can only ever have cert-manager installed just once.

jr0d:
@joejulian i went this route because the web hook not working just means we lose v1beta2 support, but sill allow names to be generated. I'll change it, when I get home.

joejulian (approved):

joejulian (approved):

joejulian:
Approved, but needs a chart version bump.

alejandroEsc (approved):
Yeah we discussed this, lgtm. yuck.

samvantran (approved):

PR #830, opened 2020-10-13
[cert-manager] feat: add upgrade orchestration

jr0d:
**What type of PR is this?**

Feature

Uses cert-manager-upgrade program to safely upgrade cert-manager from v1alpha1 to v1. 

https://jira.d2iq.com/browse/D2IQ-71558

code for cli is here: https://github.com/jr0d/cert-manager-upgrade/

```release-notes
TBD
```

**Checklist**


makkes (changes requested):
We have a pretty [strict policy](https://github.com/mesosphere/charts#repository-structure) on "staging" charts:
> A chart PR into the staging repository will not be accepted without an accompanying upstream pull request.

Would you mind creating or linking an upstream PR for this very change, please?

joejulian (approved):

staging/cert-manager-setup/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.2.3
+version: 0.2.4
 appVersion: 1.0.1
```
gracedo, line 5:
```suggestion
appVersion: 1.0.3
```

staging/cert-manager-setup/templates/v1upgrade.yaml
```diff
@@ -0,0 +1,113 @@
...
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-6"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+spec:
+  template:
+    spec:
+      serviceAccountName: cert-manager-upgrade
+      containers:
+        - name: cert-manager-upgrade
+          image: "jaredrodriguez/cert-manager-upgrade:v0.1.0"
```
samvantran, line 17: are we moving these to prod repo?

gracedo: The plan is yes. We were having a bit of a discussion where to - considered extra-steps but I think many want to deprecate it so we may create a new repo instead (perhaps you have the power to create new repos? 💪 ).

samvantran: i might know a guy 😏

staging/cert-manager-setup/templates/v1upgrade.yaml
```diff
@@ -0,0 +1,113 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: cert-manager-upgrade-backup
+  annotations:
+    "helm.sh/hook": pre-install
```
gracedo, line 8: do we need `pre-upgrade` with all the `pre-install`s?

alejandroEsc: yeah i think we do, pre-upgrade also here right?

staging/cert-manager-setup/templates/v1upgrade.yaml
alejandroEsc: If we need something fast, extra-steps seems like a legit place, just saying.

staging/cert-manager-setup/templates/v1upgrade.yaml
```diff
@@ -0,0 +1,113 @@
...
+      restartPolicy: Never
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: cert-manager-upgrade-restore
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "10"
```
alejandroEsc, line 28: upgrade restore? So this does what? and wis the hook-weight correct?

alejandroEsc (commented):
lgtm overall, a few questions, nothing major, Do we need both roles and clusterrroles?

staging/cert-manager-setup/templates/v1upgrade.yaml
```diff
@@ -0,0 +1,118 @@
+{{- if .Values.installCertManager }}
```
gracedo, line 1: Would it work to add the conditional to only run backup/restore `{{- if (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") -}}` here instead of doing it in the script?

gracedo: (referencing this part of the script https://github.com/jr0d/cert-manager-upgrade/blob/master/pkg/app/backup.go#L29)

stable/cert-manager-setup/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
-version: 0.2.3
+version: 0.2.4
 appVersion: 1.0.1
```
gracedo, line 5:
```suggestion
appVersion: 1.0.3
```

jr0d:
> We have a pretty [strict policy](https://github.com/mesosphere/charts#repository-structure) on "staging" charts:
> 
> > A chart PR into the staging repository will not be accepted without an accompanying upstream pull request.
> 
> Would you mind creating or linking an upstream PR for this very change, please?

The chart is actually not an upstream chart, it's a wrapper chart. I've moved it to stable.

staging/cert-manager-setup/templates/v1upgrade.yaml
jr0d: Backup is only for pre-install, since it needs a delete upgrade strategy. Helm doesn't support upgrading since it does deployment stuff outside of helm. If you were upgrading manually (outside of kubeaddons), you'd first do a helm delete, then let the script backup the resources and delete the CRDs.

staging/cert-manager-setup/templates/v1upgrade.yaml
jr0d: yeah, that name is a bit off.. the program is cert-manager-upgrade .. the command is restore. The weight is correct since I want to make absolutely sure this runs last, and ensure that anyone that adds hooks down the road knows that this should run last.

staging/cert-manager-setup/templates/v1upgrade.yaml
jr0d: yes it would.. and also not create jobs needlessly.

alejandroEsc:
> We have a pretty [strict policy](https://github.com/mesosphere/charts#repository-structure) on "staging" charts:
> 
> > A chart PR into the staging repository will not be accepted without an accompanying upstream pull request.
> 
> Would you mind creating or linking an upstream PR for this very change, please?

This is particularly our solution for our stack, it is an enhancement in many ways and as such need not be contributed back up. It would be nice to do so, however, i doubt it would be accepted in any reasonable time and probably not in the way we are doing so.

stable/cert-manager-setup/templates/v1upgrade.yaml
```diff
@@ -0,0 +1,124 @@
+{{- if .Values.installCertManager }}
```
gracedo, line 1: Just floating a previous comment back up here so it's not lost - to add a `{{- if (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") -}}` check to only run these backup and restore jobs if we're running the old version of certmanager

jr0d: done

PR #831, opened 2020-10-14
chore: bump the dex-controller memory limit

hectorj2f:
**What type of PR is this?**

Chore

During the MWT, we found that the dex-controller was killed due to OOM errors.

```release-note

```

**Checklist**

### DIFF
--- a/stable/dex-controller/values.yaml
+++ b/stable/dex-controller/values.yaml
@@ -9,7 +9,7 @@ controller:
     resources:
       limits:
         cpu: 100m
-        memory: 100Mi
+        memory: 200Mi
       requests:
         cpu: 100m
         memory: 20Mi
--- a/stable/dex-controller/values.yaml
+++ b/stable/dex-controller/values.yaml
@@ -9,7 +9,7 @@ controller:
     resources:
       limits:
         cpu: 100m
-        memory: 200Mi
+        memory: 250Mi
       requests:
         cpu: 100m
         memory: 20Mi
--- a/stable/dex-controller/values.yaml
+++ b/stable/dex-controller/values.yaml
@@ -9,7 +9,7 @@ controller:
     resources:
       limits:
         cpu: 100m
-        memory: 250Mi
+        memory: 512Mi
       requests:
         cpu: 100m
         memory: 20Mi
--- a/stable/dex-controller/Chart.yaml
+++ b/stable/dex-controller/Chart.yaml
@@ -3,6 +3,6 @@ name: dex-controller
 home: https://github.com/mesosphere/dex-controller
 appVersion: "v0.2.0"
 description: Dex controller
-version: 0.2.2
+version: 0.2.3
 maintainers:
-  - name: jieyu
+  - name: hectorj2f


hectorj2f:
image-2020-10-13-14-51-51-732

stable/dex-controller/values.yaml
```diff
@@ -9,7 +9,7 @@ controller:
     resources:
       limits:
         cpu: 100m
-        memory: 100Mi
+        memory: 250Mi
```
dkoshkin, line 12: During the MWT, it was stable after going to 750Mi.

At 300Mi it was still being OOMed.

hectorj2f: Yeah, let's set to 512Mi

dkoshkin (approved):

armandgrillet (approved):
LGTM.

gracedo (approved):

jr0d (approved):
lgtm

stable/dex-controller/values.yaml
```diff
@@ -9,7 +9,7 @@ controller:
     resources:
       limits:
         cpu: 100m
-        memory: 100Mi
+        memory: 512Mi
```
alejandroEsc, line 12: this is a bump of 5x's is this based on the current configurration of a default konvoy cluster or is this for a particular large node count topology?

alejandroEsc (commented):
quick question.

stable/dex-controller/values.yaml
hectorj2f: @aescobar the issue has happened in the past. I saw this problem with smaller cluster sizes. The size plays a role here but it is not the main motivation.

The main issue comes when dex-controller crashes for the first time with OOM, after that moment is hard to start it again because during the bootstrap operation crashes again. The bootstrap action requires a lot of `memory`. We saw this behavior in the past with different cluster sizes, but this time we tuned the memory limit to accommodate to that bootstrap action.

Second motivation is that `memory is cheap` and allocating 512Mi is nothing for us, in return of a better user experience.

Third, I saw this issue happening `more` often when using kommander and attaching clusters. However I have to say I also faced during our early MWTs.

alejandroEsc: Alright, thank you for answering! If this affects our current cluster then it makes sense. I was just worried that it was just for large Node cases and it wasnt worth the change, but it sounds like it is.

hectorj2f: Yes, let's say if the memory limit for big clusters would have been 9G then obviously I won't adapt it, but the required value to have a steady behavior seems quite low for us.

vespian (commented):
Do we know why the memory usage was so high? Is the memory usage independent of cluster size/number of pods in the cluster?

hectorj2f:
Is the memory usage independent of cluster size/number of pods in the cluster?

Dex is not managing pods, therefore that isn't a problem.

@vespian did you check https://github.com/mesosphere/charts/pull/831#discussion_r505697415 ?

vespian:
> @vespian did you check #831 (comment) ?

Yes, and the `...the size plays a role here...` part worries me. 

I would feel more confident if we knew why bootstrap requires so much memory. Knowing that we could better estimate if 512MB is sufficient for any cluster size.

hectorj2f:
@vespian Yes, this is related to the dex client and connector created during our mwt test. Please, note that every kommander cluster creates dex clients, so scaling that amount of clients could cause a problem. We don't want this to happen under any circurstances not even with a single client with many connectors.

I had in mind to collect the pprof files but I needed hands and extra-time to do that. I have a pending todo which consists on checking the graphs from the previous mwt.

PR #832, opened 2020-10-15
fix: delete federated resources prior to uninstallation

makkes:
**What type of PR is this?**

Bug

All the federated resources are now deleted as part of a pre-delete
hook so that the kubefed controller has the chance to remove the
finalizer from those resources. Without this change, all the federated
resources have always been left behind on clusters when the chart was
deleted because they would have the dangling finalizers on them.

The reason for this is that the controller managing these finalizers
(kubefed controller) is removed before the resources themselves.

https://jira.d2iq.com/browse/D2IQ-72106

I manually tested installation and uninstallation of this chart. After it has been deleted, no federated types are left on the cluster.

```release-note
All federated resources are now deleted as soon as Kommander is uninstalled.
```

**Checklist**


stable/kommander/templates/pre-deletion-purger.yaml
```diff
@@ -0,0 +1,91 @@
...
+  name: purger-config
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-2"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+data:
+  purge.sh: |
+    #!/bin/bash
+    for ft in $(kubectl get federatedtypeconfigs -n kommander -o jsonpath='{range .items[*]}{.spec.federatedType.kind}.{.spec.federatedType.group}{"\n"}{end}') ; do
+      k delete ${ft} --all --all-namespaces
```
jimmidyson, line 14:
```suggestion
      kubectl delete ${ft} --all --all-namespaces
```

stable/kommander/templates/pre-deletion-purger.yaml
```diff
@@ -0,0 +1,91 @@
...
+spec:
+  template:
+    metadata:
+      name: purger
+    spec:
+      restartPolicy: Never
+      serviceAccount: purger
+      containers:
+        - name: purger
+          image: "bitnami/kubectl:latest"
```
jimmidyson, line 79: Let's use a specific version, say `v1.19.3`?

stable/kommander/templates/pre-deletion-purger.yaml
```diff
@@ -0,0 +1,91 @@
...
+metadata:
+  name: purger
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-2"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+
+rules:
+  - apiGroups:
+      - '*'
```
jimmidyson, line 28: Could we limit the scope here to the specific federated types API group?

stable/kommander/templates/pre-deletion-purger.yaml
```diff
@@ -0,0 +1,91 @@
...
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    metadata:
+      name: purger
+    spec:
+      restartPolicy: Never
+      serviceAccount: purger
```
jimmidyson, line 76:
```suggestion
      serviceAccountName: purger
```

jimmidyson:
I thought we said that we would target the specific resources installed as part of the chart via label selection in the first cut, and provide the full purge as optional?

stable/kommander/templates/pre-deletion-purger.yaml
makkes: I explicitly made this a wildcard because if anyone adds a resource from an additional API group, this job would fail.

makkes:
> I thought we said that we would target the specific resources installed as part of the chart via label selection in the first cut, and provide the full purge as optional?

You are right. One problem that I encountered is that Helm doesn't put labels but annotations on the resources so I'd have to put more filtering logic into the script. Do we want to do that? Honest question.

jimmidyson:
We add labels to all other resources via the chart... https://github.com/mesosphere/yakcl/blob/master/chart/kommander-federation/templates/issuer_utility_apiserver.yaml#L6-L7... but not federated* resources deployed via the chart :facepalm:

stable/kommander/templates/pre-deletion-purger.yaml
```diff
@@ -0,0 +1,91 @@
...
+metadata:
+  name: purger-config
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-2"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+data:
+  purge.sh: |
+    #!/bin/bash
+    for ft in $(kubectl get federatedtypeconfigs -n kommander -o jsonpath='{range .items[*]}{.spec.federatedType.kind}.{.spec.federatedType.group}{"\n"}{end}') ; do
```
hectorj2f, line 13: What about the federated resources created in namespace kommander-system?

hectorj2f: Do we need to delete them?

makkes: This line only searches for `FederatedTypeConfigs` and the Kommander chart does not create any `FederatedTypeConfigs` outside of the `kommander` namespace.

hectorj2f: True!

hectorj2f (approved):

armandgrillet (approved):

makkes:
Will wait for @jimmidyson's final approval.

jimmidyson (changes requested):
Chart version needs a bump but otherwise :ok_hand:

jimmidyson (approved):

PR #835, opened 2020-10-15
Merge in forks of stable charts from helm/charts

joejulian:
**What type of PR is this?**

chore

The default helm charts are being retired by mid November. In order to support customers as we work toward a migration to using upstream charts we have decided to temporarily fork these charts from upstream.

https://jira.d2iq.com/browse/D2IQ-72544

```release-note
none
```

**Checklist**


makkes (commented):
Shouldn't these at least partly belong into `staging/`?

joejulian:
> Shouldn't these at least partly belong into `staging/`?

I don't think so. This isn't a temporary fork that we're maintaining while waiting for our changes to be merged upstream. This particular upstream is going away. We need these charts moved in order to support our current customers on the KBA versions they already have installed.

alejandroEsc (approved):
Not blocking. This is a bit scary, im guessing that we are doing this and not squashing cause we are keeping the history as well. I am ok with this being in stable etc.. as it will be maintained until  we no longer have tools pointing to these charts.

joejulian:
> ... not squashing cause we are keeping the history as well ...
Also, we do need to publish some previous versions so the legacy customers aren't interrupted.

dkoshkin (approved):

jr0d (approved):

PR #834, opened 2020-10-15
[KNative Serving] adding support to both Helm 2 and 3

akirillov:
**What type of PR is this?**
chore

**What this PR does/ why we need it**:
This PR adds backwards-compatible CRD handling for Helm 3 in a similar way it is done for e.g. Istio chart.

**Which issue(s) this PR fixes**:
[D2IQ-72522: KNative addon installation fails on Konvoy 1.6](https://jira.d2iq.com/browse/D2IQ-72522)

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
NONE

**Checklist**

* [x] *If a chart is changed, the chart version is correctly incremented.*
* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #838, opened 2020-10-16
[KNative Serving] adding Helm 2 annotations to CRDs

akirillov:
**What type of PR is this?**
bug

This PR is a follow-up to https://github.com/mesosphere/charts/pull/834. It adds Helm webhook annotations back to CRDs for Helm 2. The annotations were mistakenly removed in the previous PR.

**Checklist**


staging/knative/charts/serving/crds/crds.yaml
```diff
@@ -26,7 +27,6 @@ kind: CustomResourceDefinition
 metadata:
   name: certificates.networking.internal.knative.dev
   labels:
-    serving.knative.dev/release: {{ .Chart.AppVersion | quote }}
```
armandgrillet, line 29: What is the reason behind removing this label?

armandgrillet (approved):
LGTM.

staging/knative/charts/serving/crds/crds.yaml
akirillov: I was getting the following error during the CRD installation steep with Helm 3:
```
Error: failed to install CRD crds/crds.yaml: error parsing : error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Chart.AppVersion | quote":interface {}(nil)}
```
According to [Helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you):
> There is now a special directory called crds that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart.

As these labels are not used anywhere in KNative code, I've just removed them to resolve the templating issue.

akirillov:
soft ping @alejandroEsc and @joejulian. Can you please provide you reviews as you're familiar with the previous PR for KNative Helm 3 support? Thanks!

staging/knative/charts/serving/files/crds.yaml
```diff
@@ -395,9 +414,11 @@ spec:
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    helm.sh/hook: crd-install
+    helm.sh/hook-delete-policy: before-hook-creation
```
alejandroEsc, line 419: `helm.sh/hook-delete-policy: before-hook-creation` these are not recommended. These will delete any resources whenever they are run that are of the types defined by the CRD. I would remove these.

alejandroEsc (commented):
One item of concern, but lgtm overall.

staging/knative/charts/serving/files/crds.yaml
akirillov: Thanks for pointing that out. Removed.

joejulian (approved):

PR #851, opened 2020-10-23
defaultstorageclass: Enable metrics collection

gracedo:
**What type of PR is this?**

Bug

Annotation-based service discovery has been deprecated in prometheus-operator so these metrics were no longer getting collected. Switch to use the service monitor model by adding this label to allow it to get scraped by Prometheus. This requires a change to the Prometheus addon config to add this metrics endpoint (will put up kba pr for this later after including all other necessary prom config changes to get metrics collected from other components):

```diff
--- a/addons/prometheus/prometheus.yaml
+++ b/addons/prometheus/prometheus.yaml
@@ -100,6 +100,24 @@ spec:
               # configuration.
               - targetPort: 10902
                 interval: 30s
+          - name: kubeaddons-service-monitor-metrics-defaultstorageclass
+            selector:
+              matchLabels:
+                servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
+                kubeaddons.mesosphere.io/name: "defaultstorageclass"
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - port: https
+                interval: 30s
+                scheme: https
+                bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+                tlsConfig:
+                  caFile: "/etc/prometheus/secrets/dstorageclass-webhook-server-cert/ca.crt"
+                  certFile: "/etc/prometheus/secrets/dstorageclass-webhook-server-cert/tls.crt"
+                  keyFile: "/etc/prometheus/secrets/dstorageclass-webhook-server-cert/tls.key"
+                  insecureSkipVerify: true
           - name: kubeaddons-service-monitor-api-v1-metrics-prometheus
             selector:
               matchLabels:
@@ -243,6 +261,7 @@ spec:
           walCompression: true
           secrets:
             - etcd-certs
+            - dstorageclass-webhook-server-cert
           externalUrl: "/ops/portal/prometheus"
           storageSpec:
             volumeClaimTemplate:
```

https://jira.d2iq.com/browse/D2IQ-72691

```release-note
NONE
```

**Checklist**

### DIFF
--- a/stable/defaultstorageclass/templates/service.yaml
+++ b/stable/defaultstorageclass/templates/service.yaml
@@ -5,7 +5,8 @@ metadata:
   creationTimestamp: null
   labels:
     control-plane: controller-manager
-    servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
+    servicemonitor.kubeaddons.mesosphere.io/path: metrics
+    kubeaddons.mesosphere.io/name: defaultstorageclass
   name: dstorageclass-controller-manager-metrics-service
   namespace: {{ .Release.Namespace }}
 spec:
--- a/stable/defaultstorageclass/templates/service.yaml
+++ b/stable/defaultstorageclass/templates/service.yaml
@@ -2,13 +2,10 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    prometheus.io/port: "8443"
-    prometheus.io/scheme: https
-    prometheus.io/scrape: "true"
   creationTimestamp: null
   labels:
     control-plane: controller-manager
+    servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
   name: dstorageclass-controller-manager-metrics-service
   namespace: {{ .Release.Namespace }}
 spec:
--- a/stable/defaultstorageclass/Chart.yaml
+++ b/stable/defaultstorageclass/Chart.yaml
@@ -4,7 +4,7 @@ description: Webhooks and controllers relating to default storage classes
 name: defaultstorageclass
 maintainers:
   - name: joejulian
-version: 0.0.3
+version: 0.0.4
 kubeVersion: ">=1.15.0"
 home: https://github.com/mesosphere/defaultstorageclass
 sources:


gracedo:
Tested this manually, saw the target come up healthy
image

stable/defaultstorageclass/templates/service.yaml
```diff
@@ -2,13 +2,11 @@
...
 metadata:
-  annotations:
-    prometheus.io/port: "8443"
-    prometheus.io/scheme: https
-    prometheus.io/scrape: "true"
   creationTimestamp: null
   labels:
     control-plane: controller-manager
+    servicemonitor.kubeaddons.mesosphere.io/path: metrics
+    kubeaddons.mesosphere.io/name: defaultstorageclass
```
gracedo, line 9: I'm not in love with this label. Open to using other unique identifying labels for this chart/addon

hectorj2f: Yes, I am not either.

hectorj2f (changes requested):
@gracedo there is another way such as reusing one of the existing scraping configurations for addons. The solution would be to adapt the port-name to be named metrics. Let's sync later.

gracedo:
> @gracedo there is another way such as reusing one of the existing scraping configurations for addons. The solution would be to adapt the port-name to be named metrics. Let's sync later.

The issue I ran into here (and with dex-controller) was that both these endpoints are https and needed tlsConfigs. I could reuse the metrics servicemonitor definition as is by moving this entire endpoint into that def, but dex-controller uses the same port/name so both endpoints would be targeted, however they have different certs. We could rename the port in both so they are unique, but then I’m not sure if that is any better than defining two servicemonitors. I can be persuaded tho :) and maybe there is another workaround I haven’t thought of which is likely!

hectorj2f:
For systems such as dex-controller, I won't touch it and add this mechanism you used here (it is better to keep the service and do not restart it). For this chart, we can simply change the metrics port, it has no major impact to expose these metrics of this chart.

hectorj2f:
@gracedo For the sake of fixing the issue now, let's merge this approach.

hectorj2f (approved):

branden (approved):

PR #853, opened 2020-10-23
external-dns: Add service.labels option

gracedo:
**What type of PR is this?**

Feature/chore

Add option to define service labels, mainly so we can use a label to get its metrics discovered and scraped by prom. This func was already added to a later version in the upstream chart (https://github.com/bitnami/charts/commit/41d917aba5946d290580c96afb295b9b9207dc59#diff-b37fd4e4217d56abe479147564c4c1afc64d680c8c0612701f6e14ba7e3a8ef7) but we should not jump a major version so close to the release date in case something breaks. So, since we already have this fork, I've ported over this func here and will point us to this fork in KBA.

https://jira.d2iq.com/browse/D2IQ-72698

```release-note
NONE
```

**Checklist**

### DIFF
--- a/stable/external-dns/templates/service.yaml
+++ b/stable/external-dns/templates/service.yaml
@@ -3,6 +3,9 @@ kind: Service
 metadata:
   name: {{ template "external-dns.fullname" . }}
   labels: {{ include "external-dns.labels" . | nindent 4 }}
+  {{- if .Values.service.labels -}}
+  {{ toYaml .Values.service.labels | nindent 4 }}
+  {{- end }}
   {{- if .Values.service.annotations }}
   annotations: {{ toYaml .Values.service.annotations | nindent 4 }}
   {{- end }}
--- a/stable/external-dns/templates/service.yaml
+++ b/stable/external-dns/templates/service.yaml
@@ -30,5 +30,12 @@ spec:
     {{- if and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort)) }}
     nodePort: {{ .Values.service.nodePort  }}
     {{- end }}
+  - name: metrics
+    port: {{ .Values.service.port }}
+    protocol: TCP
+    targetPort: http
+    {{- if and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort)) }}
+    nodePort: {{ .Values.service.nodePort  }}
+    {{- end }}    
   selector: {{ include "external-dns.matchLabels" . | nindent 6 }}
   type: {{ .Values.service.type }}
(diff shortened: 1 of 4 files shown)


hectorj2f:
it looks good to me, as long as we add servicemonitor.kubeaddons.mesosphere.io/path: metrics to the eaddon.

hectorj2f (approved):

hectorj2f (commented):
Screen Shot 2020-10-23 at 1 27 06 PM
Adding a new port name metrics and the label worked to collect the metrics

stable/external-dns/templates/service.yaml
```diff
@@ -27,5 +30,12 @@ spec:
     {{- if and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort)) }}
     nodePort: {{ .Values.service.nodePort  }}
     {{- end }}
+  - name: metrics
```
gracedo, line 33: My only concern here is that this change deviates from upstream, but we can merge this first and then see if we can contribute something back upstream to make it work for us (since we are switching to our fork for the release anyways)

gracedo: hmm does this work? I tried manually adding this new `metrics` endpoint to the running service but I'm getting `Service "external-dns-kubeaddons" is invalid: spec.ports[1]: Duplicate value: core.ServicePort{Name:"", Protocol:"TCP", AppProtocol:(*string)(nil), Port:7979, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:""}, NodePort:0}`, I don't think we can specify two endpoints for the same port

alejandroEsc: Why are we defining duplicates here?

gracedo: I think the intention was to reuse the existing servicemonitor endpoint targeting port named `metrics` in prometheus here https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/prometheus.yaml#L93-L94 but since we can't define dupe ports in the service we can either add `port: http` to the list of prom endpoints, or `targetPort: 7979` which is likely unique to external-dns service. I tested adding `http`, it works but also pulls in thanos query endpoints so to resolve that we should then remove the thanos query target https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/prometheus.yaml#L97-L102

gracedo: I'm going to revert this change

gracedo: We can decide how to pull this into prom in the KBA PR

branden (approved):

PR #908, opened 2020-11-23
chore(prometheus-operator): bump prometheus image tag to 2.20.0

geekbass:
**What type of PR is this?** Chore

**What this PR does/ why we need it**: Default image tag needs updated to what we run in SOAK. This will also be used to ensure Prometheus image tag is updated for [federation](https://github.com/mesosphere/yakcl/pull/270). 

**Which issue(s) this PR fixes**: D2IQ-73110

```release-note
NONE
```

**Checklist**

### DIFF
--- a/staging/prometheus-operator/values.yaml
+++ b/staging/prometheus-operator/values.yaml
@@ -1611,7 +1611,7 @@ prometheus:
     ##
     image:
       repository: quay.io/prometheus/prometheus
-      tag: v2.18.2
+      tag: v2.20.0
       sha: ""
 
     ## Tolerations for use with node taints
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 9.3.3
+version: 9.3.4
 appVersion: 0.38.1
 home: https://github.com/coreos/prometheus-operator
 keywords:


gracedo:
There's actually a PR in KBA to bump prom image to 2.22.2 https://github.com/mesosphere/kubernetes-base-addons/pull/681 - question of whether that should be defined here in charts or in the KBA layer

joejulian:
I agree with @gracedo. I'll close this in favor of that.

geekbass:
This was a discussion that I had with @hectorj2f regarding also managing federation. He and I both thought it made more sense to have a single location to manage this and mentioned above. Adding him to discussion and reopening.

gracedo:
@geekbass I think we may want to consider updating our prometheus-operator fork to the latest version then to pull in a more recent version of prometheus. and if we want an even higher version, then contribute version bump that upstream. the upgrade unfortunately may not be straightforward as the helm chart was deprecated; the prometheus-operator chart has since been renamed and moved to https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack

another option to have a single location to set the prom version would be to create a parent chart that pulls prometheus-operator in as a subchart, which would solve many other problems (namely, moving off our fork and pulling in upstream chart) but would involve a different values hierarchy and I'm not sure what that upgrade strategy would look like.

geekbass:
I'm fine with whatever is decided honestly. Just need to make sure that our prometheus addon is consistent across all deployments whether through federation etc...

hectorj2f:
yes, let's upgrade it but let's make sure we don't need to add unnecessary settings to the prometheus addon (making it hard to maintain or understand), e.g. https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/prometheus.yaml#L233

staging/prometheus-operator/values.yaml
```diff
@@ -1611,7 +1611,7 @@ prometheus:
     ##
     image:
       repository: quay.io/prometheus/prometheus
-      tag: v2.18.2
+      tag: v2.20.0
```
hectorj2f, line 1614: I suggest we bump it to 2.21.0 which is now available.

sebbrandt87: it could already break with switching to `v2.20.0`
the prometheus-operator in the version we use doesn't have it in the compatibility list
https://github.com/prometheus-operator/prometheus-operator/blob/v0.38.1/Documentation/compatibility.md#prometheus

hectorj2f: We need to bump the operator to support this new version.

gracedo: We have a ticket to spike out upgrading prometheus-operator to using the new chart https://jira.d2iq.com/browse/D2IQ-73452 (anybody is free to work on it ;) )

gracedo:
I am closing this as prometheus has been upgraded to a recent version that bumped prometheus to 2.22.1

PR #911, opened 2020-11-25
kommander: Clean up SAs/roles

gracedo:
**What type of PR is this?**

Chore

Create SAs/roles per chart instead of reusing SA's created in parent charts (new SA/roles per subchart, create new SA for kommander chart instead of using the same one also defined in kommander-ui). This also allows us to finally get rid of the hardcoded SA names we were passing into the subcharts.

no issue

I tested this by hosting it on my GH - config branch `gracedo/kommander_0132`. Tested upgrade from prev version + fresh install on kommander cluster

```release-note

```

**Checklist**

### DIFF
--- a/stable/kommander/charts/kommander-karma/values.yaml
+++ b/stable/kommander/charts/kommander-karma/values.yaml
@@ -5,12 +5,6 @@
 # Internal address for the cluster's alertmanager.
 # alertmanagerAddress: "HOST:PORT"
 alertmanagerAddress: ""
-# Kommander service account used to delete the karma configmap
-# Do not use, this will be deprecated
-kommanderServiceAccount: kommander-kubeaddons
-# Name of the karma configmap
-# Do not use, this will be deprecated
-kommanderKarmaConfigMap: kommander-kubeaddons-config
 
 federate:
   addons: true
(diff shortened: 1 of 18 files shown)


DanielMSchmidt (approved):

stable/kommander/charts/kommander-karma/values.yaml
```diff
@@ -6,8 +6,10 @@
 # alertmanagerAddress: "HOST:PORT"
 alertmanagerAddress: ""
 # Kommander service account used to delete the karma configmap
+# Do not use, this will be deprecated
 kommanderServiceAccount: kommander-kubeaddons
```
makkes, line 10: Should this not just get removed and the SA created above be used instead?

stable/kommander/charts/kommander-thanos/values.yaml
```diff
@@ -6,6 +6,7 @@
 # thanosAddress: "HOST:PORT"
 thanosAddress: ""
 # Kommander service account used to delete Thanos store configmaps
+# Do not use, this will be deprecated
 kommanderServiceAccount: kommander-kubeaddons
```
makkes, line 10: same question as with karma sub-chart.

makkes (approved):
2 questions, otherwise lgtm. Thanks for taking point here! Very important change.

stable/kommander/charts/kommander-karma/values.yaml
gracedo: yeah, I think it should be safe to just remove the value. I was following another old example but I think it's not necessary to add a deprecation warning here.

makkes (approved):

PR #917, opened 2020-12-04
Faiq/kommander catalog api bumps

faiq:
**What type of PR is this?**

Bug 

The kommander catalog bump wasn't working before due to errors in Dispatchfile

```release-note

```

**Checklist**


Dispatchfile
```diff
@@ -157,7 +157,7 @@ git_resource(
     url="https://github.com/mesosphere/" + kubeaddons_kommander_git,
     revision="master")
 do_bump_kommander = bump_addon("bump-kommander", kubeaddons_kommander_git)
-do_bump_catalog="bump-kommander-catalog-api"
+do_bump_catalog="bump-catalog-api"
```
cprovencher, line 160: the format "do_something" should be a function call
I don't think this variable is used except the line directly below

Dispatchfile
```diff
@@ -201,6 +201,7 @@ action(tasks=[do_bump_addon], on=push(branches=["master"], paths=["!" + kommande
 action(tasks=["test-helm2", "test-helm3"], on=pull_request(chatops=["test"]))
 action(tasks=["test-helm2"], on=pull_request(chatops=["test-helm2"]))
 action(tasks=["test-helm3"], on=pull_request(chatops=["test-helm3"]))
+action(tasks=[do_bump_kommander_catalog_api], on=pull_request(chatops=["bump-kommander"]))
```
cprovencher, line 204: should the chatop be more accurate? you're not bumping kommander but kommander catalog api

cprovencher (changes requested):

PR #919, opened 2020-12-06
fix: do not patch webhook

jr0d:
**What type of PR is this?**

Bug

Allow upgrades to succeed without intervention when kommander is disabled

no issue

This "fix" for the authorized lister aggregated API should not be needed as kommander should update the service with the correct API group: https://github.com/mesosphere/yakcl/blob/master/chart/kommander-federation/templates/apiservice_authorizedlister.yaml#L5

The reason we added this fix was we were seeing some inconsistent behaviour where  the old apigroup for cert-manager was still present after the kommander upgrade. If we continue to see this behaviour, a fix should be made to the kommander chart. 
 
```release-note
NONE
```

**Checklist**


stable/cert-manager-setup/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: cert-manager-setup
 home: https://github.com/mesosphere/charts
 version: 0.2.6
```
branden, line 4:
```suggestion
version: 0.2.7
```

branden (changes requested):

branden (approved):

joejulian (approved):

PR #922, opened 2020-12-09
prometheus-operator: Add patch/ to .helmignore

gracedo:
**What type of PR is this?**

chore

Add patch script that adds `patch/` into .helmignore so helm does not include that dir when building the chart.

no issue

```release-note
NONE
```

**Checklist**


staging/prometheus-operator/.helmignore
```diff
@@ -24,3 +24,5 @@ OWNERS
 hack/
 ci/
 prometheus-operator-*.tgz
+# Mesosphere-specific files to ignore
+patch/
```
dkoshkin, line 28: Whats the function of the `staging/prometheus-operator/patch/patch_10_mesosphere_helmignore.sh` file, is this not enough?

gracedo: Since this is our fork of the upstream chart, we want to include any modifications we make on top of the upstream chart in these patch files to make upgrading easier - so in the next upgrade, we can just directly apply these patches on top of the updated chart (like a rebase) instead of manually going through and re-applying each change.

branden (approved):

samvantran (approved):
Nice

dkoshkin (approved):

PR #943, opened 2021-01-06
chore: Adds zipkin service for jaeger integration

GoelDeepak:
**What type of PR is this?**
Chore

Istio depends on zipkin service for jaeger integration

https://jira.d2iq.com/browse/D2IQ-73677

```release-note

```

**Checklist**


staging/istio/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: istio
-version: 1.8.1
+version: 1.8.2
```
vespian, line 3: It would be nice to come up with a different versioning schema in the future. The current approach may be confusing as sometimes the chart version matches the app version sometimes it does not.

staging/istio/templates/configmap-crd.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
   name: istio-crd
 data:
   crd.yaml: |-
-{{.Files.Get "files/crd.yaml" | printf "%s" | indent 4}}
+{{.Files.Get "crds/crd.yaml" | printf "%s" | indent 4}}
```
vespian, line 8: Can we somehow integrate these two directories?
```diff
$ diff -Nwrup files/ crds/
diff -Nwrup files/crd.yaml crds/crd.yaml
--- files/crd.yaml      2021-01-06 17:44:18.959912347 +0100
+++ crds/crd.yaml       2021-01-06 17:44:18.959912347 +0100
@@ -2,8 +2,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: istiooperators.install.istio.io
-  annotations:
-    "helm.sh/hook": crd-install
   labels:
     chart: istio
 spec:
```

staging/istio/templates/zipkin-svc.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
```
vespian, line 1: How/where is zipkin installed? I wonder if this file belongs here.

vespian (commented):
Left few comments.

staging/istio/Chart.yaml
GoelDeepak: yeah, I agree. I wished if we could have something like 1.8.1-1. But, I remember that lint doesn't like it

staging/istio/templates/configmap-crd.yaml
GoelDeepak: Good Catch! I missed that annotation. Let me carry over and test it out.

staging/istio/templates/zipkin-svc.yaml
GoelDeepak: Its very confusing the way it is implemented in Istio. Bear with me! The actual application is Jaeger. Jaeger use to be part of Istio installation but no more. So now it is deployed as an independent addon. However, Istio still depend on a service named `zipkin` (don't ask me why the service name is `zipkin`) to connect to Jaeger. As Jaeger is separated from Istio, it has stopped creating this service. Hence, we have to create this service for istio to talk to Jaeger

vespian: lgtm.

can we put this as a comment somewhere?

staging/istio/templates/configmap-crd.yaml
GoelDeepak: I have included these annotations and verified that helm3 doesn't complain about these extra annotations

staging/istio/Chart.yaml
joejulian: Even better is to not even try to make the chart version match the app version. They're not related, nor should they be.

joejulian (approved):

vespian (approved):

PR #969, opened 2021-02-01
prometheus: update jq images

gracedo:
**What type of PR is this?**

chore / bug (cve)

Since the jobs using these images are added as patches on top of upstream, we should just update the problem images directly in the chart (currently done in kba https://github.com/mesosphere/kubernetes-base-addons/pull/828). The upgraded images also required [this chart change](https://github.com/mesosphere/charts/commit/822b1fd0a52d0458a3c5f43dcc708d2298c58ca1#diff-2c7a422c328fd7df5eb69b66a18fd5b69aab94f48451a1d39ad2096d8e8da8a6) so another reason why I think it makes sense to bump the image in the chart.

note: this img was already applied in kba configs so this chart bump is not necessary for this next release.

note #2: I will also be adding these updates to a patch version for this chart 9.3.x for use in yakcl for federated prometheus addon

```release-note

```

**Checklist**


joejulian (commented):
Why would you prefer to make the change here, requiring yet another patch, instead of just making the change in the addon?

gracedo:
> Why would you prefer to make the change here, requiring yet another patch, instead of just making the change in the addon?

The image changes (applied in the addon) ended up breaking the chart, which required a chart change (https://github.com/mesosphere/charts/commit/822b1fd0a52d0458a3c5f43dcc708d2298c58ca1#diff-2c7a422c328fd7df5eb69b66a18fd5b69aab94f48451a1d39ad2096d8e8da8a6) so to me it felt appropriate to move the image change up to the chart. These images are only used in our mesosphere files anyways so it didn't feel like a big deal to change the default here. Not a big deal either way, I just preferred to push up those changes up to the chart, especially if the older images had some CVEs

joejulian (approved):

hectorj2f (approved):

hectorj2f:
Is there anything blocking to merged this PR?

jimmidyson (approved):

PR #992, opened 2021-02-11
kubecost: Bump kubecost to v1.74.0

gracedo:
**What type of PR is this?**

chore/bug

Bumps kubecost to latest version which includes a [fix](https://github.com/kubecost/cost-analyzer-helm-chart/pull/765) for the previous v1.73.0 version which introduced a change to always deploy a psp which caused collision issues in kommander. 

```release-note
fix(kubecost): Disable PSP which causes name collisions if multiple instances of Kubecost are running in a cluster
```

**Checklist**


hectorj2f (approved):

stable/kubecost/values.yaml
```diff
@@ -57,6 +57,9 @@ cost-analyzer:
     enabled: true
     # storageClass: "-"
 
+  podSecurityPolicy:
+    enabled: false
```
hectorj2f, line 61: Do we want that ? Was it the default behavior before ?

hectorj2f: The default for kubecost users is true :/. 

Should we disable it for the main mgmt or better use kubefed overrides to disable it in the federated Addon kubecost when placed in the mgmt cluster (which we might have a label for 🤔 ) ?

gracedo: It was defaulted to false in the past, the latest releases changed the default to true which caused this to break in kommander. I think maybe the correct fix is to use a unique name for PSP so there won't be name conflicts, I can try and put up a pr against kubecost

hectorj2f: okay, sounds good. Let's track it in a jira ticket. It sounds good for me disabling now.

gracedo: https://jira.d2iq.com/browse/D2IQ-74182 will open a pr upstream

hectorj2f: Great @gracedo :).

gracedo: https://github.com/kubecost/cost-analyzer-helm-chart/pull/773

gracedo:
There are a couple of patches on 0.13.x for kommander 1.3/konvoy 1.7, let's get those in first before bumping the version here

makkes (approved):

hectorj2f (approved):

joejulian (approved):

branden (approved):

PR #993, opened 2021-02-11
prometheus: Add missing pre-delete hook to SA

gracedo:
**What type of PR is this?**

bug (COPS)

A prometheus pre-delete SA hook was missing, which was required for the pre-delete jobs, causing addon deletion/cleanup to fail. Will open another PR to backport this fix to 9.3.x

https://jira.d2iq.com/browse/D2IQ-74181
https://jira.d2iq.com/browse/COPS-6807

```release-note
fix(prometheus): Adds a missing service account which caused addon deletion/cleanup to fail
```

**Checklist**

### DIFF
--- a/staging/prometheus-operator/patch/mesosphere/templates/hooks/prometheus-cluster-id-hooks.yaml
+++ b/staging/prometheus-operator/patch/mesosphere/templates/hooks/prometheus-cluster-id-hooks.yaml
@@ -9,7 +9,7 @@ metadata:
     app: {{ template "kube-prometheus-stack.name" . }}-prometheus
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
   annotations:
-    helm.sh/hook: pre-install
+    helm.sh/hook: pre-install,pre-delete
     helm.sh/hook-weight: "-5"
     helm.sh/hook-delete-policy: hook-succeeded,before-hook-creation
 ---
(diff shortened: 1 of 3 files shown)


branden (approved):

dkoshkin (approved):

staging/prometheus-operator/patch/mesosphere/templates/hooks/prometheus-cluster-id-hooks.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     app: {{ template "kube-prometheus-stack.name" . }}-prometheus
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
   annotations:
-    helm.sh/hook: pre-install
+    helm.sh/hook: pre-install,pre-delete
```
alejandroEsc, line 12: I have a few questions about this, including checking kba testing as why this was not an issue prior. Still without the hook the SA would be available right?

gracedo: I don't think this happens often, but in some scenarios perhaps the SA was removed prematurely causing the pre-delete jobs to fail. The SA is also created here https://github.com/mesosphere/charts/blob/master/staging/prometheus-operator/templates/prometheus/serviceaccount.yaml - perhaps it would be cleaner for our mesosphere hooks to create and use a different SA?

PR #970, opened 2021-02-01
[release/prometheus-operator-9.3.x] fix images (cve)

gracedo:
**What type of PR is this?**

bug (cve)

**What this PR does/ why we need it**:

Update images to mitigate cve's. This is a patch on v9.3.x for federated addons in yakcl, since delete strategies for upgrades are not yet supported for federated addons. After this is merged, we need to run the publish job on the release branch to publish this patch chart version.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

* [ ] *If a chart is changed, the chart version is correctly incremented.*
* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #994, opened 2021-02-11
[release/prometheus-operator-9.3.x] Add missing pre-delete hook to SA

gracedo:
**What type of PR is this?**

bug (backport)

Backports #993 to pull into Konvoy 1.6 (and 1.5?) patch release. My concern is that we have already applied a patch to 9.3.x here: https://github.com/mesosphere/charts/pull/970 which was done for konvoy 1.7 to mitigate some CVEs; this patch was pulled into yakcl federated prometheus to avoid upgrading the federatedaddon. So these extra CVE mitigations would be pulled into a konvoy 1.6/1.5 patch.

```release-note

```

**Checklist**

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 9.3.6
+version: 9.3.7
 appVersion: 0.38.1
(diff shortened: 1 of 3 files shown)


branden (approved):

dkoshkin (approved):
This could be used in [KBA 2.x](https://github.com/mesosphere/kubernetes-base-addons/blob/testing-1.17-2.6.0/addons/prometheus/prometheus.yaml) used by Konvoy v1.5.x.
Konvoy v1.6 can use the same KBA 3.x version as Konvoy 1.7.x.

PR #1002, opened 2021-02-17
kommander / opsportal: Bump kommander-ui to 6.85.1

gracedo:
**What type of PR is this?**

bug

Bump kommander-ui to fix non-konvoy cluster platform services page https://github.com/mesosphere/kommander-ui/pull/3292. Would be great to include this in the next kommander 1.3.1 patch release.

https://jira.d2iq.com/browse/D2IQ-74138

```release-note
fix(kommander): Fix empty non-Konvoy cluster Platform Services tab 
```

**Checklist**


gracedo:
@TattdCodeMonkey @natmegs We discussed back at the last release that we would release this ui fix in a patch release - I noticed there is a jump in version from 6.84.3 to 6.85.1 which includes this fix. Can you guys double check what other changes would be included in this version jump and whether it's okay to include them in the next kommander/konvoy patch release?

TattdCodeMonkey (approved):

stable/kommander/Chart.yaml
```diff
@@ -9,4 +9,4 @@ maintainers:
   - name: gracedo
   - name: dkoshkin
 name: kommander
-version: 0.13.36
+version: 0.13.37
```
joejulian, line 12: This includes a minor version update to kommander-ui so the chart version should also have a minor version bump.
```suggestion
version: 0.14.0
```

joejulian (changes requested):

stable/kommander/Chart.yaml
s12chung: makes sense.

gracedo: makes sense to me too. thus far kommander has (more or less) bumped patch versions for changes within a kommander release, then bumped the minor when the release goes out. I think it makes sense to actually follow semver, we just need to get everyone on the same page so it's consistent

joejulian (approved):

branden (approved):

PR #1003, opened 2021-02-22
[Gatekeeper] feature: enable mutations on gatekeeper and allow proxy settings

alejandroEsc:
**What type of PR is this?**

feature

Allow for mutations and proxy settings to gatekeeper so that new pods will have those correct environment variables.

https://jira.d2iq.com/browse/D2IQ-70021

```release-note
Gatekeeper changes to allow for proxy settings and mutate pods with those
```

**Checklist**


staging/gatekeeper/UPGRADING.md
```diff
@@ -0,0 +1,17 @@
...
+```sh
+./upgrade.sh
+```
+
+The upgrade script:
+- clones the helm/charts repo `master` branch
+- copies all the files under `patch/crds` to the crds folder
+- replaces all the files, of the same name, in this fork with the newer upstream versions; adds the files if they're new.
+- applies all the patches stored in the [/patch](./patch) folder.
+  - note: each of these patch files adds a git commit message when applied. This makes it easier to review as each commit will show what changes were needed.
```
s12chung, line 17: 👍

staging/gatekeeper/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "3.2.2"
```
s12chung, line 2: shouldn't this change....? `3.3.0` perhaps?

staging/gatekeeper/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
...
 description: Gatekeeper - Policy Controller for Kubernetes
 name: gatekeeper
-version: 0.5.2
+version: 0.5.3
```
s12chung, line 5: of `appVersion` changes a minor or major version, change this to reflect that.

staging/gatekeeper/patch/patch_2_upgrade_crds_job.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# This patch adds mesosphere specific crd upgrade .
+
+source $(dirname "$0")/helpers.sh
+
+set -x
+
+SRCFILE="${BASEDIR}"/templates/crds.yaml
+
+sed -i '' -e '/# Create mesosphere value entries/,$d' ${SRCFILE}
```
s12chung, line 11: this won't work on linux btw :p maybe changing the `README` to tell devs to run `patch.sh` in the container is best? that way we don't need to think about OS diffs.

s12chung (commented):
overall, lgtm, just a bunch of little things.

staging/gatekeeper/patch/patch_2_upgrade_crds_job.sh
alejandroEsc: yeah thats definitely an improvement! something for the near future, but i thought this does work on linux?

staging/gatekeeper/Chart.yaml
alejandroEsc: yeah at this point;

staging/gatekeeper/Chart.yaml
alejandroEsc: ill bump to minor version instead of the patch.

staging/gatekeeper/patch/patch_2_upgrade_crds_job.sh
alejandroEsc: this was taken from the promtheus-operator chart, fyi

joejulian (approved):

staging/gatekeeper/patch/patch_2_upgrade_crds_job.sh
s12chung: this `SEDI` pattern is all over our makefiles :p.... ex. https://github.com/mesosphere/konvoy/blob/e1bd7668650c9adc98405d94dbeb8f1971476b12/auto-provisioning/make/platform.mk#L17

s12chung: but sure, we can change it later. let the linux guys see it later :p

s12chung (approved):

PR #1021, opened 2021-03-05
[kommander] chore: allow adding labels to kommander namespace

alejandroEsc:
**What type of PR is this?**

chore

Allow appending namespaces to the deployed namespace where kommander lives

tested this locally

```release-note

```

**Checklist**


stable/kommander/templates/namespace-labeler.yaml
```diff
@@ -0,0 +1,62 @@
...
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+spec:
+  template:
+    spec:
+      serviceAccountName: namespace-label
+      containers:
+        - name: namespace-label
+          image: "bitnami/kubectl:1.19.7"
```
s12chung, line 18: forgot if this image had a CVE

s12chung (approved):
just a note. otherwise, lgtm.

stable/kommander/templates/namespace-labeler.yaml
gracedo: this image should be free of cves

gracedo (approved):

stable/kommander/templates/namespace-labeler.yaml
```diff
@@ -0,0 +1,62 @@
...
+          command: ["kubectl"]
+          args: ["label", "namespace", "{{ .Release.Namespace }}", {{ join "," .Values.namespaceLabels  }} ]
```
joejulian, line 20: This job will fail if the namespace does not exist. That _is_ what we want, though, isn't it?

alejandroEsc: Yes it is, in addition if we do this post-install we would defeat the purpose of adding the label to the namespace. Do you suspect that the .Release.Namespace may be empty? such as in the default case?

PR #1034, opened 2021-03-13
(release/prometheus-operator-9.3.x) use busybox sh instead of the no-longer-existing bash

branden:
**What type of PR is this?**

Bug

This backports #964 to prometheus-operator 9.3.x.

The copy of the script at `staging/prometheus-operator/patch/mesosphere/templates/grafana/cronjob-home-dashboard.yaml` is already updated.

```release-note
[prometheus-operator] Fixes the cronjob that sets the Grafana home dashboard.
```

**Checklist**

### DIFF
--- a/staging/prometheus-operator/Chart.yaml
+++ b/staging/prometheus-operator/Chart.yaml
@@ -12,7 +12,7 @@ sources:
   - https://github.com/coreos/kube-prometheus
   - https://github.com/coreos/prometheus-operator
   - https://coreos.com/operators/prometheus
-version: 9.3.7
+version: 9.3.8
 appVersion: 0.38.1
 home: https://github.com/coreos/prometheus-operator
 keywords:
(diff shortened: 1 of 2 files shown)


gracedo (approved):

joejulian:
Why do we need this backported to a minor version?

gracedo:
> Why do we need this backported to a minor version?

`release/2` (and federated prometheus in yakcl) are still on 9.3.x of prometheus-operator

joejulian:
But it's a minor version. Can we just upgrade release/2?

joejulian:
The only reasons I know of that we shouldn't is:
- it won't work with kubernetes 1.17
- it won't work with helm 2
- it introduces a breaking change
- it significantly changes a UI

If none of those are true, let's reduce our complexity and just upgrade the version in release/2.

branden:
Whoops I neglected to follow up on this. @joejulian The intention with this PR is to update KBA `release/2`'s prometheus-operator 9.3.7 -> 9.3.8. Are you suggesting that we upgrade to the version on master (12.11.9) in order to pull in https://github.com/mesosphere/charts/pull/964? That we're maintaining this 9.3.x branch suggests there's a reason we haven't upgraded `release/2` to the latest. Three major versions seems like a risky upgrade that would demand additional testing.

joejulian (approved):

PR #1036, opened 2021-03-16
chore: add kubefed fork

hectorj2f:
Signed-off-by: Hector Fernandez <hectorj@gmail.com>

**What type of PR is this?**

Chore

While we are waiting to get https://github.com/kubernetes-sigs/kubefed/pull/1377 merged, we need to workaround that delay by getting our internal kubefed version based on v0.5.1.

```release-note

```

**Checklist**

### DIFF
--- a/staging/kubefed/Chart.yaml
+++ b/staging/kubefed/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v2
+description: KubeFed helm chart
+name: kubefed
+version: 0.5.2
+dependencies:
+- name: controllermanager
+  version: 0.5.2
+  repository: "https://localhost/" # Required but unused.
+  condition: controllermanager.enabled
--- a/staging/kubefed/charts/controllermanager/crds/crds.yaml
+++ b/staging/kubefed/charts/controllermanager/crds/crds.yaml
@@ -1,0 +1,1157 @@
+
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: clusterpropagatedversions.core.kubefed.io
+spec:
+  group: core.kubefed.io
+  names:
+    kind: ClusterPropagatedVersion
+    listKind: ClusterPropagatedVersionList
+    plural: clusterpropagatedversions
+    singular: clusterpropagatedversion
+  scope: Cluster
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: ClusterPropagatedVersion holds version information about the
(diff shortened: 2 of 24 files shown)


jimmidyson:
We can get https://github.com/kubernetes-sigs/kubefed/pull/1377 merged quickly (like tomorrow morning) I'm sure. Better than using a fork here imo.

staging/kubefed/charts/controllermanager/crds/crds.yaml
```diff
@@ -0,0 +1,1157 @@
...
+                description: DisabledTLSValidations defines a list of checks to ignore
+                  when validating the TLS connection to the member cluster.  This
+                  can be any of *, SubjectName, or ValidityPeriod. If * is specified,
+                  it is expected to be the only option in list.
+                items:
+                  type: string
+                type: array
+              proxyURL:
+                description: ProxyURL allows to set proxy URL for the cluster.
+                type: string
```
mhrabovcin, line 397: This field is the reason we're adding fork of `kubefed` chart.

mhrabovcin (approved):
This will unblock https://jira.d2iq.com/browse/D2IQ-74314

mhrabovcin:
@jimmidyson the upstream PR is against `master` branch. As @hectorj2f mentioned `yakcl` upgrade of `kubefed` would be non-trivial and it is not in the scope of our unidirectional work.

With merging this PR we'll get unblocked by temporarily using forked kubefed chart + custom built kubefed `v0.5.2` containers. Once we have confidence in delivering unidirectional feature on time we'll invest time in upstream PR and bumping `yakcl` to use latest `kubefed`.

staging/kubefed/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+description: KubeFed helm chart
+name: kubefed
+version: 0.5.2
```
joejulian, line 4: Please don't match the chart version to the app version. These should be expected to diverge and when they do it causes confusion.

This is true for all Chart.yaml in this PR.

joejulian (changes requested):

joejulian (commented):
I see that this is only a temporary fork so I'll approve despite my CR.

joejulian (approved):

gracedo (approved):
This is only pulled in for yakcl right? Just wondering if it's easier to pull the chart directly into the `kommander-federation` local charts dir instead of using an extra m/charts fork, esp if this is just temporary

mhrabovcin:
@gracedo I don't think we've considered this option. But you are right, this is temporary solution and the chart will be removed from our charts repo.

PR #1043, opened 2021-03-17
Kubecost Bumps should be against stable/kubecost.

orsenthil:
### Kubecost Bumps should be against stable/kubecost.

**Checklist**


gracedo (approved):
And when stable/kubecost is bumped, stable/kommander should be bumped :p 

I wonder.. could we also bump federated addon versions in yakcl when charts are bumped here? e.g for stable/kubecost, here https://github.com/mesosphere/yakcl/blob/master/chart/kommander-federation/templates/federatedclusteraddon_kubecost.yaml#L39

alejandroEsc (approved):
lgtm

PR #1051, opened 2021-03-19
chore: bump thanos to 0.3.31

d2iq-dispatch:
semi-automated bumps for chore: bump thanos to ~~0.3.22~~  0.3.31 - this is the latest patch before the jump to 0.4.0. Which we should also upgrade to but we'll do it in another PR.


gracedo (approved):

faiq:
i'm currently investigating why the chart version wasn't patched it should have been 1.0.17

joejulian (approved):

PR #1056, opened 2021-03-24
bug: fix gatekeeper pod mutations

s12chung:
**What type of PR is this?**

Bug

applyTo does not accept *, see https://open-policy-agent.github.io/gatekeeper/website/docs/mutation#extent-of-changes

no issue, but part of a new feature for D2IQ-70021

```release-note
NONE, since these changes aren't user facing yet.
```

**Checklist**

### DIFF
--- a/staging/gatekeeper/Chart.yaml
+++ b/staging/gatekeeper/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "3.3.0"
+appVersion: "v3.4.0-rc.1"
 description: Gatekeeper - Policy Controller for Kubernetes
 name: gatekeeper
-version: 0.6.5
+version: 0.6.6
 home: https://github.com/open-policy-agent/gatekeeper
 icon: https://www.openpolicyagent.org/img/opa-logo.svg
 maintainers:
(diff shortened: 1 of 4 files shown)


gracedo (approved):

joejulian:
Well... I guess there is nothing but v1 pods, but why does this matter?

s12chung:
@joejulian with the old code, the mutation webhook won't do anything.

since they're release an rc, might as well add it to this PR later tonight too.

alejandroEsc:
> @joejulian with the old code, the mutation webhook won't do anything.
> 
> since they're release an rc, might as well add it to this PR later tonight too.

What does this mean exactly? Granted it probably wont matter at this point, however we do want to mutate all pod versions right?

s12chung:
@alejandroEsc yeah, we want to mutate all pod version, but the `*` syntax isn't accepted by gatekeeper (no validation error, but it doesn't work and the docs say it won't work either), so we're making due with `v1` instead.

alejandroEsc:
> @alejandroEsc yeah, we want to mutate all pod version, but the `*` syntax isn't accepted by gatekeeper (no validation error, but it doesn't work and the docs say it won't work either), so we're making due with `v1` instead.
> 
> was testing with the new gatekeeper tag and discovered this.

At this point you are the expert, this was working with the new version so could you please add a comment on those places you are making a change so we know historically why we are setting it to v1, that way in the future we have a good idea if we ever want to switch back to "*"

s12chung:
> add a comment on those places you are making a change so we know historically why we are setting it to v1

done.

s12chung:
btw, I suspect it was working before because of the bug.

alejandroEsc:
> btw, I suspect it was working before because of the bug that we're waiting the new version to patch.

lol yeah i bet. This has been a headache but what do you expect on v1alpha1 rigth? Anyway thank you for being on top of it.

s12chung:
Lol yeah, didn't notice the v1alpha1 until after finding the fix. No worries! Enjoy your time off :)

armandgrillet (approved):
LGTM, thanks for the provided explanation.

PR #1058, opened 2021-03-24
chore: bump thanos to 0.3.23

d2iq-dispatch:
automated bumps for chore: bump thanos to 0.3.23


faiq:
Deleting this PR which was create from a chatops command in #1051

PR #1072, opened 2021-03-30
chore: batch requirements 0.11.0

d2iq-dispatch:
automated bumps for chore: batch requirements 0.11.0

### DIFF
--- a/stable/kommander/requirements.lock
+++ b/stable/kommander/requirements.lock
@@ -10,10 +10,10 @@ dependencies:
   version: 0.1.16
 - name: kommander-federation
   repository: https://mesosphere.github.io/yakcl/charts
-  version: v0.11.0
+  version: 0.11.0
 - name: kommander-licensing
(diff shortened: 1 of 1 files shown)


stable/kommander/requirements.lock
```diff
@@ -10,10 +10,10 @@ dependencies:
   version: 0.1.16
 - name: kommander-federation
   repository: https://mesosphere.github.io/yakcl/charts
-  version: v0.11.0
+  version: 0.11.0
```
joejulian, line 13: Dropping the v?

joejulian (changes requested):
This seems wrong. Blocking until I'm at my desk and can validate, or until someone else can.

gracedo:
This was probably triggered by the v0.12.0 yakcl release, but not done quite right @faiq I’ll leave this open in case you need it to troubleshoot but we should close it afterwards. Hector opened up another pr #1074 to bump yakcl

PR #1078, opened 2021-03-31
fix: use map value for issuer to set kind

TattdCodeMonkey:
**What type of PR is this?**
Bug

another follow-up to #1073 if we set the issuer we need to also be able to set the kind. moved it to its own value since its not only used in the init cert job.

In support of D2IQ-74604

```release-note
- fix: allow setting issuer kind with chart value
```

**Checklist**

### DIFF
--- a/staging/traefik/templates/certificate.yaml
+++ b/staging/traefik/templates/certificate.yaml
@@ -28,8 +28,7 @@ spec:
   {{- end }}
   issuerRef:
   {{- if .Values.certIssuerRef }}
-    name: {{ .Values.certIssuerRef.name }}
-    kind: {{ .Values.certIssuerRef.kind }}
+    {{- toYaml .Values.certIssuerRef | nindent 4 }}
   {{- else }}
     name: kubernetes-ca
     kind: ClusterIssuer


staging/traefik/templates/certificate.yaml
```diff
@@ -27,12 +27,13 @@ spec:
   secretName: {{ template "traefik.fullname" . }}-certificate
   {{- end }}
   issuerRef:
-  {{- if .Values.initCert }}
-    name: {{ .Values.initCert.ingressCertificateIssuer | default "kubernetes-ca" | quote }}
+  {{- if .Values.certIssuerRef }}
+    name: {{ .Values.certIssuerRef.name }}
+    kind: {{ .Values.certIssuerRef.kind }}
```
joejulian, line 32:
```suggestion
  {{- if .Values.certIssuerRef }}
    {{ toYaml .Values.certIssuerRef | indent 4 }}
```

joejulian (approved):

joejulian:
I tested with the values from the addon, adding certIssuerRef and it rendered the template correctly. :+1:

gracedo (approved):

PR #1079, opened 2021-03-31
[mtls-proxy] [kommander] fix ghosttunnel tag

gracedo:
**What type of PR is this?**

bug

mtls-proxy was using a nonexistent version of ghostunnel, fixed to use the correct one. Bumped mtls-proxy versions in kommander (kommander-thanos/karma)

```release-note
fix: Use correct ghostunnel version
```

**Checklist**

### DIFF
--- a/stable/mtls-proxy/values.yaml
+++ b/stable/mtls-proxy/values.yaml
@@ -5,8 +5,8 @@
 replicaCount: 1
 
 image:
-  repository: squareup/ghostunnel
-  tag: v1.5.2
+  repository: ghostunnel/ghostunnel
+  tag: v1.5.3
   pullPolicy: IfNotPresent
 
 imagePullSecrets: []
--- a/stable/mtls-proxy/values.yaml
+++ b/stable/mtls-proxy/values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: squareup/ghostunnel
-  tag: v1.5.3
+  tag: v1.5.2
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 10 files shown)


hectorj2f (approved):

stable/mtls-proxy/values.yaml
```diff
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: squareup/ghostunnel
-  tag: v1.5.3
+  tag: v1.5.2
```
branden, line 9: It looks like ghostunnel moved to a new repo, https://github.com/square/ghostunnel now redirects to ghostunnel/ghostunnel. So they moved the tag to a new corresponding dockerhub account: https://hub.docker.com/layers/ghostunnel/ghostunnel/v1.5.3/images/sha256-0f37069c6825b54211bd13f4f594318a454f31e65be90f46724c402451c37532?context=explore. I think we can change the repository to `ghostunnel/ghostunnel` and leave the tag in place.

branden (changes requested):
Thanks for catching this! I think we want to change the image repo, the v1.5.3 tag still exists but it was moved to a new repo. Suggestion inlined below.

stable/mtls-proxy/values.yaml
gracedo: oh nice find! @martinhanzik @hectorj2f I'll make the change to use the new repo and keep the image at the newer v1.5.3

branden (approved):

dkoshkin (approved):

PR #1038, opened 2021-03-16
use existing storage config if it already exists

joejulian:
**What type of PR is this?**

bug

**What this PR does/ why we need it**:

If either prometheus or alertmanager resources exist and have a storage
attribute defined in their spec, reuse that instead of what's defined
in the values.

This prevents the accidental redefinition of the volumeClaimTemplate and
the abaondomnet of old data.

**Which issue(s) this PR fixes**:

https://jira.d2iq.com/browse/COPS-6870

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Fixed an upgrade issue where upgrading from a prior chart version that had a bug where the desired PVC name prefix was being discarded. This bug caused upgrades from that version to create and use new storage claims instead of the existing ones. This feature will now always use the storage setting defined in the cluster, if there is one, reusing the existing claims.
```

**Checklist**

* [x] *If a chart is changed, the chart version is correctly incremented.*
* [x] The commit message explains the changes and why are needed.
* [x] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #1094, opened 2021-04-06
[release/prometheus-operator-9.3.x][BACKPORT] use existing storagespec

gracedo:
**What type of PR is this?**

bug

backports https://github.com/mesosphere/charts/pull/1038

If either prometheus or alertmanager resources exist and have a storage
attribute defined in their spec, reuse that instead of what's defined
in the values.

This prevents the accidental redefinition of the volumeClaimTemplate and
the abandonment of old data.

https://jira.d2iq.com/browse/COPS-6870

```release-note
Fixed an upgrade issue that occurred when upgrading from a prior chart version to a version >=9.3.0 and <=9.3.7 which had a bug where the desired PVC name prefix was being discarded. This bug caused upgrades to that version to create and use new storage claims instead of the existing ones. This feature will now always use the storage setting defined in the cluster, if there is one, reusing the existing claims.
```

**Checklist**

### DIFF
--- a/Dispatchfile
+++ b/Dispatchfile
@@ -202,7 +202,7 @@ kommander_chart_path = "stable/kommander/Chart.yaml"
 
 # Robot Actions
 action(tasks=["test-helm2", "test-helm3"], on=pull_request())
-action(tasks=["publish"], on=push(branches=["master"]))
+action(tasks=["publish"], on=push(branches=["master", "release/*"]))
 action(tasks=[do_bump_kommander], on=push(branches=["master"], paths=[kommander_chart_path]))
 action(tasks=[do_bump_addon], on=push(branches=["master"], paths=["!" + kommander_chart_path, "stable/*/Chart.yaml", "staging/*/Chart.yaml"]))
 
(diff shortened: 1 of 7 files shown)


Dispatchfile
```diff
@@ -202,7 +202,7 @@ kommander_chart_path = "stable/kommander/Chart.yaml"
 
 # Robot Actions
 action(tasks=["test-helm2", "test-helm3"], on=pull_request())
-action(tasks=["publish"], on=push(branches=["master"]))
+action(tasks=["publish"], on=push(branches=["master", "release/*"]))
```
faiq, line 205: looks good to me

staging/prometheus-operator/patch/mesosphere/patch/11_use_existing_storage_definitions.patch
```diff
@@ -0,0 +1,94 @@
...
+    If either prometheus or alertmanager resources exist and have a storage
+    attribute defined in their spec, reuse that instead of what's defined
+    in the values.
+
+    This prevents the accidental redefinition of the volumeClaimTemplate and
+    the abaondomnet of old data.
+
+diff --git a/staging/prometheus-operator/templates/NOTES.txt b/staging/prometheus-operator/templates/NOTES.txt
+index 4ef1bc3c8..f08a8afaa 100644
+--- a/staging/prometheus-operator/templates/NOTES.txt
```
faiq, line 16: is this stuff supposed to be here?

faiq (approved):
Dispatch bits look good

staging/prometheus-operator/patch/mesosphere/patch/11_use_existing_storage_definitions.patch
gracedo: yeah, it's a patch file that gets applied on top of the upstream code so we can apply our changes on top of the fork.

joejulian (approved):

joejulian:
This cannot pass CI because CI only tests against master.

PR #1107, opened 2021-04-15
Add karma chart value livenessProbe.enabled defaulting to false

branden:
**What type of PR is this?**

Feature

This adds a karma chart value `livenessProbe.enabled` that defaults to `false`. Setting this value to `true` will add a liveness probe for karma's container. This also adds corresponding `readinessProbe` values. `readinessProbe.enabled` defaults to `true`. This change breaks backwards compatibility.

This is necessary because it may be desirable to disable karma's liveness probe. Karma's ability to serve requests and pass its liveness check is dependent on the availability of its backing alertmanagers. This means that a slow alertmanager, or an alertmanager with a lot of alert data to process, could cause the karma container to fail its liveness checks and get restarted repeatedly. Because the liveness probe is dependent on more than the container process, I decided to remove it by default.

no issue

```release-note
[karma] Adds chart value `livenessProbe.enabled` that defaults to `false`, and corresponding `readinessProbe` values. `readinessProbe.enabled` defaults to `true`.
```

**Checklist**


gracedo (approved):

branden:
I noticed that the deployment's `readinessProbe` was also configured to use the `livenessProbe` settings, which is confusing. I decided to replace those with corresponding `readinessProbe` settings. And since that breaks backwards compatibility, I took the opportunity to change `livenessProbe.enabled` to `false`, which I think is a more sensible default.

stable/karma/values.yaml
```diff
@@ -68,6 +68,14 @@ securityContext: {}
 
 # configuration for liveness probe
 livenessProbe:
+  enabled: false
```
gracedo, line 71: maybe just a comment why we're defaulting to false?

gracedo (approved):

s12chung (approved):

PR #1117, opened 2021-04-23
chore: bump thanos to 0.4.0

d2iq-dispatch:
automated bumps for chore: bump thanos to 0.4.0


gracedo:
let's hold on this till post-GA

gracedo:
Will deploy latest thanos in komm2, this chart is only for komm1 so no need to bump as no bugs reported with this thanos version

PR #1124, opened 2021-04-30
chore: batch requirements 7.0.0

d2iq-dispatch:
automated bumps for chore: batch requirements 7.0.0


alejandroEsc:
could we get someone to verify these changes?

gracedo:
7.0.0 is the UI version for kommander2+, which is used in the new kommander repo that has a new kommander chart, so we shouldn't make the bump here. We will continue to use <7.0.0 bumps for this chart cc @TattdCodeMonkey to confirm

PR #1135, opened 2021-05-13
[kubecost] fix: Set default pvc size to 32Gi

gracedo:
**What type of PR is this?**

bug

CBP ran into issues with the kubecost cost-analyzer PVC size getting set to an extremely large number, which was reported in kubecost here: https://github.com/kubecost/cost-analyzer-helm-chart/issues/507. To resolve this, the team increased the default pv size from 0.2Gi (causes issues with some storage providers, in CBP case they use portworx, I think) to 32Gi. We originally avoided making this change due to concerns around upgrades whether the disk would handle automatic resizing with no loss of data, but Kubecost has a helpful doc on how to deal with such issues that we can reference in our upgrade docs as well https://github.com/kubecost/docs/blob/master/storage.md. The bulk of the data stored in this PVC is caching data for faster startup time, so I envision this not having too much risk if data loss occurs. The other data stored is configuration value changes if users were using the kubecost UI to update settings. I am not sure if this is a common use case for customers to be changing settings via the UI; we supply base configuration via helm values. In any case, we will supply the necessary docs so that an upgrade can go smoothly.

https://jira.d2iq.com/browse/D2IQ-75770
[COPS ticket TBA]

```release-note
fix(kubecost): Increase default cost-analyzer persistent volume size to 32Gi. Upgrades may require manual intervention if your provisioner does not support volume expansion.
```

**Checklist**

### DIFF
--- a/stable/kubecost/Chart.yaml
+++ b/stable/kubecost/Chart.yaml
@@ -3,6 +3,6 @@ appVersion: 1.79.1
 description: Kubecost
 name: kubecost
 home: https://github.com/mesosphere/charts
-version: 0.10.0
+version: 0.10.1
 maintainers:
   - name: gracedo
--- a/stable/kubecost/Chart.yaml
+++ b/stable/kubecost/Chart.yaml
@@ -3,6 +3,6 @@ appVersion: 1.79.1
 description: Kubecost
 name: kubecost
 home: https://github.com/mesosphere/charts
-version: 0.10.1
+version: 0.11.0
 maintainers:
   - name: gracedo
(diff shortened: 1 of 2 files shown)


stable/kubecost/Chart.yaml
```diff
@@ -3,6 +3,6 @@ appVersion: 1.79.1
 description: Kubecost
 name: kubecost
 home: https://github.com/mesosphere/charts
-version: 0.10.0
+version: 0.10.1
```
samvantran, line 6: in this case let's bump to `0.11` bc it has potential breaking effects

samvantran (approved):

s12chung (approved):
> Kubecost has a helpful doc on how to deal with such issues that we can reference in our upgrade docs as well https://github.com/kubecost/docs/blob/master/storage.md.

super helpful. should be in docs.

PR #1146, opened 2021-05-21
chore: bump grafana to 5.0.0

d2iq-dispatch:
automated bumps for chore: bump grafana to 5.0.0


gracedo:
Will deploy latest grafana in komm2, this subchart is only for komm1 so no need to bump as no bugs reported with this grafana version

PR #1147, opened 2021-05-24
[chore] Upgrading Knative Serving to 0.18.3

akirillov:
**What type of PR is this?**
chore

This PR upgrades Knative Serving version to 0.18.3

https://jira.d2iq.com/browse/D2IQ-75359

The changes are ported from the KNative Serving 0.18.3 release with minor modifications (Go template escaping in `ConfigMap`s): https://github.com/knative/serving/releases/tag/v0.18.3.

The changes were tested manually:
* by upgrading the addon to the new version with the changes from this PR
* running Kaptain workloads that depend on KNative, performing the addon upgrade, and verifying the workloads are not affected

None

**Release notes**:
```release-note
Knative Serving upgraded to version 0.18.3
```

**Checklist**


joejulian (approved):
lgtm

joejulian:
Is there an expectation that the CRDs should be upgraded when the release is updated? This won't happen automatically. You would need to follow other examples and load the CRDs into a configmap and use a preinstall job to upgrade them. Of course this will only work if the new CRD also has the old storage version (v1alpha1 _and_ v1beta1 for example) or else any custom resources that rely on the old CRD will be deleteed.

akirillov:
thanks for the review, @joejulian. I was about to implement your suggestion looking at the other charts in this repo. Let me implement it and update the PR.

joejulian (approved):

joejulian:
Loos perfect!

staging/knative/charts/serving/templates/configmap-crd.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: knative-serving
+  name: knative-crd
+data:
+  crd.yaml: |-
+{{.Files.Get "crds/crds.yaml" | printf "%s" | indent 4}}
```
filintod, line 8: do you need the printf?

filintod (commented):
just a little comment

staging/knative/charts/serving/templates/configmap-crd.yaml
akirillov: I did it by analogy with Istio (https://github.com/mesosphere/charts/blob/master/staging/istio/templates/configmap-crd.yaml#L8). I guess you need a `printf` to feed its output to `indent` but not sure about that.

akirillov: just checked that `printf` is not required here. updated the PR

filintod (approved):
LGTM

PR #1159, opened 2021-06-11
[Kube-prometheus-stack] added elasticsearch rules

alejandroEsc:
**What type of PR is this?**

Chore

We have a need to alert users when their elasticsearch cluster is not healthy or running out of space

```release-note
kube-prometheus-stack: added rules for elasticsearch
```

**Checklist**

### DIFF
--- a/staging/kube-prometheus-stack/Chart.yaml
+++ b/staging/kube-prometheus-stack/Chart.yaml
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 15.4.6
+version: 15.5.0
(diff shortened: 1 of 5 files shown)


branden (approved):

samvantran (approved):

staging/kube-prometheus-stack/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: kube-prometheus-stack
 sources:
   - https://github.com/prometheus-community/helm-charts
   - https://github.com/prometheus-operator/kube-prometheus
-version: 15.4.6
+version: 15.5.0
```
gracedo, line 21: IMO for forks I think we should stick with patch versions since we're not technically on the next minor version of the upstream chart yet.

PR #1244, opened 2021-11-01
fix: Use absolute domains, fixes COPS-7078

vespian:
**What type of PR is this?**
Bug

As COPS-7078 showed, some of the URLs we were using were relative thus causing them to be proxied through proxy if `http_proxy` variables were set. In order to prevent that, we need to use absolute names instead, so that the requests fall into `no_proxy` whitelist by default, without the need for the user to manually adjust anything.

https://jira.d2iq.com/browse/COPS-7078 

I have not tested it yet. Correctly recreating the environment customer is using is around 0.5day. I asked our support to help customer test it instead and come back to us as this will be much easier for them - the already have the test environment up&running.

```release-note
Some of the URLS used kommander are now absolute - have cluster.local extension by default.
```

**Checklist**


vavy:
Is the 10903 - a misprint for the Thanos? Should it be 10902?

vespian:
> Is the 10903 - a misprint for the Thanos? Should it be 10902?

Great catch! Fixing it now.

stable/kommander/values.yaml
```diff
@@ -151,12 +151,12 @@ grafana:
       datasources:
         - name: ThanosQuery
           type: prometheus
-          url: http://kommander-kubeaddons-thanos-query-http.kommander:10902/
+          url: http://kommander-kubeaddons-thanos-query-http.kommander.svc.cluster.local:10902/
```
jimmidyson, line 154: As discussed, we don't need `.cluster.local` suffix here as `.svc` is included in `noproxy` documentation. Please remove it as the cluster domain is actually configurable in a kubernetes cluster (although this is rarely changed...).
```suggestion
          url: http://kommander-kubeaddons-thanos-query-http.kommander.svc:10902/
```

jimmidyson (changes requested):

stable/kommander/values.yaml
vespian: I have made the requested changes, please have another look.

The `cluster.local` suffix is used all across the repo, including this file. I have not changed this though as I think this is outside of the scope of this change.

vespian:
Test failing on unrelated reloader error https://jira.d2iq.com/browse/D2IQ-80600

stable/kommander/values.yaml
```diff
@@ -233,13 +233,16 @@ kubecost:
 
     fullnameOverride: "kommander-kubecost-cost-analyzer"
     global:
+      # We need to specify full domain name explicity in order to make
+      # http_proxy work.  Otherwise requests go through proxy instead directly
+      zone: cluster.local
```
makkes, line 238: What happens if a customer uses a different cluster domain?

vespian: They can specify it in the override. The idea is that A domain needs to be specified to make charts use absolute domain names.

makkes: Can we please document how to set the cluster domain name?

vespian: Yes, but this is out of the scope of this Jira/PR I think.

makkes: >  Yes, but this is out of the scope of this Jira/PR I think.

I'm kind of on the fence about this. Before this change this component was agnostic to the cluster domain. Now it is not so we should document that fact right away so that users are able to apply the correct configuration in case they use a different cluster domain.

vespian: These are defaults. If the customer was using custom domain before, they most probably were using an override of some sort, so nothing changes for them. The users this change addresses are the ones that did not provide any domain. For them, the chart was falling back to non-absolute dns names which break with http_proxy.

vespian: Responded in https://mesosphere.slack.com/archives/C01LNF6KML5/p1637688810412800

branden (approved):

PR #1246, opened 2021-11-15
feat: use gatekeeper constraints to enforce spec.serviceAccountName

takirala:
Flux kustomizations and HelmRelease created by kommander are populated with `spec.serviceAccountName` to make sure they do not deploy manifests outside the scope of their namespace.

A user might unintentionally (or with mal intent) deploy resources on a namespace they do not have access to by just creating a helmRelease or a flux kustomization. We must ensure we guard against this. Flux recommends having a admissioncontroller (like gatekeeper) to enforce this.

This PR adds constraint templates to ensure ALL kustomizations / HelmReleases in a namespace have a service account in them.

**What type of PR is this?**

```release-note

```

**Checklist**

### DIFF
--- a/staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
+++ b/staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.disableDkpConstraints }}
 ---
 apiVersion: templates.gatekeeper.sh/v1beta1
 kind: ConstraintTemplate
@@ -48,3 +49,4 @@ spec:
     namespaceSelector:
       matchLabels:
         kommander.d2iq.io/managed-by-kind: "Project"
+{{- end }}
--- a/staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
+++ b/staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
@@ -1,0 +1,50 @@
+---
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: requiredserviceaccountname
+  annotations:
+    description: >-
+      Requires the given resource to have the
+      .spec.serviceAccountName field set.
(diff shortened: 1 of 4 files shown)


staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if not .Values.disableDkpConstraints }}
```
takirala, line 1: Please suggest better naming :D 

i thought of reusing the value `disableValidatingWebhook` but having a new value lets us disable just this without disabling the validation.

gracedo: maybe we can add flux somewhere in there? in case we want to add more constraints in the future but allow them to be disabled separately `disableDKPFluxConstraints` or even drop the DKP. And I find the `disable` double-negative a little unnatural, can we instead use `enable`? I do see we are using another disable above..

gracedo (approved):
Haven't worked much with gatekeeper but assuming the manifests have been tested out on a cluster and works, lgtm

staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
branden: +1 @gracedo's suggestions. I'd normally default to favoring consistency and be OK with `disabled` but there's only one other similarly named value and I dislike the double negative enough that I'm willing to overlook it lol.

branden (approved):

staging/gatekeeper/templates/gatekeeper-dkp-constraints.yaml
takirala: renamed to `enableFluxConstraints`

makkes (approved):

takirala:
Superseeded by https://github.com/mesosphere/kommander-applications/pull/58/files

PR #1269, opened 2022-01-12
[gatekeeper] fix: Allow configuring sideeffects

gracedo:
**What type of PR is this?**

bug

Missing a `sideEffects: None` in the `mutatingwebhookconfiguration` which is required for flux (komm 2.x). Adding `mutations.sideEffects` field to allow configuring `sideEffects` without potentially breaking backwards-compat with 1.x if gatekeeper ever needs to be bumped there. Default would still be to omit the field - we will set it when bumping to this version in 2.x

https://jira.d2iq.com/browse/COPS-7127

(on another note, the patch scripts are not really updated. doesn't look like they've been used for some time so some scripts are outdated. I've only run the one relevant patch I needed here which didn't break anything)

```release-note

```

**Checklist**

### DIFF
--- a/staging/gatekeeper/Chart.yaml
+++ b/staging/gatekeeper/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "3.4.0-rc.1"
 description: Gatekeeper - Policy Controller for Kubernetes
 name: gatekeeper
-version: 0.6.8
+version: 0.6.9
 home: https://github.com/open-policy-agent/gatekeeper
 icon: https://www.openpolicyagent.org/img/opa-logo.svg
 maintainers:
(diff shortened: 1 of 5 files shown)


takirala (approved):
we might end up not using this version of gatekeeper chart anymore in preference of using upstream gatekeeper chart (which already includes this fix), but LGTM.

https://github.com/mesosphere/kommander-applications/pull/95

gracedo:
Ah, got it. Ok yeah we can still get this merged for 1.x and 2.1, and maybe until mesosphere/kommander-applications#95 is ready/merged we can at least get this change in just in case switching to upstream doesn't happen for 2.2

samvantran (approved):

joejulian (approved):

PR #1289, opened 2022-02-21
feat: Provie cert-manager CRDs chart

mikolajb:
The chart provides `cert-manager` CRDs.

**What type of PR is this?**
Feature

We don't want to rely on the upstream `cert-manager` chart to create and manage CRDs. We would like to keep the CRDs on the attached cluster even when `cert-manager` kommander app is uninstalled. Also, we would like to manage CRDs using flux. Both goals can be achieved by using to helmreleases in kommander `cert-manager` app - the upstream `cert-manager` chart and the chart provided in this PR.

https://jira.d2iq.com/browse/D2IQ-85306

The chart was tested in https://github.com/mesosphere/kommander-applications/pull/201 and https://github.com/mesosphere/kommander/pull/1453

```release-note

```

**Checklist**

### DIFF
--- a/stable/cert-manager-crds/Chart.yaml
+++ b/stable/cert-manager-crds/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v2
+description: A chart that bundles cert-manager CRDs with cert-manager helm chart
+home: https://github.com/mesosphere/kommander
+name: cert-manager-crds
+version: v0.0.1
+appVersion: "0.0.1"
+maintainers:
+  - name: mhrabovcin
+  - name: mikolajb
--- a/stable/cert-manager-crds/Chart.yaml
+++ b/stable/cert-manager-crds/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: A chart that bundles cert-manager CRDs with cert-manager helm chart
 home: https://github.com/mesosphere/kommander
 name: cert-manager-crds
-version: "v0.0.1"
+version: "v1.7.1"
 appVersion: "v1.7.1"
 maintainers:
   - name: mhrabovcin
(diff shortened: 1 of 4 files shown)


stable/cert-manager-crds/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+description: A chart that bundles cert-manager CRDs with cert-manager helm chart
+home: https://github.com/mesosphere/kommander
+name: cert-manager-crds
+version: v0.0.1
+appVersion: "0.0.1"
```
mhrabovcin, line 6: The `appVersion` should be matching the version of `cert-manager` for which are CRDs installed. In our case `v1.7.1`.

stable/cert-manager-crds/Makefile
```diff
@@ -0,0 +1,5 @@
+CERT_MANAGER_VERSION ?= v1.7.1
+
```
mhrabovcin, line 2: Can we add a new make target that would update `Chart.yaml` `appVersion`?

stable/cert-manager-crds/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+description: A chart that bundles cert-manager CRDs with cert-manager helm chart
+home: https://github.com/mesosphere/kommander
+name: cert-manager-crds
+version: v0.0.1
```
mhrabovcin, line 5: I'd think about having the version match the `appVersion` as it will be very clear in `kommander-applications` repo which versions are bundled.

mikolajb: @mhrabovcin I like this idea, I'm applying it but also keeping this conversation open till it gets accepted (since we don't follow this pattern for other charts).

mhrabovcin: Given that we now update both properties with CRD i think we should also have one additional make target that will update both, the CRDs and these properties.

mhrabovcin: I see this was already implemented, thanks.

mhrabovcin (approved):
🚀

jimmidyson (approved):

PR #1296, opened 2022-03-05
[stable/dex] fix: Prevent name collision for config and cert secrets

branden:
**What type of PR is this?**

Bug

The hardcoded secret name "dex" will collide with the name of the config secret if the chart release is named "dex". This will lead to a failed chart install or upgrade if cert-manager creates the secret before Helm can.

https://jira.d2iq.com/browse/COPS-7193

I looked at updating the deployment to reflect the new cert secret name, but it already expects it by default: https://github.com/mesosphere/charts/blob/7e5e0248cef5bc23e5716a4f8d3a13c9145609b9/stable/dex/templates/deployment.yaml#L1-L3

So it looks like we'll also need to remove this line from the default dex chart config to receive this fix: https://github.com/mesosphere/kommander-applications/blob/dd82bdd3ce7d2c87c1ad45598dfb9ea8b1662aba/services/dex/2.9.12/defaults/cm.yaml#L35

```release-note
[stable/dex] Fixes a race condition between Helm and cert-manager that may cause a chart install to fail.
```

**Checklist**

### DIFF
--- a/stable/dex/Chart.yaml
+++ b/stable/dex/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: dex
-version: 2.9.12
+version: 2.9.13
 appVersion: 2.22.0
 description: Dex
 keywords:
--- a/stable/dex/templates/certificate.yaml
+++ b/stable/dex/templates/certificate.yaml
@@ -15,7 +15,7 @@ metadata:
     # https://github.com/helm/helm/issues/5482
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  secretName: dex
+  secretName: {{ template "dex.fullname" . }}-tls
   issuerRef:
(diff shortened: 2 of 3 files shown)


branden:
I looked at updating the deployment to reflect the new cert secret name, but it already expects it by default: https://github.com/mesosphere/charts/blob/7e5e0248cef5bc23e5716a4f8d3a13c9145609b9/stable/dex/templates/deployment.yaml#L2-L3

So it looks like we'll also need to remove this line from the default dex chart config to receive this fix: https://github.com/mesosphere/kommander-applications/blob/dd82bdd3ce7d2c87c1ad45598dfb9ea8b1662aba/services/dex/2.9.12/defaults/cm.yaml#L35

gracedo (approved):

mhrabovcin (approved):

PR #1308, opened 2022-03-30
fix: Remove home dashboard cronjob template

cbuto:
**What type of PR is this?**

bug 

This PR removes the home dashboard cronjob template completely. It was initially disabled and replaced in https://github.com/mesosphere/charts/pull/1294 but since the default name for the cronjob was removed as well, it can cause issues during an upgrade or install if someone has `homeDashboard.cronJob.enabled=<value>` set at all (helm will try to evaluate the template even if its disabled). 

To avoid any issues, we can remove the template entirely and helm will just ignore the home dashboard values if they are still set in any overrides.

```release-note

```

**Checklist**


gracedo (commented):
KPS is a fork so we tried to make patching easy - as have this patch dir https://github.com/mesosphere/charts/tree/master/staging/kube-prometheus-stack/patch that is supposed to be the "source of truth" for any patches that we apply on top of the fork. So we would want to actually remove https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch_2_mesosphere_cron_jobs.sh, https://github.com/mesosphere/charts/tree/master/staging/kube-prometheus-stack/patch/mesosphere/templates/grafana and https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch_6_mesosphere_values.sh#L21-L23. then if you run https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch.sh it should make the changes to the actual chart (essentially what you've done).

cbuto:
> KPS is a fork so we tried to make patching easy - we have this patch dir https://github.com/mesosphere/charts/tree/master/staging/kube-prometheus-stack/patch that is supposed to be the "source of truth" for any patches that we apply on top of the fork. So we would want to actually remove https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch_2_mesosphere_cron_jobs.sh, https://github.com/mesosphere/charts/tree/master/staging/kube-prometheus-stack/patch/mesosphere/templates/grafana and https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch_6_mesosphere_values.sh#L21-L23. then if you run https://github.com/mesosphere/charts/blob/master/staging/kube-prometheus-stack/patch/patch.sh it should make the changes to the actual chart (essentially what you've done).

totally missed that! I think it should be good now, thanks!

gracedo (approved):

branden (approved):

PR #1323, opened 2022-05-19
feat: add kommander-cert-federation chart

takirala:
**What type of PR is this?**

Because of how we ship the current `cert-federation` chart with a static version 0.0.1 - https://github.com/mesosphere/kommander/blob/main/chart/cert-federation/Chart.yaml#L7 recent changes (https://github.com/mesosphere/kommander/commits/main/chart/cert-federation/templates/cert-federation.yaml) have updated the chart here - https://github.com/mesosphere/kommander/blob/gh-pages/charts/cert-federation-v0.0.1.tgz (overwritten) even though we shipped the same chart in older versions of kommander. This is not desired as it effectively makes it an unknown of what "version" of the chart a customer is running as its updated (overwritten) with every change to that chart. 

This PR fixes above issue by adding this chart into `m/charts` repo and thus enforcing a proper versioning scheme (e.g.: bumping version everytime there is a change).

https://jira.d2iq.com/browse/D2IQ-89706

```release-note
NONE
```

**Checklist**

### DIFF
--- a/stable/kommander-cert-federation/Chart.yaml
+++ b/stable/kommander-cert-federation/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v2
+appVersion: "0.0.2"
+description: A Helm chart to create and federate TLS certificates used by kommander internals
+home: https://github.com/mesosphere/charts
+name: kommander-cert-federation
+type: application
+version: v0.0.2
+maintainers:
+  - name: mikolajb
+  - name: takirala
--- a/stable/kommander-cert-federation/Chart.yaml
+++ b/stable/kommander-cert-federation/Chart.yaml
@@ -4,7 +4,7 @@ description: A Helm chart to create and federate TLS certificates used by komman
 home: https://github.com/mesosphere/charts
 name: kommander-cert-federation
 type: application
-version: v0.0.2
+version: 0.0.2
 maintainers:
   - name: mikolajb
   - name: takirala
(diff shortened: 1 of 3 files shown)


gracedo (approved):

gracedo:
Just one note if we can publish 0.0.1 with the previous version of the chart that 2.1/2.2 are deploying right now?

stable/kommander-cert-federation/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+appVersion: "0.0.2"
+description: A Helm chart to create and federate TLS certificates used by kommander internals
+home: https://github.com/mesosphere/charts
+name: kommander-cert-federation
+type: application
+version: v0.0.2
```
gracedo, line 7:
```suggestion
version: 0.0.2
```

`version: A SemVer 2 version (required)` I think we can't use `v` to be semver compatible

takirala:
> Just one note if we can publish 0.0.1 with the previous version of the chart that 2.1/2.2 are deploying right now?

we can use 0.0.2 chart in 2.2.1+ 
we can stick with 0.0.1 chart in kommander repo for 2.1.2+ (open to other ideas ;))

no need to publish a 0.0.1 chart here i think :thinking:

takirala:
CI failing on `master` branch as well with 
```
+ docker exec --interactive ct kubectl wait --timeout=180s --for=condition=Available --selector=app=dummylb deploy
[18:28:45](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Konvoy_ChartPr/3457789?buildTab=log&focusLine=229&linesState=229)
  error: timed out waiting for the condition on deployments/dummylb
```

gracedo:
CI seems to be broken across the repo. This chart has been tested manually. Admin-merging

PR #1348, opened 2022-08-11
feat: configurable log level for tfa

takirala:
**What type of PR is this?**

Make the logging level configurable for tfa chart

Current log level is `trace` which can get very noisy.

```release-note
- Configurable log level for `traefik-forward-auth`
```

**Checklist**

### DIFF
--- a/staging/traefik-forward-auth/values.yaml
+++ b/staging/traefik-forward-auth/values.yaml
@@ -81,7 +81,8 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
-env: {}
+env:
+  LOG_LEVEL: info
 
 nodeSelector: {}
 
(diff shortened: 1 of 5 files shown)


cbuto (approved):

staging/traefik-forward-auth/values.yaml
```diff
@@ -81,7 +81,8 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
-env: {}
+env:
+  LOG_LEVEL: info
```
gracedo, line 85: should we default it to `trace` since that is what it was hardcoded to before?

gracedo (approved):

staging/traefik-forward-auth/values.yaml
takirala: i think `trace` isn't an ideal default so i think we can keep this at info here.

takirala:
This is no longer (strictly) needed

PR #1367, opened 2022-10-24
fix: remove invalid dependency repository value [D2IQ-93788]

aweris:
**What type of PR is this?**

Chart metadata fix

Invalid repository value for dependency causing failures in dependency updates. 

--

--

--

**Checklist**

### DIFF
--- a/staging/kubefed/Chart.yaml
+++ b/staging/kubefed/Chart.yaml
@@ -5,5 +5,4 @@ version: 0.5.2
 dependencies:
 - name: controllermanager
   version: 0.5.2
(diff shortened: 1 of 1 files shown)


cbuto (approved):
it looks like we use the upstream chart in k-apps 

but the upstream also sets this [repository value](https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/Chart.yaml#L9) fwiw

aweris:
But Jimmi made that upstream change. So it doesn't count :D

PR #1404, opened 2023-04-04
feat: Add upgrade script and priority class support for istio

gracedo:
**What type of PR is this?**

feat

Modeling off the KPS upgrade script, add an upgrade script for istio to automate upgrading our fork so that it is easier to do (and anybody can do it).

To upgrade the fork, update the `ISTIO_TAG` envvar in upgrade_operator.sh, commit/push it, and run `./upgrade_operator.sh`.
It will automatically:
- Copy the upstream chart at ISTIO_TAG
- Add any new templates that we want in addition to existing upstream templates
- Apply patches to various files (changes to values.yaml, Chart.yaml, deployment.yaml)

I have run the script against this branch, and you can see that a few updates were made to Chart.yaml, but otherwise a pretty clean diff (which is good here, to ensure that the patches give us what we want our fork to look like which in the past we were manually updating). Take a look at the first set of  `chore` commits for the automated commits made by the script to double check the patches look correct.

I also added a *new* patch on top of existing patches to add support for `priorityClassName` option. The latter set of patches are for these.

I will open a followup PR after this merges to test upgrading to 1.17.1.

https://d2iq.atlassian.net/browse/D2IQ-96278

```release-note

```

**Checklist**


staging/istio/Chart.yaml
```diff
@@ -1,21 +1,19 @@
 apiVersion: v2
 name: istio
-version: 1.16.2
-appVersion: 1.16.2
-description: Helm chart for Istio
+version: 1.16.3
+appVersion: 1.16.3
```
gracedo, line 4: I bumped the patch version here to get CI passing. will follow up with upgrade to 1.17.3 upstream after this merges

lukeogg (approved):
LGTM!

PR #1419, opened 2023-04-26
feat: Bump KPS to 45.21.0

gracedo:
**What type of PR is this?**

feat

Bump KPS to latest upstream version

```release-note

```

**Checklist**


staging/kube-prometheus-stack/values.yaml
```diff
@@ -3533,7 +3605,7 @@ thanosRuler:
     image:
       registry: quay.io
       repository: thanos/thanos
-      tag: v0.30.1
+      tag: v0.30.2
```
cwyl02, line 3608: do we want to align this with `thanos` k-app to use `v0.31.0`?

cwyl02 (approved):

staging/kube-prometheus-stack/README.md
```diff
@@ -80,6 +80,23 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
...
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_probes.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.63.0/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
+```
+
```
lukeogg, line 99: It looks like we apply the CRDs ourselves, but it is unlikely that any of our users will see this documentation and won't know to apply/update CRDs manually. Is it necessary to apply them manually?

lukeogg (approved):

staging/kube-prometheus-stack/README.md
gracedo: Flux actually does this for us! 
https://github.com/mesosphere/kommander-applications/blob/3288f2fa0ef1b0c98f57fc08dde36f54af80fae2/services/kube-prometheus-stack/44.2.1/helmrelease/kube-prometheus-stack.yaml#L23
https://github.com/fluxcd/helm-controller/blob/main/docs/api/helmrelease.md#crdspolicystring-alias

staging/kube-prometheus-stack/values.yaml
gracedo: hmm I might just stick with the verified version that works with KPS. esp if it doesn't help us if it would still be a separate image going into the airgapped bundle.

(and instead of making the change here, we should override this in k-apps defaults if we want to use a different image)

PR #1421, opened 2023-05-10
fix: remove priorityClass from job

cwyl02:
**What type of PR is this?**

feature

this fails kommander upgrade -- `priorityClass` field is immutable on `Job`

```release-note

```

**Checklist**

### DIFF
--- a/stable/dex/Chart.yaml
+++ b/stable/dex/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: dex
-version: 2.12.1
+version: 2.12.2
 appVersion: 2.35.2
 description: Dex
 keywords:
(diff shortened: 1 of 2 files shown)


cbuto (commented):
looks like you just need to bump the chart version

this seems good to me, i suppose the alternative is doing some pre-upgrade script to delete the job prior to the chart upgrade but given that this only runs once i think its 👌

cbuto (approved):

gracedo (approved):
there is always the `force` field but not sure if it's worth it here
```
         // Force forces resource updates through a replacement strategy.
	// +optional
	Force bool `json:"force,omitempty"`
```

PR #1435, opened 2023-06-12
feat: Bump kubecost

gracedo:
**What type of PR is this?**

feat

Bump kubecost to latest to pull in support for adding priority to additional components

https://d2iq.atlassian.net/browse/D2IQ-97194

```release-note

```

**Checklist**


stable/kubecost/values.yaml
```diff
@@ -63,6 +63,8 @@ cost-analyzer:
     serviceAccounts:
       nodeExporter:
         create: false
+    kube-state-metrics:
+      priorityClassName: dkp-high-priority
```
cbuto, line 67: do you think its better to define these in the k-apps repo to align with the other applications? it might be easier for users to override if they see the priority classes directly in the default configmaps 🤔

cbuto (approved):

stable/kubecost/values.yaml
gracedo: yeah that's a good reason.. my reasoning here was just because this is a parent chart that is basically kommander-specific but that's the only reason why really. I'll move it into the default value overrides

gracedo: reverted

cbuto (approved):

takirala (approved):

PR #1439, opened 2023-06-21
feat: Add priority to dex jobs

gracedo:
**What type of PR is this?**

feat

Set priority class on dex jobs. However, for this upgrade, we will still hit this immutable error since the job already exists. So, I've added a pre-upgrade hook here to clean up the Jobs prior to upgrading. This will only target jobs deployed in chart version 1.11.1 (released in DKP 2.5), so will only recreate jobs during upgrade from DKP 2.5 to 2.6 (and not within 2.6). We can remove this hook in following releases.

The certs jobs will re-run on this upgrade but from my testing I don't see any effects, but would like some confirmation from our sso stack sme's.

https://d2iq.atlassian.net/browse/D2IQ-97813

```release-note

```

**Checklist**


gracedo:
```
Error: UPGRADE FAILED: pre-upgrade hooks failed: warning: Hook pre-upgrade dex/templates/pre-upgrade-delete-jobs.yaml failed: Job.batch "dex-1l8c8myp4b-delete-jobs" is invalid: spec.template.spec.containers[0].image: Required value
```
upgrade test keeps failing :( I wonder if it's because the new values are not getting used in the upgrade? (but I feel like this would have failed in more tests whenever we added new value fields 🤔 ) When I do `helm template` the image is populating just fine
```
# Source: dex/templates/pre-upgrade-delete-jobs.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: release-name-dex-delete-jobs
  namespace: default
  annotations:
    "helm.sh/hook": pre-upgrade
    "helm.sh/hook-weight": "4"
    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
spec:
  template:
    metadata:
      name: release-name-dex-delete-jobs
    spec:
      serviceAccountName: release-name-dex-pre-upgrade
      restartPolicy: OnFailure
      containers:
        - name: kubectl
          image: "bitnami/kubectl:1.26.4"
          command:
            - sh
            - -c
            - kubectl delete jobs.batch -l 'app.kubernetes.io/component in (job-grpc-certs, job-web-certs),app.kubernetes.io/name=dex' --cascade=orphan -n default
            - ```

**Comment by gracedo** [comment]
```
      --reset-values                               when upgrading, reset the values to the ones built into the chart
      --reuse-values                               when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored
```
It looks like ct is using `--reuse-values` flag for helm upgrade, but here we probably want `--reset-values` to pull in the new set of values. I did attempt to pass in `--reset-values` into the helm args but it was causing the install command to fail (and would be ignored anyways since `--reuse-values` is already being passed in). Searching the ct repo yielded this PR to enable this, by one of our very own alum 😄 https://github.com/helm/chart-testing/pull/531 since this is not currently supported, we'll need to ignore the failing test and admin merge to get this in.

I tested this with a chart hosted from my GH.

```
dex-delete-jobs-vr6jv                                                     dkp-critical-priority
dex-grpc-certs-vvc8m                                                      dkp-critical-priority
```

File: stable/dex/templates/pre-upgrade-delete-jobs.yaml [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+      {{- end }}
+      containers:
+        - name: kubectl
+          image: "{{ .Values.kubectlImage }}"
+          command:
+            - sh
+            - -c
+            - kubectl delete jobs.batch -l 'app.kubernetes.io/component in (job-grpc-certs, job-web-certs),app.kubernetes.io/name=dex' --cascade=orphan -n {{ .Release.Namespace }}
```
cbuto, line 75: i'm not sure what this job exactly is used for but now that we are deleting it, i think it would be recreated and reran during the upgrade, right?

I'm wondering if rerunning it would cause any issues during upgrade 🤔. since it regenerates some certs, are there other clients that need to get the new CA or something?

gracedo: I tested this upgrade on the daily and didn't see any (obvious) issues with regenerating the certs, but I'd like @mhrabovcin or @palexster to confirm. Since this job really doesn't need to run on every upgrade though.. I will remove the TTL, and also move the pre-upgrade hook out of the chart and into k-apps, to limit it to only run during upgrade from 2.5 -> 2.6.

gracedo: dex helmrelease isn't behind a flux kustomization in k-apps, so upgrade will result in a reinstall 😬 should be a small amount of downtime in dex.. I don't think it will have significant impact.

gracedo: ok, went all over the place with this one 😓 but.. weighing all the things. I wanted to make a pre-upgrade hook here somewhat agnostic from DKP versioning but maybe it's fine, since this chart is really only used in the context of kommander. So to avoid the reinstall of dex (due to moving it behind a flux kustomization in k-apps), I can leave the pre-upgrade hook here with the version constraint to only delete the jobs if they were deployed as part of the 2.11.1 dex chart version (the version deployed in dkp 2.5). That way it only deletes the job for upgrades from dkp 2.5 to 2.6, and not within 2.6. Keeping it in the chart means that the priority of the pre-upgrade hook can also be configurable (vs putting it in a manifest in k-apps)

gracedo: done in https://github.com/mesosphere/charts/pull/1439/commits/d1c83bb945494e9673dc8a8418eb883402aa2b04

queuing up k-apps branch: https://github.com/mesosphere/kommander-applications/pull/1350

stable/dex/templates/pre-upgrade-delete-jobs.yaml
```diff
@@ -0,0 +1,77 @@
...
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+      {{- end }}
+      containers:
+        - name: kubectl
+          image: "{{ .Values.kubectlImage }}"
+          command:
+            - sh
+            - -c
+            - kubectl delete jobs.batch -l 'app.kubernetes.io/component in (job-grpc-certs, job-web-certs),app.kubernetes.io/name=dex,helm.sh/chart=dex-2.11.1' --cascade=orphan -n {{ .Release.Namespace }}
```
cbuto, line 76: should we add `--ignore-not-found` to this for future releases?

stable/dex/templates/pre-upgrade-delete-jobs.yaml
cbuto: sounds good to me!

stable/dex/templates/pre-upgrade-delete-jobs.yaml
cbuto: oh yeah that's right 😅

cbuto (approved):
lgtm! though still unsure of the consequences of rerunning this job

gracedo:
@palexster can you verify that it would be fine to rerun the dex job to regenerate certs? When I tested on the daily cluster, I didn't notice any issues (but maybe I wasn't looking at the right thing :) )

palexster:
> @palexster can you verify that it would be fine to rerun the dex job to regenerate certs? When I tested on the daily cluster, I didn't notice any issues (but maybe I wasn't looking at the right thing :) )

I think we are safe. Those jobs are only re-bootstrapping a CA and new certificates, so every new execution would just bootstrap them again. 

Btw, I don't see the purpose of re-bootstrapping a new CA at every update (the lowest default value is 10y of lifespan). We should probably consider migrating to cert-manager, managing them properly, and getting rid of those jobs, as suggested by @mhrabovcin.

palexster (approved):

gracedo:
> We should probably consider migrating to cert-manager, managing them properly, and getting rid of those jobs

@palexster Can you (or @mhrabovcin) create a ticket in the backlog so we don't lose track of this improvement please? Thank you!

PR #1448, opened 2023-08-31
feat: add required flag to be able to view grafana dashboard

Azhovan:
**What type of PR is this?**
Bug

**What this PR does/ why we need it**:
Although users with the users with `kommander-federated-view` role could access the grafana UI they are unable to receive any data. since [this role ](https://github.com/mesosphere/kommander/blob/df5ae6fe0481b192da74e1ee585947b0e2509b5f/chart/kommander/templates/federatedclusterroles_default.yaml#L62)is an aggregated role, I think it is necessary to update the labels on  `dkp-centralized-grafana-grafana-view` to propagate them into  `kommander-federated-view`

**Which issue(s) this PR fixes**:
https://d2iq.atlassian.net/browse/D2IQ-98716

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

* [ ] *If a chart is changed, the chart version is correctly incremented.*
* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #1449, opened 2023-09-07
fix: add required flag to be able to view grafana dashboard

gracedo:
**What type of PR is this?**

bug

backports https://github.com/mesosphere/charts/pull/1448 into previous KPS release (for dkp 2.6)
This is going into the `release/kube-prometheus-stack-46.8.x` release branch which was created off the SHA where 46.8.0 was released for DKP 2.6.0. Since we have already bumped KPS to 48.x on `main` (for DKP 2.7) we need to backport this fix to the version released in DKP 2.6 in order to avoid jumping 2 major app versions in DKP patch release.

I _think_ we just need to trigger the TC publish job for the release branch after this PR merges. It looks like this is what we did the last time we had to publish a version behind the latest (https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Konvoy_Chart/3441173)

https://d2iq.atlassian.net/browse/D2IQ-98716

```release-note

```

**Checklist**


Azhovan (approved):

gracedo:
I don't think tests will work/pass on backports like these :/

Azhovan:
> I don't think tests will work/pass on backports like these :/

can we just simply override the [version] kube-prometheus-stack (https://github.com/mesosphere/charts/blob/dd9c36131c9d37e8cb39c948bdc9c9c0160c3af2/staging/kube-prometheus-stack/upgrade_operator.sh#L16)  and release again the chart 
```shell
PROMETHEUS_TAG=kube-prometheus-stack-47.1.0
```

cbuto (approved):

gracedo:
@Azhovan I'm not entirely sure what would happen since we haven't done it that way before, but it is a little weird to have an older version on `main` (we would need to bump again after). And since we'll need to backport this fix to 2.4/2.5, we would have to keep releasing older versions on `main`, and I don't think all the patches will apply correctly on older versions, potentially.

For this PR since it's already here, let's try this backporting/release branch method and see if it works. Since this is how we've done it in the past, I think it's probably the better option than trying something new 🤞🏻

gracedo:
The TC job failed (probably targets are now out of date), I manually ran `make publish` on the release branch which published the version: https://github.com/mesosphere/charts/commit/a4bd1a0849d45899ae77d850c13a6b9ef39057b5

PR #1575, opened 2025-04-21
feat(cosi-driver-nutanix): Add priorityclass to cosi controller

gracedo:
**What type of PR is this?**

feat

add priorityclass to cosi objectstorage-controller if set in values

https://jira.nutanix.com/browse/NCN-105642

```release-note

```

**Checklist**

### DIFF
--- a/staging/cosi-driver-nutanix/Chart.yaml
+++ b/staging/cosi-driver-nutanix/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cosi-driver-nutanix
 description: A Helm chart to deploy Nutanix COSI driver
 type: application
-version: 0.0.10
+version: 0.0.11
 appVersion: "v0.3.0"
 icon: https://www.nutanix.com/content/dam/nutanix/global/icons/products/svg/Nutanix-Objects-40.svg
 maintainers:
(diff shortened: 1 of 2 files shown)


takirala (commented):
I don't think there is any usecase of updating the cosi controller in cosi-driver-nutanix chart.

Konvoy (and caren) uses cosi chart directly and its configured in konvoy as well - via patches https://github.com/mesosphere/konvoy2/blob/main/pkg/capi/generator/templates/cosi/kustomize/controller/kustomization.yaml 

There cosi-driver-nutanix is not supposed to install the cos controller in scope of NKP and i guess its mainly here for non konvoy clusters.

gracedo:
> This cosi-driver-nutanix is not supposed to install the cos controller in scope of NKP and i guess its mainly here for non konvoy clusters.

Would attached non-konvoy clusters use this chart to install it? If they do, maybe just worth including it unless we don't actually want users enabling it

takirala:
> > This cosi-driver-nutanix is not supposed to install the cos controller in scope of NKP and i guess its mainly here for non konvoy clusters.
> 
> Would attached non-konvoy clusters use this chart to install it? If they do, maybe just worth including it unless we don't actually want users enabling it

Since those clusters are usually non nutanix those users will not install this chart (which is specifically for nutanix cosi storage). I guess there is no harm, i just don't want us to diverge from upstream in case we ever get rid of this chart (as was the intention when writing this chart, we may get rid of this chart once kommander supports OCI)

PR #1578, opened 2025-04-30
feat: add docker-registry Chart

dkoshkin:
**What type of PR is this?**

Feature

**What this PR does/ why we need it**:

Adds a public chart https://github.com/twuni/docker-registry.helm, with some additional features as patches to support deploying it as a StatefulSet and syncing images across multiple replicas.

**Which issue(s) this PR fixes**:

https://jira.nutanix.com/browse/NCN-107252

**Special notes for your reviewer**:
Tested the installation locally enabling these new features.

**Does this PR introduce a user-facing change?**:

```release-note

```

**Checklist**

* [ ] *If a chart is changed, the chart version is correctly incremented.*
* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [x] The documentation is updated where needed.

PR #1579, opened 2025-05-02
fix: set a release tag for docker-registry regsync image

dkoshkin:
**What type of PR is this?**

Bug

Follow up to https://github.com/mesosphere/charts/pull/1578 to set a stable tag for the regsync image.

```release-note

```

**Checklist**

### DIFF
--- a/staging/docker-registry/values.yaml
+++ b/staging/docker-registry/values.yaml
@@ -272,8 +272,8 @@ statefulSet:
   syncer:
     enabled: true
     image:
-      repository: regclient/regsync
-      tag: latest
+      repository: docker.io/regclient/regsync
+      tag: v0.8.3
       pullPolicy: IfNotPresent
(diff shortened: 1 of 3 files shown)


staging/docker-registry/values.yaml
```diff
@@ -272,8 +272,8 @@ statefulSet:
   syncer:
     enabled: true
     image:
-      repository: regclient/regsync
-      tag: latest
```
gracedo, line 276: should these be applied as a patch on top of upstream values?

dkoshkin: Ahh good point, fixing.

dkoshkin: These are new values added in the fork, added another patch so we can easily update the regsync version in the future.

gracedo (approved):

PR #1583, opened 2025-05-14
feat: Update Knative chart to use Knative operator

lukeogg:
**What type of PR is this?**

Replace the Knative chart with a chart that deploys Knative using the operator CRDs. This chart requires the upstream operator helm chart to be installed. It then creates CRDs for `knative-serving` and `knative-eventing` with the appropriate configuration. The configuration for these values is overridable via the values.yaml.

[NCN-107271](https://jira.nutanix.com/browse/NCN-107271)

Testing on a local Nutanix cluster using the chart hosted in GH Pages.

```release-note
Knative chart updated to restore full configurability.
```

**Checklist**


johnliu25:
thanks @lukeogg for the change. It seems it could help us to remove some of the steps currently documented under
1. https://github.com/nutanix-cloud-native/nkp-nutanix-product-catalog/blob/main/services/nutanix-ai/metadata.yaml#L72-L106
2. https://github.com/nutanix-cloud-native/nkp-nutanix-product-catalog/blob/main/services/nutanix-ai/metadata.yaml#L118-L119

@lukeogg Could you please address the failure under chart testing?

asekretenko:
@lukeogg Am I getting this correctly that all the resources created by this chart are highly specific to NAI?

If so, why keeping this chart at all?

Wouldn't it be better to move all this NAI-specific stuff into NAI itself, so that https://github.com/mesosphere/kommander-applications/pull/3523 would serve only one function - installing the KNative Operator?

lukeogg:
@johnliu25 Yes, we should remove those steps from the app's description.

lukeogg:
@asekretenko They are not specific to NAI. Knative is generic and could be used for other things. That said, we don't. We included Knative for Kaptain and now keep it for NAI.

This chart does the same thing the old chart does. It deploys `knative-serving` and `knative-eventing` in their own namespaces. It does it with the Operator instead of our own chart. This will reduce maintenance significantly. This also restores the configurability that we lost at some point.

Technically, this should be moved into a repository that NAI owns, probably the [nkp-nutanix-product-catalog](https://github.com/nutanix-cloud-native/nkp-nutanix-product-catalog). It can then be maintained and deployed along with the NAI app.

staging/knative/Chart.yaml
```diff
@@ -1,16 +1,9 @@
...
-dependencies:
-  - name: eventing
-    version: 1.17.2
-    condition: eventing.enabled
-  - name: serving
-    version: 1.17.0
-    condition: serving.enabled
+  - name: lukeogg
+source:
+  -
```
alembiewski, line 9: Is it left blank by purpose?

alembiewski (approved):

johnliu25 (approved):

gracedo (approved):

staging/knative/Chart.yaml
lukeogg: Nope. :-)

PR #1592, opened 2025-06-26
fix: set pod anti affinity for docker-registry deployment and statefulset

supershal:
**What type of PR is this?**

Sets pod anti affinity for docker-registry pods and deployments
exposes nodeAffinity configuration for docker-registry.

Directly setting soft pod anti affinity on the stateful set to always distribute the pods on different nodes.
This is done directly in templates so that we can use helper template functions in matching labels.

I tested using single and multi CP node kind clusters.

```release-note

```

**Checklist**

### DIFF
--- a/staging/docker-registry/README.md
+++ b/staging/docker-registry/README.md
@@ -94,7 +94,7 @@ their default values.
 | `proxy.secretRef`           | The ref for an external secret containing the proxyUsername and proxyPassword keys         | `""`            |
 | `namespace`                 | specify a namespace to install the chart to - defaults to `.Release.Namespace`             | `{{ .Release.Namespace }}` |
 | `nodeSelector`              | node labels for pod assignment                                                             | `{}`            |
-| `affinity`                  | affinity settings                                                                          | `{}`            |
+| `nodeAffinity`              | node affinity settings                                                                          | `{}`            |
 | `tolerations`               | pod tolerations                                                                            | `[]`            |
 | `ingress.enabled`           | If true, Ingress will be created                                                           | `false`         |
 | `ingress.annotations`       | Ingress annotations                                                                        | `{}`            |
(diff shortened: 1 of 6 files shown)


staging/docker-registry/README.md
```diff
@@ -94,7 +94,7 @@ their default values.
 | `proxy.secretRef`           | The ref for an external secret containing the proxyUsername and proxyPassword keys         | `""`            |
 | `namespace`                 | specify a namespace to install the chart to - defaults to `.Release.Namespace`             | `{{ .Release.Namespace }}` |
 | `nodeSelector`              | node labels for pod assignment                                                             | `{}`            |
-| `affinity`                  | affinity settings                                                                          | `{}`            |
+| `nodeAffinity`              | node affinity settings                                                                          | `{}`            |
```
supershal, line 97: tolerations and nodeAffinity overrides will be set in CAREN's values.yaml.

staging/docker-registry/templates/statefulset.yaml
```diff
@@ -99,9 +99,21 @@ spec:
       {{- if .Values.nodeSelector }}
       nodeSelector: {{ toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.affinity }}
-      affinity: {{ toYaml .Values.affinity | nindent 8 }}
+      affinity:
+      {{- if .Values.nodeAffinity }}
+        nodeAffinity: {{ toYaml .Values.nodeAffinity | nindent 10 }}
       {{- end }}
+        podAntiAffinity:
```
dkoshkin, line 106: You don't think we can get away without this change and just leave it to the client to set the `affinity` values directly? This chart already exposes `.Values.podLabels` that can be used to select.

Making this change deviates us further from the upstream chart and removes existing ability to set `podAffinity`

PR #1651, opened 2026-04-14
feat: pc version check for ndk

pawan-bobbili-ntnx:
 - Adding a precheck of PC version check as a pre-install and pre-upgrade step for ndk 2.1


rohitkumar0510:
please check these CI checks failing are expected, I can see them passing for recent PRs that have been merged.

PR #1659, opened 2026-07-07
feat:istio helm upgrade script

pawan-bobbili-ntnx:
**Summary**
 - Adds a script to automate upgrading our forked **istio-helm-** charts to a new Istio version while **automatically retaining our customizations**
 
**JIRA**
 - [NCN-114159](https://jira.nutanix.com/browse/NCN-114159)[NCN-114159](https://jira.nutanix.com/browse/NCN-114159)
 
 ## Approach — "derive and replay"
On upgrade, for each chart the script:
1. Reads the current Istio version from the wrapper `Chart.yaml` `appVersion` (the merge **base**).
2. `helm pull`s **both** the base and target upstream charts from the published OCI registry (`oci://gcr.io/istio-release/charts/<name>`).
3. Computes our delta (`base` → `live`) and **replays it onto the target** upstream via a 3-way merge (`git merge-file`):
   - files only we changed are re-applied,
   - files only upstream changed are taken as-is,
   - a genuine line-level overlap is surfaced as a **conflict** (loud stop, markers left, no commit).
4. Bumps the wrapper-level Istio references (`appVersion`, sub-chart dependency `version`, pinned image `tag`).
5. Commits one change per chart for easy review.
On any non-conflict error, a `trap` rolls the working tree back to the starting
commit (`set -Eeuo pipefail`, so the trap fires for failures inside helpers too).

### DIFF
--- a/hack/istio-helm-upgrade/lib/helpers.sh
+++ b/hack/istio-helm-upgrade/lib/helpers.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+# Shared helpers for the istio-helm upgrade flow.
+#
+# Expects REPO_ROOT to be exported by the caller and charts.sh to be sourced
+# (for CHART_REGISTRY / upstream_chart_name).
+
+# Pull an upstream chart at a tag and extract it into <destdir>/<name>,
+# replacing anything already there. Normalizes the extracted directory to
+# <name> even if the published chart name differs.
+function pull_upstream_chart {
+  local name="$1"
+  local tag="$2"
+  local destdir="$3"
+  local uname
+  uname=$(upstream_chart_name "${name}")
+  rm -rf "${destdir:?}/${name}" "${destdir:?}/${uname}"
+  helm pull "${CHART_REGISTRY}/${uname}" --version "${tag}" --untar --untardir "${destdir}"
+  if [ "${uname}" != "${name}" ]; then
+    mv "${destdir}/${uname}" "${destdir}/${name}"
+  fi
+}
+
+# Replay our customizations onto a freshly pulled upstream tree.
+#
+# Our customizations are, by definition, the difference between the upstream we
+# are currently on (base) and our live chart (live). We reproduce that same
+# difference on top of the new upstream (result), file by file, using a 3-way
+# merge so that upstream's own changes to a file are preserved and only genuine
+# overlaps become conflicts.
+#
+#   live   = staging/istio-helm-<name>/charts/<name>   (source of truth)
+#   base   = upstream chart at the CURRENT version
+#   result = upstream chart at the NEW version (mutated in place)
+#
+# Prints the relative path of any file that could not be merged cleanly (one per
+# line) to stdout; informational notes go to stderr.
+function replay_customizations {
+  local live="$1"
+  local base="$2"
+  local result="$3"
+  local rel
+
+  # Files we added or modified relative to the upstream we came from.
+  while IFS= read -r rel; do
+    if [ ! -e "${base}/${rel}" ]; then
+      # A file we added -> carry it over verbatim.
+      mkdir -p "${result}/$(dirname "${rel}")"
+      cp "${live}/${rel}" "${result}/${rel}"
+    elif ! cmp -s "${base}/${rel}" "${live}/${rel}"; then
+      # A file we modified -> replay our change onto the new upstream copy.
+      if [ ! -e "${result}/${rel}" ]; then
+        mkdir -p "${result}/$(dirname "${rel}")"
+        cp "${live}/${rel}" "${result}/${rel}"
+        echo "  note: upstream removed ${rel}; kept our modified copy" >&2
+      elif ! git merge-file --quiet "${result}/${rel}" "${base}/${rel}" "${live}/${rel}" 2>/dev/null; then
+        echo "${rel}"
+      fi
+    fi
+  done < <(cd "${live}" && find . -type f | sed 's|^\./||')
+
+  # Files we deleted relative to the upstream we came from.
+  while IFS= read -r rel; do
+    if [ ! -e "${live}/${rel}" ] && [ -e "${result}/${rel}" ]; then
+      rm -f "${result}/${rel}"
+    fi
+  done < <(cd "${base}" && find . -type f | sed 's|^\./||')
+}
+
+# Stage everything under the given path and commit, but only if there is
+# something to commit.
+function git_commit_if_changes {
+  local path="$1"
+  local msg="$2"
+  git -C "${REPO_ROOT}" add -A "${path}"
+  if git -C "${REPO_ROOT}" diff --cached --quiet -- "${path}"; then
+    echo "No changes to commit for: ${msg}"
+    return 0
+  fi
+  git -C "${REPO_ROOT}" commit -q -m "${msg}"
+}
+
+# Portable in-place sed (works with both GNU and BSD/macOS sed).
+function sed_inplace {
+  local expr="$1"
+  local file="$2"
+  local tmp
+  tmp=$(mktemp)
+  sed "${expr}" "${file}" > "${tmp}" && mv "${tmp}" "${file}"
+}
+
+# Replace the old Istio version string with the new one in a wrapper-level file
+# (Chart.yaml / values.yaml). This only touches Istio version references: the
+# mesosphere sub-chart versions (grafana/prometheus-operator/security) and the
+# kubectl image tags never equal the Istio tag, so they are left untouched.
+function bump_istio_version {
+  local file="$1"
+  local old="$2"
+  local new="$3"
+  [ -f "${file}" ] || return 0
+  [ "${old}" = "${new}" ] && return 0
+  sed_inplace "s/${old//./\\.}/${new}/g" "${file}"
+}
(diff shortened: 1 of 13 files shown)


hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,107 @@
...
+# -E (errtrace) is required so the ERR/rollback trap also fires for failures
+# inside helper functions (e.g. helm pull), not just top-level commands.
+set -Eeuo pipefail
+shopt -s dotglob
+
+# Pinned default. Bump this (or pass a tag) when upgrading.
+DEFAULT_ISTIO_TAG=1.29.0
+ISTIO_TAG="${1:-${ISTIO_TAG:-${DEFAULT_ISTIO_TAG}}}"
+
+BASEDIR=$(dirname "$(realpath "$0")")
```
rohitkumar0510: can we mention the portability of `realpath`, if it is compatible with macOS or not?

hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,107 @@
...
+REPO_ROOT=$(git -C "${BASEDIR}" rev-parse --show-toplevel)
+STARTING_REV=$(git -C "${REPO_ROOT}" rev-parse HEAD)
+export BASEDIR REPO_ROOT ISTIO_TAG
+
+source "${BASEDIR}/lib/charts.sh"
+source "${BASEDIR}/lib/helpers.sh"
+
+# The Istio version we are upgrading *from*, read from a wrapper Chart.yaml.
+OLD_ISTIO_TAG=$(sed -n 's/^appVersion:[[:space:]]*//p' "${REPO_ROOT}/staging/istio-helm-base/Chart.yaml" | head -1)
```
rohitkumar0510, line 52: can we add a comment in `values.yaml` wherever `appVersion` key exists, that if it gets updated, it needs to be updated here as well, otherwise the script breaks

hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,107 @@
...
+
+rollback() {
+    set +x
+    echo "ERROR running upgrade. Rolling back to ${STARTING_REV}." >&2
+    git -C "${REPO_ROOT}" reset --hard "${STARTING_REV}"
```
rohitkumar0510, line 71: can add a comment that this might accidentally remove unrelated uncommitted work, so to ensure a clean working tree before proceeding

hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,107 @@
...
+    exit 1
+}
+trap 'rollback' ERR
+
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "${TMPDIR}"' EXIT
+
+echo "Upgrading istio-helm charts ${OLD_ISTIO_TAG} -> ${ISTIO_TAG}"
+
+CONFLICTS=""
```
rohitkumar0510: can be an array? so that later we can loop through it, and echo them

rohitkumar0510:
The script assumes availability of: git, helm, sed, mktemp, realpath, this can be put in readme/upgrade.md so that can be made available before proceeding to run the script.

rohitkumar0510:
also, what happens if NEW_TAG==OLD_TAG, seems like it still does the steps, maybe something like

```
if [[ "${OLD_ISTIO_TAG}" == "${ISTIO_TAG}" ]]; then
    echo "Already on ${ISTIO_TAG}"
    exit 0
fi
```

hack/istio-helm-upgrade/upgrade.sh
pawan-bobbili-ntnx: using `cd "$(dirname "$0")" && pwd` now.

hack/istio-helm-upgrade/upgrade.sh
pawan-bobbili-ntnx: Thanks, done

pawan-bobbili-ntnx:
> The script assumes availability of: git, helm, sed, mktemp, realpath, this can be put in readme/upgrade.md so that can be made available before proceeding to run the script.

Added PRereq comment

pawan-bobbili-ntnx:
> also, what happens if NEW_TAG==OLD_TAG, seems like it still does the steps, maybe something like
> 
> ```
> if [[ "${OLD_ISTIO_TAG}" == "${ISTIO_TAG}" ]]; then
>     echo "Already on ${ISTIO_TAG}"
>     exit 0
> fi
> ```

While doing testing i made it intentionally like this, to make sure of the correctness of this script.
I was thinking, it can be just kept as if tomorrow we upgrade the script, this one scenario can be used to test the correctness.

rohitkumar0510 (approved):

rohitkumar0510:
@pawan-bobbili-ntnx please help adding a reference `upgrade PR` from the current version to some other version either in the ticket or in PR description

pawan-bobbili-ntnx:
> @pawan-bobbili-ntnx please help adding a reference `upgrade PR` from the current version to some other version either in the ticket or in PR description

Thanks for the suggestion.
https://github.com/mesosphere/charts/pull/1660 is the dummy PR raised by script

shivendra-ntnx (approved):

hack/istio-helm-upgrade/lib/helpers.sh
```diff
@@ -0,0 +1,103 @@
...
+      # A file we added -> carry it over verbatim.
+      mkdir -p "${result}/$(dirname "${rel}")"
+      cp "${live}/${rel}" "${result}/${rel}"
+    elif ! cmp -s "${base}/${rel}" "${live}/${rel}"; then
+      # A file we modified -> replay our change onto the new upstream copy.
+      if [ ! -e "${result}/${rel}" ]; then
+        mkdir -p "${result}/$(dirname "${rel}")"
+        cp "${live}/${rel}" "${result}/${rel}"
+        echo "  note: upstream removed ${rel}; kept our modified copy" >&2
+      elif ! git merge-file --quiet "${result}/${rel}" "${base}/${rel}" "${live}/${rel}" 2>/dev/null; then
```
NRoshanRai: git merge-file returns:
0 — clean merge
positive — number of conflict regions
negative (128+) — hard error (e.g. cannot read a file)

The ! negation and 2>/dev/null collapse errors and conflicts into the same path. An I/O error or permissions problem would appear as a "conflict" in the summary rather than triggering the ERR trap and rolling back. 

The user would see conflict markers instructions for a file that might actually be corrupt.

It's a rare case, likely low frequency, but the impact is high when it does happen: a true operational failure gets mislabeled as “manual conflict,” which can hide the root cause.

hack/istio-helm-upgrade/lib/helpers.sh
```diff
@@ -0,0 +1,103 @@
...
+    fi
+  done < <(cd "${live}" && find . -type f | sed 's|^\./||')
+
+  # Files we deleted relative to the upstream we came from.
+  while IFS= read -r rel; do
+    if [ ! -e "${live}/${rel}" ] && [ -e "${result}/${rel}" ]; then
+      rm -f "${result}/${rel}"
+    fi
+  done < <(cd "${base}" && find . -type f | sed 's|^\./||')
+}
```
NRoshanRai, line 87: If we deleted a file relative to old upstream, and the new upstream also modified that same file, the script silently deletes the new upstream copy. Our deletion wins without any notification. Contrast this with the inverse case (upstream deleted a file we modified) on line 55, which correctly prints a note: to stderr.

better to add a symmetric note when removing a file upstream still ships

hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,123 @@
...
+#   ./upgrade.sh 1.30.0          # upgrade to a specific tag
+
+# -E (errtrace) is required so the ERR/rollback trap also fires for failures
+# inside helper functions (e.g. helm pull), not just top-level commands.
+set -Eeuo pipefail
+shopt -s dotglob
+
+# Pinned default. Bump this (or pass a tag) when upgrading.
+DEFAULT_ISTIO_TAG=1.29.0
+ISTIO_TAG="${1:-${ISTIO_TAG:-${DEFAULT_ISTIO_TAG}}}"
```
NRoshanRai, line 35: better to handle early exit for same-version "upgrade" 
if already in 1.29.0 and then execute `./upgrade.sh 1.29.0`

hack/istio-helm-upgrade/upgrade.sh
```diff
@@ -0,0 +1,123 @@
...
+    # Bump the wrapper-level Istio version references (appVersion + sub-chart
+    # dependency version in Chart.yaml, and pinned image tag in values.yaml).
+    bump_istio_version "${wrapper}/Chart.yaml" "${OLD_ISTIO_TAG}" "${ISTIO_TAG}"
+    bump_istio_version "${wrapper}/values.yaml" "${OLD_ISTIO_TAG}" "${ISTIO_TAG}"
+done
+
+if [ "${#CONFLICTS[@]}" -gt 0 ]; then
+    # Conflicts are expected human decisions, not script failures: keep the
+    # merged tree (with markers) so they can be resolved, and do not roll back.
+    trap - ERR
```
NRoshanRai: The trap - ERR on line 103 disarms the rollback before the exit 1 so that conflicts don't trigger a hard reset. This is correct, but the logic is position-dependent and fragile 

A comment at the trap site documenting this relationship would help, or the rollback function itself could check a flag.

hack/istio-helm-upgrade/upgrade.sh
pawan-bobbili-ntnx: we are deliberately not handling to ensure the correctness, that is same version comes again, no fike change should happen.

NRoshanRai (approved):

PR #1663, opened 2026-07-27
fix: istio ambient upgrade redundant envs

pawan-bobbili-ntnx:
***Summary***
 - `ztunnel.env.CA_ADDRESS/XDS_ADDRESS` is rendered twice once from chart's internal block and one from user config override while enabling ambient mode in 2.17 NKP. In 2.18, all HRs are server-side apply, which is making this redunandant envs fail.
 - Thus, in template of ztunnel daemonset, skipping these envs when entered from config overrdie

### DIFF
--- a/staging/istio-helm-ztunnel/charts/ztunnel/templates/daemonset.yaml
+++ b/staging/istio-helm-ztunnel/charts/ztunnel/templates/daemonset.yaml
@@ -162,9 +162,15 @@ spec:
               resource: limits.cpu
               divisor: "1"
         {{- with .Values.env }}
+        {{/* CA_ADDRESS and XDS_ADDRESS are already rendered above; skip them here so a
+             legacy override (ztunnel.env.CA_ADDRESS/XDS_ADDRESS) can't create a duplicate
+             env entry that strict server-side apply rejects. */}}
+        {{- $reserved := list "CA_ADDRESS" "XDS_ADDRESS" }}
         {{- range $key, $val := . }}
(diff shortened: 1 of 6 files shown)


staging/istio-helm-ztunnel/charts/ztunnel/templates/daemonset.yaml
```diff
@@ -162,11 +162,17 @@ spec:
               resource: limits.cpu
               divisor: "1"
         {{- with .Values.env }}
+        {{/* CA_ADDRESS and XDS_ADDRESS are already rendered above; skip them here so a
```
rohitkumar0510, line 165: why did we choose to ignore the override instead of Chart defaults? or are we saying that we shouldn't allow these env to be overridden

pawan-bobbili-ntnx: Yrah, these shouldn't be toched by customer in our qualified scenarios.

rohitkumar0510:
bump all the components' versions of Istio, they are shipped together since

rohitkumar0510 (approved):

staging/istio-helm-ztunnel/charts/ztunnel/templates/daemonset.yaml
rohitkumar0510: note: to be mentioned in Istio app doc

NRoshanRai (approved):

